PR #17, opened 2015-07-15
Fix free with install attempt of non-existent packages. Fix reinstall bug.

ppadmavilasom:
When attempting to find version of a non-existent package, after the call fails, null out out var so it wont be freed on exit.
Include reinstall packages in transaction.


toliaqat:
:+1:

PR #24, opened 2015-10-13
Add installroot command line option.

xiaolin-vmware:


ppadmavilasom:
This supports installs. Can we also add the rootdir (if set) when we create a sack (hy_sack_create)
Also, please send the pull request to dev

toliaqat:
It does not seems to be creating a chroot to install the rpms in installroot. Can we check if yum is doing the same thing. I thought creating a chroot to isolate the installation was part of requirement for this task. Not having a chroot the postinstall scripts of rpm could harm the actual root???

ppadmavilasom:
@toliaqat assumes chroot exists and rpmdb is initialized. This is the same with yum
<code>rpm --root /myroot --initdb</code>

toliaqat:
rpm will fill out the new chroot with db, but a rpm package during installation can execute some post-install scripts that could be pointing to / instead of installdir.

ppadmavilasom:
@toliaqat: do we have an instance that this fails? rpmtsSetRootDir is called with the new root.

PR #27, opened 2015-10-14
Use existing sack on search and check-local

luiszun:
Use existing sack for searching packages and check-local, otherwise it operates with the currently installed packages.


luiszun:
Done. I had missed what check-local was supposed to do.

ppadmavilasom:
Can you please squash the commits or just create one with the relevant changes only? history will stay cleaner that way

luiszun:
Done

PR #33, opened 2015-10-21
Exit with status 0 when result is nothing todo

luiszun:
Nothing to do doesn't mean an error with an operation.


toliaqat:
:+1:

luiszun:
*facepalm*
I corrected it.

PR #37, opened 2015-10-23
Print install size of packages

luiszun:


luiszun:
Done, used built in function and removed extras

PR #86, opened 2019-10-09
Minor fixes all over the place.

sshedi:
Changes include:
1. Fix build warnings like unused parameters, sign mismatch while comparing.
2. Using printf instead of fprintf(stdout, ...).
3. Added header guard in few .h files.
4. Added a rule for silent build in configure.ac
5. Removal of trailing spaces, my vim does it automatically

PR looks big but all are minor fixes, hope it's okay and reviewing won't be a headache :)

### DIFF
--- a/client/includes.h
+++ b/client/includes.h
@@ -18,6 +18,12 @@
  * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
  */
 
+#ifndef __CLIENT_INCLUDES_H__
+#define __CLIENT_INCLUDES_H__
+
+/* Use this to get rid of variable/parameter unused warning */
+#define UNUSED(var) ((void)(var))
+
 #include <stdio.h>
 #include <stdint.h>
 #include <stdlib.h>
@@ -59,3 +65,4 @@
 #include "../common/prototypes.h"
 #include "prototypes.h"
 
+#endif /* __CLIENT_INCLUDES_H__ */
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -31,27 +31,22 @@ progress_cb(
 {
     double dPercent;
 
-    if(dlTotal > 0)
-    {
-        dPercent = ((double)dlNow / (double)dlTotal) * 100.0;
-        if(!isatty(STDOUT_FILENO))
-        {
-            fprintf(stdout, "%s %3.0f%% %ld\n",
-                (char*)pUserData,
-                dPercent,
-                dlNow);
-        }
-        else
-        {
-            fprintf(
-                stdout,
-                "%-35s %10ld  %5.0f%%\r",
-                (char*)pUserData,
-                dlNow,
-                dPercent);
-        }
-        fflush(stdout);
+    UNUSED(ulNow);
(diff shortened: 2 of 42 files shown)


client/includes.h
```diff
@@ -18,6 +18,12 @@
  * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
  */
 
+#ifndef __CLIENT_INCLUDES_H__
+#define __CLIENT_INCLUDES_H__
+
+/* Use this to get rid of variable/parameter unused warning */
+#define UNUSED(var) ((void)(var))
```
ppadmavilasom: lets just get rid of unused variables instead

client/remoterepo.c
```diff
@@ -31,27 +31,20 @@ progress_cb(
...
-                (char*)pUserData,
-                dlNow,
-                dPercent);
-        }
-        fflush(stdout);
-    }
+    UNUSED(ulNow);
+    UNUSED(ulTotal);
+
+    if (dlTotal <= 0)
```
ppadmavilasom: please add explicit braces always
if (dlTotal <= 0)
{
    return 0;
}

ppadmavilasom (changes requested):
Looks good
please use explicit braces
please remove unused variables if possible
please find out why the CI is failing. it fails at 3 tests

sshedi:
@ppadmavilasom - I have update my PR please review.

python/tdnfbase.c
```diff
@@ -19,7 +19,7 @@ base_dealloc(PPY_TDNF_BASE self)
 static PyObject *
 base_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
 {
-    PPY_TDNF_BASE self = NULL;
+    PPY_TDNF_BASE self;
```
ppadmavilasom: please keep the init. it might look redundant as the next line here is assigning value to self. however, code could change later. if left uninitialized, it is more of a burden for review. so please leave init code as is here and elsewhere

python/tdnfpyrepodata.c
```diff
@@ -26,8 +26,8 @@ TDNFPyRepoDataNew(
     PyObject *args,
     PyObject *kwds)
 {
-    uint32_t dwError = 0;
-    PPY_TDNF_REPODATA self = NULL;
+    uint32_t dwError;
```
ppadmavilasom: do not leave uninitialized

ppadmavilasom (changes requested):
please leave init code as is as mentioned in review discussions

python/tdnfbase.c
sshedi: I partially agree with you on this one, yes initialising variables during declaration is a good approach(in defensive programming) and there's absolutely no harm in it AFAIK. But IMO a programmer should know what he is doing and and he should be in total control. I'll make the modifications as per your suggestions and push the changes, I dis this because that's what linux kernel code has taught me so far, same goes with not using curly braces in single line if else blocks :)

ppadmavilasom (approved):
thanks for your help.

PR #92, opened 2019-11-28
Switch to cmake builds and fix a bunch of other stuffs

sidcha:
This PR attempts to cleanup the tdnf repo top-level by minimizing build related files. To that end, it creates a single `CMakeLists.txt` (and a directory `cmake/` with helpers) files to manage all targets. By design it avoids the subdirectory-specific build script chaining.

### Summary of changes

* All `auto*` build scripts are removed.
* Builds only shared libraries.
* The autotools based testing framework does not play well with CMake and hence needed to be removed.
* CMake supports a native CPack based RPM building method. In favour of this, the tdnf spec file that was earlier used generate RPM was removed.
* Files like` tdnf-completion.bash` and `tdnf.conf` that where in the top-level was moved into `resources/`.
* `build-docker.sh` was renamed as `docker-build.sh` for better grouping (so all docker related files show up in the same place in ls).
* the newly built tdnf binary is passed over to pytests for testing.
* Switch to a git-tag based version generation system. ie., a git tag like `v2.0.0` or `v2.0` or just `2.0.0` somewhere in the current tree's history is used to generate the package version markup. It plays very nicely with https://github.com/vmware/tdnf/releases.
* `PACKAGE_NAME` and `PACKAGE_VERSION` are not made available to source code as `const char *` instead of passing it as a build-time macro. Suitable getters are added in `common/`.

### Build Steps

As with any cmake project,

```
mkdir build  && cd build
cmake ..
make
```
### Targets:

The following targets are now supported by the new build scripts.

* **make check** - run pytests
* **make install** - install binaries (supports DESTDIR=<path> for non standard install)
* **make tdnfcli** - builds only libtdnfcli.so
* **make tdnfclient** - builds only libtdnf.so
* **make tdnf** - builds only tdnf binary
* **make python** - builds tdnf python module
* **make package** - produces an RPM package out of the built binaries
* **make package_pytdnf** - produces RPMs for tdnf python module
* **make package_source** - produces source rpms


sidcha:
@keerthanakalyan @ppadmavilasom, please review. 

My tree looks like this: https://github.com/cbsiddharth/tdnf/tree/topic/cmake_build

ppadmavilasom (commented):
@cbsiddharth thanks for the PR. very well organized. good job.
A path to integrating this might be:
migrate the current make check tests to pytest - we should check that in (while continuing to pull changes and running ci in your fork)
please send a separate PR with the api version and resource relocation changes from this PR. we should be able to get them in right away while we work towards the rest

sidcha:
@ppadmavilasom, moving the API version out of this patch-set would be rather time consuming. I forgot to add a point in the summary. I have edited the PR to include this point now.

> Switch to a git-tag based version generation system. ie., a git tag like `v2.0.0` or `v2.0` or just `2.0.0` somewhere in the current tree's history is used to generate the package version markup. It plays very nicely with https://github.com/vmware/tdnf/releases.

I have played with git-tag  based release management in the past, I believe this implementation is best of them all.

ppadmavilasom (commented):
Looks good. let us convert and engage existing tests. if you need help, please split and assign existing tests to other contributors or tag help wanted.

sidcha:
@ppadmavilasom, With the last 9 commits, I have migrated all tests that were being done in `tests/` (and some more) into this PR.

Now this can be reviewed and merged without any gaps in test coverage. I have edited the initial comment to remove the **"MUST NOT BE MERGED"** status of this patch.

ppadmavilasom:
@cbsiddharth thank you

@keerthanakalyan @tapakund please help review

@cbsiddharth in your original summary, tests are todo. now that you have completed test migration, please adjust accordingly. also(in summary), cmoka is not related to this change.

keerthanakalyan:
@cbsiddharth : Could you also mention in the summary about the multiversion tdnf pytests which are skipped now and will be added in the later PRs. These tests require a test repo which is hosted in our internal discus-filer , we need to find out an alternative way for these tests.

ppadmavilasom (commented):
LGTM

sidcha:
> multiversion tdnf pytests which are skipped now

@keerthanakalyan, not anymore.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+)
+
+install(TARGETS ${TDNF_BIN} RUNTIME DESTINATION bin COMPONENT binary)
+
+##
+## Other static resources
+##
+
+install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/resources/tdnf.conf"
+    DESTINATION "/ect/tdnf"
```
keerthanakalyan: Typo. /etc/

sidcha: Which one? `tdnf.conf` is installed inside `/etc/tdnf/`.

keerthanakalyan: DESTINATION "/ect/tdnf" 
Instead of /etc/ its mentioned as /ect/

tapakund: typo here as well.

tapakund: /ect

sidcha: Ahh... Sorry. Got it now.

sidcha: Ack.

cmake/CreatePackages.cmake
```diff
@@ -0,0 +1,19 @@
...
+
+set(CPACK_PACKAGE_NAME "tdnf")
+set(CPACK_PACKAGE_VENDOR "VMware, Inc.")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "tdnf is a yum/dnf equivalent which uses libsolv and libcurl")
+set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
+set(CPACK_PACKAGE_VERSION ${VERSION_STRING})
+set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/COPYING")
+
+set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}-${CMAKE_SYSTEM_PROCESSOR}")
+#set(CPACK_RPM_PACKAGE_ARCHITECTURE "noarch")
```
tapakund, line 13: tdnf is build with x86_64 arch, shall we add this here.

pytests/conftest.py
```diff
@@ -47,11 +69,21 @@ def check_package(self, package):
...
+        if pkgversion:
+            pkg = pkgname + '-' + pkgversion
+        else:
+            pkg = pkgname
+        self.run([ 'tdnf', 'erase', '-y', pkg ])
+        assert(self.check_package(pkgname) == False)
+
+    def install_package(utils, pkgname, pkgversion=None):
+        if pkgversion:
+            pkg = pkgname + '-' + pkgversion
```
tapakund: are we passing pkgversion, release, ph1/ph2/ph3 as well in the pkgversion string? or just package version? If only pkg version then it will install pkg with the same pkg version with latest release.

pytests/conftest.py
```diff
@@ -47,11 +69,21 @@ def check_package(self, package):
...
+        else:
+            pkg = pkgname
+        utils.run([ 'tdnf', 'install', '-y', '--nogpgcheck', pkg ])
```
keerthanakalyan: I see that by default "--nogpgcheck" is added for all installs. Did we have test case scenario which tests gpgcheck?

ppadmavilasom: we have a requirement to sign and use a local gpgkey to verify. we should add this with the external gpgkey change. good for now

pytests/tests/test_clean.py
```diff
@@ -19,28 +19,28 @@ def teardown_test(utils):
 
 def test_clean_no_args(utils):
     ret = utils.run([ 'tdnf', 'clean' ])
-    assert (ret['retval'] == 135)
+    assert (ret['retval'] == 903)
```
tapakund: do we have any hash defined for these ret values? it is returning from tdnf, can we use readable names.

keerthanakalyan (approved):
Reviewed. Changes looks good.

tapakund (commented):
minor comments. Pls address and merge.

tapakund:
@cbsiddharth  pls hold on for today. I am still going through the test files.

pytests/conftest.py
sidcha: No GPG checks have not been added yet. I believe it was not part of the initial test tuit (in tdnf as well). If they were done as a part of harness tests, then I will be migrated in another PR.

sidcha:
> @cbsiddharth pls hold on for today. I am still going through the test files.

Sure.

pytests/conftest.py
sidcha: No release is not being passed. Correct me if I am wrong, I was under the impression that the latest version from that release (in case of photon-3, ph3) would be picked up if no version and release is mentioned. If that's the case, then this shouldn't cause any issues and work across releases without modifications.

pytests/tests/test_clean.py
sidcha: Thought about this, felt it was not needed. Keerthana brought this up earlier. I will explain in a regular comment so no one misses it.

pytests/conftest.py
keerthanakalyan: Earlier test does not use --nogpgcheck in the command line. Instead it uses both local and photon repos and use the .repo file to decide whether gpgcheck has to be enabled for that repo using gpgcheck=1 inside /etc/yum.repos.d/*.repo. All photon repos are defaulted to gpgcheck enabled and test repo is non gpgchecked by using gpgcheck=0 in repo file.

sidcha:
@ppadmavilasom, @keerthanakalyan, @tapakund, @sshedi:

I was evaluating the options for having friendly names for return codes when keerthana brought it up earlier. After implementing the changes it looked like it adds some needless complexity and had to revert it. Here is my rational:

* If you are writing a new test case, it's far far easier to just put the number (as the tdnf error message prints the error code) than to go to tdnf source code and find the friendly name associated with the error code, add an alias to that name here and then use it. To me, tests must be as easy to write as possible.
* Ideally these error codes aren't likely to change. If they do change, in some PR, then that PR would fail CI builds and the author will invariably be forced to fix the tests too. So there is no burden from a maintenance perspective.
* Test code once written is not going to be frequently revisited (unlike regular software). So it doesn't need to have all the bells and whistles all the time. We can do the perl-moto and vote in favour "developer convenience".

pytests/conftest.py
keerthanakalyan: If you are trying to install a specific version, tdnf expects the version number which includes the photon version (ph1/ph2/ph3) or else you will get an error saying no matching package.
Ex: tdnf install tdnf-2.0.0-10.ph3 
If you remove .ph3 from the version string, tdnf reports an error.

pytests/conftest.py
sidcha: Got it. Thanks. When I go over the gpg section of tests (in another PR), I will handle this correctly.

pytests/conftest.py
sidcha: We pass only `package_name` and `package_version`. that would  try to install `package_name-package_version`. Is there a problem with this?

sidcha: @keerthanakalyan, Ohh I see, then in that case the version that we pass must have the `.ph1/.ph2/.ph3` in it.

tapakund: if you are in photon3 and all your repo is pointing to ph3, then if you just mention pkg-version, then it will do it, but will install the latest release for that version.

But it will be better to include ph1/ph2/ph3 with this.

root@tkundu-photon3 [ /WS/pywbem/pywbem-0.15.0 ]# tdnf install cloud-init-18.3

Downgrading:
cloud-init               noarch       18.3-6.ph3       photon-updates   3.56M 3732716

Total installed size:   3.56M 3732716
Is this ok [y/N]:y

Downloading:
cloud-init                             1084052    100%
Testing transaction
Running transaction
Installing/Updating: cloud-init-18.3-6.ph3.noarch
Removing: cloud-init-19.1-3.ph3.noarch

Complete!
root@tkundu-photon3 [ /WS/pywbem/pywbem-0.15.0 ]#

tapakund:
> @ppadmavilasom, @keerthanakalyan, @tapakund, @sshedi:
> 
> I was evaluating the options for having friendly names for return codes when keerthana brought it up earlier. After implementing the changes it looked like it adds some needless complexity and had to revert it. Here is my rational:
> 
> * If you are writing a new test case, it's far far easier to just put the number (as the tdnf error message prints the error code) than to go to tdnf source code and find the friendly name associated with the error code, add an alias to that name here and then use it. To me, tests must be as easy to write as possible.
> * Ideally these error codes aren't likely to change. If they do change, in some PR, then that PR would fail CI builds and the author will invariably be forced to fix the tests too. So there is no burden from a maintenance perspective.
> * Test code once written is not going to be frequently revisited (unlike regular software). So it doesn't need to have all the bells and whistles all the time. We can do the perl-moto and vote in favour "developer convenience".

If the error code is changed in tdnf source, the return value will be changed in test as well. So test will fail. if we use the defined name in tdnf then the chances for test to fail is lesser(people do not need to change tests) and also we will know what we are referring to with respect to the return value. Just my point of view.

pytests/conftest.py
sidcha: @tapakund: I get what you are saying. I think if we do need to mention the release, it can be part of the version string itself. `tdnf lists` command lists it that way.

sidcha:
> people do not need to change tests

It's very neat if test don't have to be changed with changes to the return code. Unfortunately this not possible.

The problem is, there is no way to export the C macro into python for synchronous consumption (there is one way but its very dirty, let's not get into it). The alternative it to have those macros defined as variables with the same names inside the TestUtils class and then use them. If we need to add/modify any return code, it must be added/modified in the TestUtils class as well.

tapakund (approved):

Conan-Kudo:
Uhh, you guys realize you just broke pretty much anybody packaging `tdnf`, right?

Two big problems:

1. pkgconfig files are gone (I'm working on PR to restore these)
2. Versioning is now exclusively set by Git (Tarballs fetched from GitHub will be broken, since there's no Git information in there)

@ppadmavilasom I'm honestly surprised here. These two things are pretty big and obvious breakages. They even break VMware Photon's own packaging for tdnf!

Conan-Kudo:
Oh and apparently this CMake config is broken for building and installing on Linux... PR incoming for fixing that...

sidcha:
@Conan-Kudo, sorry about the trouble. We are aware of both of those problems, please allow me to clarify.

1. pkg-config was deflected (not sure why) to a subsequent PR I was working on. I'd get to it ASAP but you can send the PR if you already have it.
2. Yes, the new version system doesn't play well with tarballs; we expect the build system to patch the version info (we haven't finalized the approach yet) in such cases. As an initial step, we only put place holders when built out of git-tree.

And, no we didn't break Photon's packaging of tndf as this is an un-released change. Of course the most recent dev branch cannot be built with old-spec file which is expecting the old build system. If you need consistent behaviour, don't build off the HEAD of dev branch; use a known-good-commit hash.

> CMake config is broken for building and installing on Linux

How so?

ppadmavilasom:
@Conan-Kudo once fixed, we will cover package dependencies and versioning with tests to prevent regressions. 
@cbsiddharth we should move cmake changes to a different branch till we address these

ppadmavilasom:
> Oh and apparently this CMake config is broken for building and installing on Linux... PR incoming for fixing that...

Please hold off on this. i am moving cmake changes to a "cmake" branch. we will cover it with tests for the issues @Conan-Kudo raised and merge when ready

Conan-Kudo:
@ppadmavilasom I've prepared a PR for this: #99.

Conan-Kudo:
> i am moving cmake changes to a "cmake" branch.

We shouldn't need to move to a separate branch anymore, as we're now back to parity pre-cmake.

ppadmavilasom:
> > i am moving cmake changes to a "cmake" branch.
> 
> We shouldn't need to move to a separate branch anymore, as we're now back to parity pre-cmake.

@Conan-Kudo Agree. Thanks for the quick PR and continued support.

Issue #95, opened 2019-12-10
Stream pointer might be NULL


QiAnXinCodeSafe:
'pCacheFile' is defined here, but may be NULL, please check 'if(pCacheFile== NULL)'
https://github.com/vmware/tdnf/blob/94955ca8e66eb17a98bc3bf2ced9fcac6f2db86a/solv/tdnfrepo.c#L253

PR #98, opened 2019-12-19
Fix issue: https://github.com/vmware/tdnf/issues/95

ppadmavilasom:


tapakund (approved):
LGTM

tapakund (commented):
Shall we set dwError explicitly to :
dwError = ENOENT

But i think setting to errno is better.

keerthanakalyan:
@ppadmavilasom : pCacheFile can be NULL, repo_add_rpmdb_reffp 
 handles both NULL and not NULL values https://github.com/openSUSE/libsolv/blob/a7cbcc3a5cfae21afc6bd95e3907da7b276c17bd/ext/repo_rpmdb.c#L1846. I hope this change is not required.

ppadmavilasom:
> @ppadmavilasom : pCacheFile can be NULL, repo_add_rpmdb_reffp
> handles both NULL and not NULL values https://github.com/openSUSE/libsolv/blob/a7cbcc3a5cfae21afc6bd95e3907da7b276c17bd/ext/repo_rpmdb.c#L1846. I hope this change is not required.

@keerthanakalyan i added a check so we wont error out when cache file not present. if cachefile is present and we still error, we should handle.

PR #99, opened 2020-01-02
Fix CMake build of TDNF

Conan-Kudo:
This pull request addresses the issues I discovered while updating packaging for TDNF to the latest revision in the dev branch.

The major changes are:

* CMake is now using `GNUInstallDirs` so that the correct target directories are used on various platforms
* Header files are now being installed so that other projects can link to TDNF
* Project versioning is now set through CMake so there's a proper fallback when Git data isn't available
* The pkgconfig files have been restored and are configured and installed by CMake
* Fix includes in Python binding sources so that it builds on a system that doesn't already have TDNF installed in the build environment
* CI has been reworked to add a Fedora-based build and build the Python module

This should fix all the issues I've discovered so far...


ppadmavilasom:
@Conan-Kudo thanks! appreciate your help

ppadmavilasom:
@Conan-Kudo if we add an rpm build and an isolated dependent installation based on the built rpm, we can identify the pkg-config issues. do you see other test gaps here?

Conan-Kudo:
@ppadmavilasom I think that should mostly cover it. If we add a build using Fedora in here, for example, I think we can make sure things don't break like this again.

ppadmavilasom:
> @ppadmavilasom I think that should mostly cover it. If we add a build using Fedora in here, for example, I think we can make sure things don't break like this again.

Sure. we will add a build using fedora in addition to the one that uses photon now (as well as adding a dependency test via rpm installs or  client code compile from a newly installed chroot)

Conan-Kudo:
@ppadmavilasom I've added a Fedora CI build of TDNF, though I have not added an RPM spec yet...

ci/Dockerfile.fedora
```diff
@@ -0,0 +1,12 @@
+FROM fedora:latest
+
+MAINTAINER csiddharth@vmware.com
+
+
+RUN dnf -y upgrade
```
ppadmavilasom: -q as well please

Conan-Kudo: Why do quiet? Wouldn't that make it look like the CI is doing nothing for a long period of time?

Conan-Kudo: Anyway, I added the quiet option.

ppadmavilasom (approved):
LGTM. Thanks for your help.

Conan-Kudo:
@ppadmavilasom This should be good to merge now. We have significantly more coverage than we did before, which is a major step up. There's certainly other improvements to make, but let's get this merged so that it doesn't keep getting bigger. 😄

PR #102, opened 2020-01-06
fix version getting overwritten as empty

ppadmavilasom:
- remove custom command which overwrites version to empty string


sidcha:
@ppadmavilasom, if you remove this, then build_meta_data.c will not be created, and get_version() depends on it. I will check this issue and get back to you. Don't merge this yet.

CMakeLists.txt
```diff
@@ -45,17 +45,6 @@ add_c_compiler_flag(-O0 DEBUG)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
-# Custom command produces build_meta_data.c, and a dummy output that's
-# not actually produced, in order to force Version.cmake to always be
-# re-run before the build
-ADD_CUSTOM_COMMAND(
-    OUTPUT  ${CMAKE_CURRENT_BINARY_DIR}/build_meta_data.c
-            ${CMAKE_CURRENT_BINARY_DIR}/_build_meta_data.c
-    COMMAND ${CMAKE_COMMAND} -P
```
sidcha, line 54: the `-P` flag invokes cmake in script mode and hence cannot modify parent scope variables. Can you describe the problem you are facing?

ppadmavilasom:
> @ppadmavilasom, if you remove this, then build_meta_data.c will not be created, and get_version() depends on it. I will check this issue and get back to you. Don't merge this yet.

it does get created - with this include: https://github.com/vmware/tdnf/blob/dev/CMakeLists.txt#L20

sidcha:
Include is called only during cmake time (not during builds). If there is a version bump after cmake was executed, then the build_meta_data.c file will not be updated and report old version information.

CMakeLists.txt
ppadmavilasom: please try tdnf --version with your changes. also examine build_metadata.c after the make and you will see that the version string is populated empty. 
why would there be a version change after cmake is executed in a ci run?

sidcha: > please try tdnf --version with your changes

Here is my output with the version output. This is soon after a cmake.
```
Scanning dependencies of target tdnf
[ 97%] Building C object CMakeFiles/tdnf.dir/tools/cli/main.c.o
[100%] Linking C executable bin/tdnf
[100%] Built target tdnf
root@photon-ea4f47d2e8d0 [ ~/tdnf/build ]# cat build_meta_data.c
const char *PACKAGE_NAME = "tdnf";
const char *PACKAGE_VERSION = "2.0.0";
const char *GIT_REV = "fe04cff+";
const char *GIT_TAG = "NONE";
const char *GIT_BRANCH = "dev";
const char *VERSION = "2.0.0";
root@photon-ea4f47d2e8d0 [ ~/tdnf/build ]# ./bin/tdnf --version
tdnf: 2.0.0
root@photon-ea4f47d2e8d0 [ ~/tdnf/build ]#
```

> why would there be a version change after cmake

It doesn't happen on CI runs, yes :-). But when we make dev-release artifacts (such as compiled binaries for debugging, sometime in future) from a recently tagged HEAD, this will be triggered. We can be careful and run cmake again before such releases, but it's to come up with a method to deal with that scenario too. Just trying to be careful here.

ppadmavilasom:
@cbsiddharth i created this PR #103 to demonstrate this as you could not repro. In my opinion, a simple version replace will do for what we need - like a config.h.in with defines

sidcha:
@ppadmavilasom, I see the real problem now. Travis does a shallow clone where the tags aren't fetched to speedup checkout. I will send PR which defaults to using the CMAKE_PROJECT_VERSION if this happens. I will also test this whole thing from a tarball so we are sure this doesn't regress again.

Btw, how did you hit this? your setup must have had all the tags and you should never have hit this problem. Just curious.

ppadmavilasom:
to be clear, i just want a config.h.in here with a #define replace. the other bits with git version and such can come in when we need it. 
i was just doing a docker build myself and i am sure i cloned the repo without any special cases.

CMakeLists.txt
```diff
@@ -17,13 +17,19 @@ include(GNUInstallDirs)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(AddCCompilerFlag)
-include(BuildMetaData)
```
sidcha, line 20: Also remove cmake/BuildMetaData.cmake now that it is stale. We can bring it back when needed.

CMakeLists.txt
```diff
@@ -17,13 +17,19 @@ include(GNUInstallDirs)
...
 
 if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release)
 endif()
 
+#make config.h with
+#PACKAGE_NAME and PACKAGE_VERSION defined
+configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/client/config.h.in"
+    "${CMAKE_CURRENT_SOURCE_DIR}/client/config.h"
```
sidcha: Please add @only for consistency with other uses.

sidcha (commented):
LGTM.

sidcha (approved):

PR #106, opened 2020-01-22
plugin support

ppadmavilasom:
plugin support
    - read conf file, read cmdline params.
    - read plugins (default to 0. this means plugins are disabled by default)
    - read pluginpath (default to /usr/lib/tdnf-plugins)
    - read pluginconfpath (default to /etc/tdnf/pluginconf.d)
    - added tests for all config options


common/setopt.c
```diff
@@ -175,3 +175,214 @@ GetOptionAndValue(
...
+       IsNullOrEmptyString(pszOptName) || !pnHasOpt)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = _TDNFGetCmdOpt(pArgs, pszOptName, &pOpt);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    *pnHasOpt = nHasOpt;
```
tapakund, line 242: This always assigns 0 to * pnHasOpt. Is that what is expected here?

client/plugins.c
```diff
@@ -0,0 +1,471 @@
...
+    dwError = TDNFGetOptWithDefault(
+                  pConf->pArgs, TDNF_CONF_KEY_PLUGIN_PATH,
+                  TDNF_DEFAULT_PLUGIN_CONF_PATH, &pConf->pszPath);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+cleanup:
+    return dwError;
+
+error:
+    TDNF_SAFE_FREE_MEMORY(pConf->pszConfPath);
```
tapakund, line 111: makes more sense to keep free memory calls in cleanup instead of error.

common/setopt.c
```diff
@@ -175,3 +175,214 @@ GetOptionAndValue(
...
+    uint32_t dwError = 0;
+    char *pszOptValue = NULL;
+
+    dwError = TDNFGetCmdOptValue(pArgs, pszOptName, &pszOptValue);
+    if (!IsNullOrEmptyString(pszDefault) &&
+        dwError == ERROR_TDNF_FILE_NOT_FOUND)
+    {
+        dwError = TDNFAllocateString(pszDefault, &pszOptValue);
+    }
+    BAIL_ON_TDNF_ERROR(dwError);
```
tapakund, line 312: this should be inside if condition.

common/setopt.c
```diff
@@ -175,3 +175,214 @@ GetOptionAndValue(
...
+    const char *pszOptName,
+    const char *pszDefault,
+    char **ppszOptValue
+    )
+{
+    uint32_t dwError = 0;
+    char *pszOptValue = NULL;
+
+    dwError = TDNFGetCmdOptValue(pArgs, pszOptName, &pszOptValue);
+    if (!IsNullOrEmptyString(pszDefault) &&
```
tapakund, line 307: we can add one more check if dwError is not File not found?

tapakund (commented):
Looks good to me. 
Have given some comments, please check if those makes sense to you.
Thanks.

client/plugins.c
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    dwError = TDNFGetOptWithDefault(
+                  pConf->pArgs, TDNF_CONF_KEY_PLUGIN_CONF_PATH,
+                  TDNF_DEFAULT_PLUGIN_CONF_PATH, &pConf->pszConfPath);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFGetOptWithDefault(
+                  pConf->pArgs, TDNF_CONF_KEY_PLUGIN_PATH,
+                  TDNF_DEFAULT_PLUGIN_CONF_PATH, &pConf->pszPath);
```
keerthanakalyan: Should this be  TDNF_DEFAULT_PLUGIN_PATH instead of TDNF_DEFAULT_PLUGIN_CONF_PATH?

client/plugins.c
```diff
@@ -0,0 +1,471 @@
...
+    {
+        dwError = 0;
+    }
+    goto cleanup;
+}
+
+static
+uint32_t
+_TDNFGetPluginSettings(
+    struct plugin_config *pConf
```
keerthanakalyan, line 76: NULL check for pConf

client/plugins.c
```diff
@@ -0,0 +1,471 @@
...
+    DIR *pDir = NULL;
+    struct dirent *pEnt = NULL;
+    int nLen = 0;
+    int nExtLen = TDNF_PLUGIN_CONF_EXT_LEN;
+    PTDNF_PLUGIN pPlugin = NULL;
+    PTDNF_PLUGIN pPlugins = NULL;
+    PTDNF_PLUGIN pLast = NULL;
+    char *pszPluginConfig = NULL;
+
+    pDir = opendir(pConf->pszConfPath);
```
keerthanakalyan, line 215: NULL check for pConf and pszConfPath

client/init.c
```diff
@@ -103,6 +103,15 @@ TDNFCloneCmdArgs(
                                    &pCmdArgs->pSetOpt);
         BAIL_ON_TDNF_ERROR(dwError);
     }
+    else /* if there are no setopt values, prime it to ensure non null */
+    {
+        dwError = AddSetOptWithValues(
```
keerthanakalyan, line 108: I could not understand the use of this dummy option and values. Could you please help me understand what would be the use of this?

client/plugins.c
```diff
@@ -0,0 +1,471 @@
...
+    }
+    dwError = 0; /* okay to proceed without any or all plugins */
+    goto cleanup;
+}
+
+static
+uint32_t
+_TDNFLoadPluginLibs(
+    const char *pszLibPath,
+    PTDNF_PLUGIN pPlugins
```
keerthanakalyan, line 417: NULL check for pPlugins

common/setopt.c
```diff
@@ -175,3 +175,214 @@ GetOptionAndValue(
...
+    }
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    *ppszOptValue = pszOptValue;
+error:
+    return dwError;
+}
+
+uint32_t
+TDNFSetOptNoOverwrite(
```
keerthanakalyan: Is this function used anywhere? Is this added for future use?

common/setopt.c
```diff
@@ -175,3 +175,214 @@ GetOptionAndValue(
...
+            !strcmp(pOpt->pszOptName, pszOptName))
+        {
+            nHasOpt = 1;
+            break;
+        }
+    }
+
+    if (!nHasOpt)
+    {
+        dwError = ERROR_TDNF_FILE_NOT_FOUND;
```
keerthanakalyan: Does this error "ERROR_TDNF_FILE_NOT_FOUND" mean a file not found or an option not found?

common/setopt.c
ppadmavilasom: Good catch. fixed.

client/plugins.c
ppadmavilasom: this is the error path.

common/setopt.c
ppadmavilasom: Even if there is a bail inside the if, there needs to be one out as well so just doing a common one.

client/plugins.c
ppadmavilasom: Good catch. Fixed.

client/init.c
ppadmavilasom: A clone will not happen if there were no setopt from command line. In this case, further operations with the internal setopt clone would fail. Hence priming it with dummy values

common/setopt.c
ppadmavilasom: Agree. It makes sense to not assign multiple meanings to the file error. Fixed.

common/setopt.c
ppadmavilasom: unused. removed.

common/setopt.c
ppadmavilasom: even though slightly cryptic, the current code path addresses the special case and other errors.

keerthanakalyan (approved):
LGTM.

tapakund (approved):

PR #110, opened 2020-02-07
Fix Error string when repodir is invalid

keerthanakalyan:
Fix error string when repo directory provided is invalid. Currently tdnf displays "Error(2605) : Unknown error 1005" when the repodir configured is invalid. Correct Error should be "Error(1005) : Error opening repo dir. Check if the repodir configured in tdnf.conf exists (usually /etc/yum.repos.d)"

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -50,7 +50,7 @@ TDNFLoadRepoData(
     if(pDir == NULL)
     {
         dwError = ERROR_TDNF_REPO_DIR_OPEN;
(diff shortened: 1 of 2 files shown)


ppadmavilasom (approved):
if you get precaution warnings(like the ones you have fore /tmp etc), please mark as #nosec as you will note elsewhere in test code where popen, assert etc are used. we are okay with doing this as it is completely within test code

ppadmavilasom (approved):

PR #111, opened 2020-02-12
return non zero error when interactive alter is aborted

ppadmavilasom:
- when --assumeno is provided in command line
- when user aborts interactive alter with input other than 'y'
- error message "Operation aborted"
- previously tdnf did not return an error. returns non-zero error now.


tapakund (approved):
LGTM

keerthanakalyan (approved):

PR #112, opened 2020-02-24
refactor cmake for v2.1.0

ppadmavilasom:
- fix issues while packaging for photon
 - fix library installs


sidcha (approved):
LGTM. The cmake rules for pytests could also move into pytests directory. I will send a subsequent PR for this. Thanks!

PR #114, opened 2020-02-25
Handle already installed packages faster.

sshedi:
At present, tdnf install -y 'package-which-is-already-installed' takes a bit of time to finish, this change makes it faster.

### DIFF
--- a/client/goal.c
+++ b/client/goal.c
@@ -312,7 +312,7 @@ TDNFGoal(
 
     int nFlags = 0;
     int i = 0;
-    Id  dwId = 0; 
+    Id  dwId = 0;
     int nProblems = 0;
     char** ppszExcludes = NULL;
     uint32_t dwCount = 0;
@@ -324,7 +324,7 @@ TDNFGoal(
     }
 
     queue_init(&queueJobs);
-    if(nAlterType == ALTER_UPGRADEALL)
+    if (nAlterType == ALTER_UPGRADEALL)
     {
         dwError = SolvAddUpgradeAllJob(&queueJobs);
         BAIL_ON_TDNF_ERROR(dwError);
@@ -336,7 +336,13 @@ TDNFGoal(
     }
     else
     {
-        for(i = 0; i < pQueuePkgList->count; i++)
+        if (pQueuePkgList->count == 0)
+        {
+            dwError = ERROR_TDNF_ALREADY_INSTALLED;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+
+        for (i = 0; i < pQueuePkgList->count; i++)
         {
             dwId = pQueuePkgList->elements[i];
             TDNFAddGoal(pTdnf, nAlterType, &queueJobs, dwId);
--- a/client/plugins.c
+++ b/client/plugins.c
@@ -510,7 +510,7 @@ cleanup:
     return dwError;
 
 error:
-    fprintf(stderr, "Error: %d dlerror: %s\n", dwError, dlerror());
+    fprintf(stderr, "Error: %u dlerror: %s\n", dwError, dlerror());
     if (pPlugin)
     {
         if (pPlugin->pModule)
--- a/client/resolve.c
+++ b/client/resolve.c
@@ -314,7 +314,8 @@ TDNFPrepareSinglePkg(
 
     //Check if this is a known package. If not add to unresolved
     dwError = SolvCountPkgByName(pSack, pszPkgName, &dwCount);
-    if(dwError || dwCount == 0)
+    BAIL_ON_TDNF_ERROR(dwError);
+    if (dwCount == 0)
     {
         dwError = ERROR_TDNF_NO_SEARCH_RESULTS;
         BAIL_ON_TDNF_ERROR(dwError);
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -44,12 +44,7 @@ TDNFRpmExecTransaction(
                   1,
                   sizeof(TDNF_CACHED_RPM_LIST),
                   (void**)&ts.pCachedRpmsArray);
-
-    if(!ts.pCachedRpmsArray)
-    {
-        dwError = ERROR_TDNF_OUT_OF_MEMORY;
-        BAIL_ON_TDNF_ERROR(dwError);
-    }
+    BAIL_ON_TDNF_ERROR(dwError);
 
     rpmSetVerbosity(TDNFConfGetRpmVerbosity(pTdnf));
 
--- a/solv/tdnfquery.c
+++ b/solv/tdnfquery.c
@@ -281,12 +281,9 @@ SolvApplyPackageFilter(
     }
 
     ppszTmpNames = ppszPackageNames;
-    while(*ppszTmpNames)
+    while (ppszTmpNames && *ppszTmpNames)
     {
-        if(!IsNullOrEmptyString(ppszTmpNames))
-        {
-            dwPkgs++;
-        }
+        dwPkgs++;
         ppszTmpNames++;
     }
 
--- a/tools/cli/lib/installcmd.c
+++ b/tools/cli/lib/installcmd.c
@@ -258,6 +258,11 @@ cleanup:
     return dwError;
 
 error:
+    if (dwError == ERROR_TDNF_ALREADY_INSTALLED)
+    {
+        dwError = ERROR_TDNF_CLI_NOTHING_TO_DO;
+    }
+
     goto cleanup;
 }
 
--- a/tools/cli/main.c
+++ b/tools/cli/main.c
@@ -134,6 +134,7 @@ int main(int argc, char* argv[])
 
                 dwError = arCmdMap[nCommandCount].pFnCmd(&_context, pCmdArgs);
                 BAIL_ON_CLI_ERROR(dwError);
+
                 break;
             }
         };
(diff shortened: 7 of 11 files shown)


tools/cli/lib/installcmd.c
```diff
@@ -191,6 +191,11 @@ TDNFCliAlterCommand(
                   pContext,
                   nAlterType,
                   &pSolvedPkgInfo);
+
+    if (!pSolvedPkgInfo) {
```
ppadmavilasom: you are changing the contract of the function here by sometimes returning a null without an error being set. please make a new error code and return it, then handle it here appropriately

sshedi: It's goto  cleanup; I used goto error; because it was doing nothing but going to cleanup & pSolvePkgInfo will be NULL only when package is already installed, so it's not an error condition, we can set dwError to 0.
I'm still facing few issues with the changes, I'll keep updating this PR

1.patch
```diff
@@ -0,0 +1,77 @@
+diff --git a/client/api.c b/client/api.c
```
sidcha: Accidental check in?

client/api.c
```diff
@@ -924,6 +924,14 @@ TDNFResolve(
                   &nAlterType,
                   ppszPkgsNotResolved,
                   &queueGoal);
+
+    if (dwError == ERROR_TDNF_ALREADY_INSTALLED) {
+        if (ppszPkgsNotResolved) {
+            TDNFFreeStringArray(ppszPkgsNotResolved);
+        }
+        dwError = 0;
+	    goto cleanup;
```
sidcha: Fix WS here.

client/resolve.c
```diff
@@ -121,8 +122,13 @@ TDNFPrepareAllPackages(
                           *pAlterType,
                           ppszPkgsNotResolved,
                           queueGoal);
+	    if (dwError != ERROR_TDNF_ALREADY_INSTALLED) {
+            if (!dwError && !newPkginHand) {
+                newPkginHand = true;
+            }
```
sidcha: WS error here too.

tags
```diff
@@ -0,0 +1,1387 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
sidcha: Remove this. If `tags` is not already part of `.gitignore`, add it. Also `*.patch` and `*.diff` so we don't check these in.

tools/cli/lib/installcmd.c
```diff
@@ -193,6 +193,11 @@ TDNFCliAlterCommand(
                   &pSolvedPkgInfo);
     BAIL_ON_CLI_ERROR(dwError);
 
+    if (!pSolvedPkgInfo) {
+        dwError = 0;
+	    goto cleanup;
```
sidcha: WS error.

sidcha (changes requested):
Some minor issues; mostly WS related. Make sure you have these in your vimrc.

```
set expandtabs
set tabstop=4
set shiftwidth=4
```

tags
sshedi: Sorry man, I saw 1.patch in my diff. It was late night. :) I'll remove this

tools/cli/lib/installcmd.c
sshedi: I'll fix the indent and tabs

ppadmavilasom:
I was wondering if something like this would be simpler. it seems this will achieve the early exit that you are after and also keep all the other flows intact. like `tdnf install openssl curl*` for eg.

diff --git a/client/goal.c b/client/goal.c
index 0f36e0b..51a628e 100644
--- a/client/goal.c
+++ b/client/goal.c
@@ -336,6 +336,16 @@ TDNFGoal(
     }
     else
     {
\+        /*
\+         * if pQueuePkgList is empty, it means we tried to resolve
\+         * and found that there were no action to take.
+
\+        */
\+        if (pQueuePkgList->count == 0)
\+        {
\+            dwError = ERROR_TDNF_ALREADY_INSTALLED;
\+            BAIL_ON_TDNF_ERROR(dwError);
\+        }
\+
         for(i = 0; i < pQueuePkgList->count; i++)
         {
             dwId = pQueuePkgList->elements[i];

sshedi:
Yes Priyesh, this is much better solution 👍 needs a little bit of modification one another place, we need to take care of return value in main function or else exit status of program will be wrong. I'll make the modification and push the change ASAP.

client/resolve.c
```diff
@@ -316,7 +316,7 @@ TDNFPrepareSinglePkg(
     dwError = SolvCountPkgByName(pSack, pszPkgName, &dwCount);
     if(dwError || dwCount == 0)
```
ppadmavilasom: ah. this was bad. thanks for finding it.
Can we please make it clear like
BAIL_ON_TDNF_ERROR(dwError);

if (dwCount == 0)
{
    dwError = ERROR_TDNF_NO_SEARCH_RESULTS;
    BAIL_ON_TDNF_ERROR(dwError);
}

client/goal.c
```diff
@@ -324,23 +324,18 @@ TDNFGoal(
...
+    if (nAlterType == ALTER_UPGRADEALL) {
         dwError = SolvAddUpgradeAllJob(&queueJobs);
-        BAIL_ON_TDNF_ERROR(dwError);
-    }
-    else if(nAlterType == ALTER_DISTRO_SYNC)
-    {
+    } else if(nAlterType == ALTER_DISTRO_SYNC) {
         dwError = SolvAddDistUpgradeJob(&queueJobs);
-        BAIL_ON_TDNF_ERROR(dwError);
+    } else if (pQueuePkgList->count == 0) {
```
ppadmavilasom: this does change the logic a bit from what it was. i would like to keep as is. 
what i mean is to do
else
{
    if (pQueuePkgList->count == 0)
    {
        dwError = ERROR_TDNF_ALREADY_INSTALLED;
        BAIL_ON_TDNF_ERROR(dwError);
    }
    for...
}

ppadmavilasom (commented):
in general, the tdnf code base starts opening braces in new lines
like 
if (..)
{
    ...
}

instead of 
if (..) {
}
I am sorry to be picky here but i would much rather have one style than a hybrid at this time.

client/resolve.c
sshedi: Sure, I'll do it.

client/goal.c
sshedi: AFAIK, this doesn't change the logic flow. Can you please explain what is the logical error here? I'm not able to find it out

ppadmavilasom: Lets say nAlterType was ALTER_UPGRADEALL for eg. In the previous version of the code, it would have never attempted the pQueuePkgList iteration and add to goal. Now it will regardless of what nAlterType is. That is a logic change.
From the context of this function, You do not control pQueuePkgList contents as it is an argument to this function. Therefore, the future behavior is also altered (inadvertently) if pQueuePkgList has contents.

sshedi: Sigh, that's so obvious, sorry :(

ppadmavilasom: Nothing to be sorry about. Appreciate your help

client/plugins.c
```diff
@@ -510,7 +510,7 @@ _TDNFLoadPluginLib(
     return dwError;
 
 error:
-    fprintf(stderr, "Error: %d dlerror: %s\n", dwError, dlerror());
+    fprintf(stderr, "Error: %u dlerror: %s\n", dwError, dlerror());
```
sshedi, line 513: Some formatting fixes

client/rpmtrans.c
```diff
@@ -44,6 +44,7 @@ TDNFRpmExecTransaction(
                   1,
                   sizeof(TDNF_CACHED_RPM_LIST),
                   (void**)&ts.pCachedRpmsArray);
+    BAIL_ON_TDNF_ERROR(dwError);
```
sshedi, line 47: Missing error check, I believe next if condition is redundtant, please have a look.

solv/tdnfquery.c
```diff
@@ -281,12 +281,9 @@ SolvApplyPackageFilter(
     }
 
     ppszTmpNames = ppszPackageNames;
-    while(*ppszTmpNames)
+    while (ppszTmpNames && *ppszTmpNames)
```
sshedi, line 284: Redundant if inside while?

client/rpmtrans.c
ppadmavilasom: yes. please remove if. bail (that you added) is enough.

sshedi: This is done, please have a look once you're free, thanks for the quick response and reviews :) It's almost 1.30AM here, feeling sleepy, if any modifications needed, I'll do it tomorrow.

tools/cli/main.c
```diff
@@ -161,11 +162,14 @@ int main(int argc, char* argv[])
 
 error:
     TDNFCliPrintError(dwError);
-    if (dwError == ERROR_TDNF_CLI_NOTHING_TO_DO || dwError == ERROR_TDNF_NO_DATA)
+    if (dwError == ERROR_TDNF_CLI_NOTHING_TO_DO ||
+        dwError == ERROR_TDNF_NO_DATA ||
+        dwError == ERROR_TDNF_ALREADY_INSTALLED)
```
ppadmavilasom: instead of adding ERROR_TDNF_ALREADY_INSTALLED here, can we just reassign it to ERROR_TDNF_CLI_NOTHING_TO_DO from here: https://github.com/vmware/tdnf/blob/dev/tools/cli/lib/installcmd.c#L260
That will also mean not touching main.c and also not having the artificial mapping for TDNF_ALREADY_INSTALLED to "Nothing to do" in the defines file

PR #118, opened 2020-03-17
Fix tdnf rpm cache clearance issue when installation fails

### LINKED ISSUE #117: tdnf does not clean downloaded RPM cache when there is a installation failure
Refer this https://github.com/vmware/photon/issues/988#issuecomment-599505221 
When the previous installation command downloads corrupted or partial RPM file due to network issue, the subsequent installation of the same RPM fails with error Error(1471) : rpm generic error - not found. This is because the partially downloaded / corrupted RPM file is not cleared when there is a failure.

keerthanakalyan:
fixes #117

Signed-off-by: Keerthana K <keerthanak@vmware.com>

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -400,6 +400,17 @@ TDNFTransAddInstallPkg(
     char* pszDownloadCacheDir = NULL;
     char* pszUrlGPGKey = NULL;
     PTDNF_CACHED_RPM_ENTRY pRpmCache = NULL;
+    int nDownloadFile = 1;
+
+    //Check override, then repo config and launch
+    //gpg check if needed
+    dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, &pszUrlGPGKey);
+    BAIL_ON_TDNF_ERROR(dwError);
+    if(nGPGSigCheck)
+    {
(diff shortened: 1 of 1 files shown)


client/rpmtrans.c
```diff
@@ -534,6 +534,10 @@ TDNFTransAddInstallPkg(
 
 error:
     fprintf(stderr, "Error processing package: %s\n", pszPackageLocation);
+    if (!access(pszFilePath, F_OK))
```
ppadmavilasom: fix should be earlier than this. ideally during download or when download gets skipped as file exists. note that "error processing package" was not in error reports.

keerthanakalyan: From the trace, it appears that rpmReadPackageFile() call failed from TDNFTransAddInstallPkg(). Any failures from TDNFTransAddInstallPkg() API has to print "error processing package" because that's the entry point to error:. I'm wondering how this string is not printed in the error message. 
If this fix has to go earlier during the download, do you suggest to check for file stats to check the size of the file and re-download the file if file size is zero instead of failing?

ppadmavilasom: file size zero check is one of the cases. however, a complete check would be to open and check the digest. then open will fail or digest will fail if file is partially downloaded.

PR #120, opened 2020-03-24
fix error in distroverpkg lookup

ppadmavilasom:
- change RPMTAG_NAME to RPMTAG_PROVIDES
- required as system-release is provided by photon-release


sshedi (approved):
Looks good to me

keerthanakalyan (approved):

PR #121, opened 2020-03-30
Add an extra check to check the providers

sshedi:
tdnf check currently has options to skip conflicts & obsoletes.
However it reports error on providers even though the required packages
are available. This will fix that issue by adding an extra check on
providers.

### DIFF
--- a/pytests/tests/test_check_skip.py
+++ b/pytests/tests/test_check_skip.py
@@ -1,0 +1,62 @@
+#
+# Copyright (C) 2019 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the GNU General Public License v2 (the "License");
+# you may not use this file except in compliance with the License. The terms
+# of the License are located in the COPYING file of this distribution.
+#
+#   Author: Shreenidhi Shedi <sshedi@vmware.com>
+
+import os
+import tempfile
+import pytest
+import shutil
+import subprocess
+
+isNotPhoton = -1
+try:
+    isNotPhoton = subprocess.check_call("uname -a | grep -i photon > /dev/null", shell = True)
+except Exception as e:
+    isNotPhoton = 1
+
+def run_cmd(utils, cmd, retval):
+    ret = utils.run(cmd)
+    if retval == 0:
+        assert(ret['retval'] == 0)
+    else:
+        assert(ret['retval'] != 0)
+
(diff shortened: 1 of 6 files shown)


solv/tdnfpackage.c
```diff
@@ -1518,134 +1518,162 @@ SolvSplitEvr(
  *      1 : if solver problem type and user specified problem matches
  *      0 : if not matches
  */
-static uint32_t
+static bool
 __should_skip(
```
ppadmavilasom: please follow conventions elsewhere in the codebase for function names.

ppadmavilasom (commented):
Looks good. Can you please add a test case for this?

solv/tdnfpackage.c
```diff
@@ -1518,134 +1518,162 @@ SolvSplitEvr(
...
         {
-            dwProblemId = solver_findproblemrule(pSolv, i);
-            type = solver_ruleinfo(
-                       pSolv,
-                       dwProblemId,
-                       &dwSource,&dwTarget,
-                       &dwDep);
-
-            if (__should_skip(type, dwSkipProblem))
+            if (!check_for_providers(pSack, type, pszProblem, prv_pkgname))
```
ppadmavilasom, line 1668: simple check for providers might satisfy but does it then satisfy the goal? i was thinking about isolating the problem reports and trying to reach the goal independently for each.

sshedi: IMO, satisyfing the goal is a different problem, here user is trying to know is there any problem with the packages, right?

sshedi:
> Looks good. Can you please add a test case for this?

In progress

sshedi:
> > Looks good. Can you please add a test case for this?
> 
> In progress

Done

pytests/tests/test_check_skip.py
```diff
@@ -0,0 +1,62 @@
...
+@pytest.fixture(scope='module', autouse=True)
+def setup_test(utils):
+    yield
+    teardown_test(utils)
+
+def teardown_test(utils):
+    pass
+
+def test_check_skipconflicts(utils):
+    cmd = [ 'tdnf', '--config', '/etc/tdnf/tdnf.conf', 'check', '--skipconflicts']
```
sidcha, line 38: @sshedi, why are you referring to system config instead of test-repo config here?

PR #122, opened 2020-03-30
Take backup & use bind mount while doing kernel upgrade

sshedi:
This addresses: https://jira.eng.vmware.com/browse/PHO-951


ppadmavilasom:
Maybe this is better as a plugin so that photon can add the dependency in spec as it finds fit. Package specific behavior changes does not fit in generic tdnf. Thoughts?

YustasSwamp:
I believe it should be part of kernel specs.

ppadmavilasom:
I agree with @YustasSwamp

PR #124, opened 2020-04-01
Plugin to move running kernel while doing kernel upgrade

sshedi:
This plugin moves the running kernel to /tmp and does a bind mount of
the moved kernel under /lib/modules so that we won't lose the ability to
do a depmod/modprobe after upgrading the kernel.


ppadmavilasom:
@sshedi i think there was a misunderstanding in the previous pull requests comments. sorry about that. we need to tackle this at the corresponding spec file rather than a tdnf mod.

ppadmavilasom:
we have useful fixes in this PR. for event mask, the events required query that was missing and error string fix. please change to send those if you will

PR #126, opened 2020-04-06
Add metalink support

tapakund:
Added metalink support for repo files.

If metalink URL is present, we will get the URL with the highest priority.
Then we concat the metalink path to the url and download the metalink file.

Once metalink file is downloaded, we parse it and downlod the repomd.xml
file and then parse the xml to download repoMD parts.

Also, we set the BaseURL for the respective repo to the URL we got from
the metalink file for further package downloads.

### DIFF
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -106,14 +106,108 @@ error:
     }
     return dwError;
 }
+uint32_t
+TDNFParseAndGetURLFromMetalink(
+    PTDNF pTdnf,
+    const char *pszRepo,
+    const char *pszFile
+    )
+{
+    metalink_error_t metalink_error;
+    metalink_t* metalink = NULL;
+    metalink_file_t **files;
+    metalink_parser_context_t *metalink_context = NULL;
+    metalink_resource_t** resources;
+    char buf[BUFSIZ] = {0};
+    int length = 0;
+    int fd = -1;
+    uint32_t dwError = 0;
 
+    if(!pTdnf ||
+       !pTdnf->pArgs ||
+       IsNullOrEmptyString(pszRepo) ||
+       IsNullOrEmptyString(pszFile))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    metalink_context = metalink_parser_context_new();
+    if(metalink_context == NULL)
+    {
+        dwError = ERROR_TDNF_OUT_OF_MEMORY;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    fd = open(pszFile, O_RDONLY);
+    if(fd == -1)
+    {
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+    while((length = read(fd, buf, (sizeof(buf)-1))) > 0)
+    {
+        metalink_error = metalink_parse_update(metalink_context, buf, length);
+        memset(buf, 0, BUFSIZ);
+        if(metalink_error != 0)
+        {
+            if(metalink_context)
+            {
+                metalink_parser_context_delete(metalink_context);
+            }
+            fprintf(stderr, "Unable to parse metalink, ERROR: code=%d\n", metalink_error);
+            dwError = ERROR_TDNF_INVALID_PARAMETER;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+    }
+    if(length == -1)
+    {
+        if(metalink_context)
+        {
+            metalink_parser_context_delete(metalink_context);
+        }
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+    metalink_error = metalink_parse_final(metalink_context, NULL, 0, &metalink);
+    if((metalink_error != 0) || (metalink == NULL))
+    {
+        fprintf(stderr, "metalink_parse_final failed, ERROR: code=%d\n", metalink_error);
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    for(files = metalink->files; files && *files; ++files)
+    {
+        resources = (*files)->resources;
+        if(!resources)
+        {
+            continue;
+        }
+        dwError = TDNFRepoSetBaseUrl(pTdnf, pszRepo, (*resources)->url);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+cleanup:
+    if(fd != -1)
+    {
+        close(fd);
+    }
+    /* delete metalink_t */
+    if(metalink)
+    {
+        metalink_delete(metalink);
+    }
+    return dwError;
+error:
+    goto cleanup;
+}
 uint32_t
 TDNFDownloadFile(
     PTDNF pTdnf,
     const char *pszRepo,
     const char *pszFileUrl,
     const char *pszFile,
-    const char *pszProgressData
+    const char *pszProgressData,
+    int is_metalink
     )
 {
     uint32_t dwError = 0;
@@ -126,7 +220,9 @@ TDNFDownloadFile(
     if(!pTdnf ||
        !pTdnf->pArgs ||
        IsNullOrEmptyString(pszFileUrl) ||
-       IsNullOrEmptyString(pszFile))
+       IsNullOrEmptyString(pszFile) ||
+       IsNullOrEmptyString(pszRepo) ||
+       IsNullOrEmptyString(pszProgressData))
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
@@ -138,6 +234,7 @@ TDNFDownloadFile(
         dwError = ERROR_TDNF_CURL_INIT;
         BAIL_ON_TDNF_ERROR(dwError);
     }
+
     /* if callback present for extra curl options */
     dwError = _handle_curl_cb(pTdnf, pCurl, pszFileUrl);
     BAIL_ON_TDNF_ERROR(dwError);
@@ -200,6 +297,19 @@ TDNFDownloadFile(
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
+    else
+    {
+        if(is_metalink)
+        {
+            if(fp)
+            {
+                fclose(fp);
+                fp = NULL;
+            }
+            dwError = TDNFParseAndGetURLFromMetalink(pTdnf, pszRepo, pszFile);
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+    }
 cleanup:
     TDNF_SAFE_FREE_MEMORY(pszUserPass);
     if(fp)
@@ -249,6 +359,7 @@ TDNFDownloadPackage(
     char *pszPackageUrl = NULL;
     char *pszPackageFile = NULL;
     char *pszCopyOfPackageLocation = NULL;
+    int metalink = 0;
 
     if(!pTdnf ||
        !pTdnf->pArgs ||
@@ -274,7 +385,7 @@ TDNFDownloadPackage(
 
     dwError = TDNFAllocateString(pszPackageLocation,
                                  &pszCopyOfPackageLocation);
-        BAIL_ON_TDNF_ERROR(dwError);
+    BAIL_ON_TDNF_ERROR(dwError);
 
     dwError = TDNFAllocateStringPrintf(&pszPackageFile,
                                        "%s/%s",
@@ -286,7 +397,8 @@ TDNFDownloadPackage(
                                pszRepoName,
                                pszPackageUrl,
                                pszPackageFile,
-                               nSilent ? NULL : pszPkgName);
+                               nSilent ? NULL : pszPkgName,
+                               metalink);
     BAIL_ON_TDNF_ERROR(dwError);
 
     if(!nSilent)
--- a/client/repo.c
+++ b/client/repo.c
@@ -477,6 +477,9 @@ TDNFGetRepoMD(
     char *pszRepoMDUrl = NULL;
     char *pszTmpRepoDataDir = NULL;
     char *pszTmpRepoMDFile = NULL;
+    char *pszMetaLinkFile = NULL;
+    char *pszRepoMetalink = NULL;
+    char *pszTmpRepoMetalinkFile = NULL;
     PTDNF_REPO_METADATA pRepoMDRel = NULL;
     PTDNF_REPO_METADATA pRepoMD = NULL;
     unsigned char pszCookie[SOLV_COOKIE_LEN] = {0};
@@ -484,6 +487,7 @@ TDNFGetRepoMD(
     int nNeedDownload = 0;
     int nNewRepoMDFile = 0;
     int nReplaceRepoMD = 0;
+    int metalink = 0;
 
     if(!pTdnf ||
        !pRepoData ||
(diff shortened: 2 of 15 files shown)


tapakund:
I am checking the tests.

tapakund:
cc @ppadmavilasom  @keerthanakalyan

client/remoterepo.c
```diff
@@ -122,6 +123,14 @@ TDNFDownloadFile(
     char *pszUserPass = NULL;
     /* lStatus reads CURLINFO_RESPONSE_CODE. Must be long */
     long lStatus = 0;
+    metalink_error_t r;
```
ppadmavilasom: please use meaningful variable names. metalink_error for eg.

client/remoterepo.c
```diff
@@ -122,6 +123,14 @@ TDNFDownloadFile(
...
+    metalink_t* metalink = NULL;
+    metalink_file_t **files;
+    metalink_parser_context_t *metalink_parser = NULL;
+    metalink_resource_t** resources;
+    char buf[BUFSIZ] = {0};
+    int length = 0;
+    int fd = 0;
```
ppadmavilasom: initialize to -1

client/remoterepo.c
```diff
@@ -200,12 +220,60 @@ TDNFDownloadFile(
...
+            metalink_delete(metalink);
+        }
+    }     
 cleanup:
     TDNF_SAFE_FREE_MEMORY(pszUserPass);
     if(fp)
     {
         fclose(fp);
     }
+    if(fd > 0)
```
ppadmavilasom: check for >=0

client/remoterepo.c
```diff
@@ -200,12 +220,60 @@ TDNFDownloadFile(
...
+    {
```
ppadmavilasom, line 274: set to -1 after close

client/remoterepo.c
```diff
@@ -218,6 +286,10 @@ TDNFDownloadFile(
         fclose(fp);
         fp = NULL;
     }
+    if(fd > 0)
```
ppadmavilasom: same as above

ppadmavilasom:
please add test case and cover when baseurl and metalink set, baseurl commented and metalink set, baseurl and metalink not set. please have this test self contained and not rely on external repos

client/remoterepo.c
sshedi: Priyesh, will open syacall ever return 0? I believe 0,1, 2 are for stdin, stout and stderr respectively. If we close fd 0,then we will be closing stdin right?

ppadmavilasom: open returns -1 on error. note that this code had set it to 0 as initial value. if the code above sets the correct initial value and operate on values per man, it makes for better code. concerns about closing 0 and such is not something this code need not worry about as it is clearly doing https://github.com/vmware/tdnf/pull/126/files/676df2608a8be16972cd7f8fd7e9258251d1b9a0#diff-a7c98e30fa73f96ee19509a659602e53R232 and such

tapakund:
> please add test case and cover when baseurl and metalink set, baseurl commented and metalink set, baseurl and metalink not set. please have this test self contained and not rely on external repos

In Progress.

client/remoterepo.c
sidcha: @sshedi, When fd is initialized to -1, then a check for `if (fd != -1) { close(fd); }` (in that case, a check for`fd > 0` would also work) is the best way to go about this. Any other construct, IMO, will confuse newcomers to the codebase, in spite of being **not** wrong. And yes, open will always return a number larger than 2 (unless someone closes FDs 0, 1 or 2 on purpose). 

@ppadmavilasom, I propose to drop the idea to check `fd >=0`; initialize `fd = -1`; and check for `fd != -1` to close. Thoughts?

ppadmavilasom: @cbsiddharth fd != -1 is good. i am okay with this

tapakund: @cbsiddharth @ppadmavilasom  Agree with this.

client/remoterepo.c
tapakund: Done, Thanks!

client/remoterepo.c
sshedi: I agree with first solution given by Priyesh, init fd to -1 and then check for (fd >= 0).
Reason why I say (fd != -1) is not so good is, we have to consider the possibility that in future open call may return other negative values than -1 and confused new comers will slowly come up to speed.

sidcha: @sshedi, its POSIX-ed (read as, written on stone) that open errors to -1. Also, fd of 0 being STDIN is the usual case, that not strictly true always (think of daemon processes). What if we closed FD 0 elsewhere (by design, hypothetically) and then when you call another open, then FD 0 will be be allocated to you **legally**. In such cases you are expected to call a close on FD 0 too.

Rule of thumb, if open didn't fail, the kernel allocated something for you by **whatever** FD it returned and you must call close on it. Even if FD == 0.

sshedi: I agree with the latter part of closing fd 0 and I mentioned it my previous comment
{(fd >=0) while closing}
But open errors to -1 and written on stone, I don't agree with it. It's software and can be changed anytime. TBH, something may replace POSIX entirely, like how POSIX replaced SystemV. So, nothing is written on stone in science.

sidcha: Well, I can argue against that :-). But, not without breaking most serious software in the world.

sshedi: Let's discuss further on this in person at office :)

client/repo.c
```diff
@@ -489,18 +493,23 @@ TDNFGetRepoMD(
...
-                                      TDNF_REPO_METADATA_FILE_PATH);
+    dwError = TDNFAllocateStringPrintf(&pszRepoMDFile,
+                                       "%s/%s",
+                                       pszRepoDataDir,
+                                       TDNF_REPO_METADATA_FILE_NAME);
     BAIL_ON_TDNF_ERROR(dwError);
-
+    if(metalink)
+    {
+        dwError = TDNFAllocateStringPrintf(&pszMetaLinkFile,
```
keerthanakalyan: TDNF_SAFE_FREE_MEMORY(pszMetaLinkFile) is missed in the cleanup.

client/repo.c
```diff
@@ -537,6 +546,14 @@ TDNFGetRepoMD(
         BAIL_ON_TDNF_ERROR(dwError);
         nNeedDownload = 1;
     }
+    if(metalink)
+    {
+        dwError = TDNFAllocateStringPrintf(&pszRepoMetalink,
```
keerthanakalyan, line 599: TDNF_SAFE_FREE_MEMORY(pszRepoMetalink) is missed in cleanup.

client/repo.c
```diff
@@ -566,13 +583,38 @@ TDNFGetRepoMD(
                       pszTmpRepoDataDir,
                       TDNF_REPO_METADATA_FILE_NAME);
         BAIL_ON_TDNF_ERROR(dwError);
-
+        if(metalink)
+        {
+            dwError = TDNFAllocateStringPrintf(
```
keerthanakalyan, line 636: TDNF_SAFE_FREE_MEMORY(pszTmpRepoMetalinkFile) is missed in cleanup.

client/repoutils.c
```diff
@@ -39,6 +39,105 @@ TDNFRepoMakeCacheDirs(
     goto cleanup;
 }
 
+uint32_t
+TDNFRepoGetMetalink(
```
keerthanakalyan: Is this function used anywhere? I could not find its usage.

client/repoutils.c
```diff
@@ -39,6 +39,105 @@ TDNFRepoMakeCacheDirs(
...
+error:
+    TDNF_SAFE_FREE_MEMORY(pszMetalink);
+    goto cleanup;
+
+}
+uint32_t
+TDNFRepoSetBaseUrl(
+    PTDNF pTdnf,
+    const char* pszRepo,
+    char* ppszBaseUrl
```
keerthanakalyan: Is this a double pointer? Naming says ppszBaseUrl but looks to be a char pointer.

client/repoutils.c
```diff
@@ -39,6 +39,105 @@ TDNFRepoMakeCacheDirs(
...
+    )
+{
+    uint32_t dwError = 0;
+    PTDNF_REPO_DATA_INTERNAL pRepos = NULL;
+
+    if(!pTdnf || IsNullOrEmptyString(pszRepo) || !ppszBaseUrl)
```
keerthanakalyan: Does this !ppszBaseUrl needs a change to IsNullOrEmptyString(pszRepo)?

client/repoutils.c
```diff
@@ -77,6 +176,12 @@ TDNFRepoGetBaseUrl(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    if(!pRepos->pszBaseUrl)
```
keerthanakalyan: Is this check needed here, do we ever hit a NULL baseurl here? I believe this has to be checked when we retrieve the baseUrl from repo file or from metalink file right?

keerthanakalyan (commented):
Comments for the current changes. I had offline chat regarding the --refresh marker changes for metalink file similar to repomd.xml which Tapas is adding now.

client/remoterepo.c
tapakund: Done. Thanks!

client/remoterepo.c
tapakund: Done. Thanks!

client/repo.c
tapakund: Yeah, Thanks!

client/repo.c
tapakund: Yeah, Thanks!

client/repo.c
tapakund: Yeah, Thanks!

client/repoutils.c
tapakund: Yeah, not used. But i was keeping it for future use.. but now removed it.. Lets add when required.

client/repoutils.c
tapakund: Yeah, better to rename. Thanks!

client/repoutils.c
tapakund: No, its a pointer we get from metalink resource. So i think NULL check should suffice.

client/repoutils.c
tapakund: We will not, added one condition in MDFileDownload.. Removed this.

tapakund:
> I am checking the tests.

Wrote the tests, for 4 conditions.

1. Keep both metalink and baseurl.
2. Keep metalink and no commented baseurl.
3. Keep baseurl and commented metalink.
4. Commented both baseurl and metalink.

After setting each conditions, we will do , "tdnf repolist" and "install and uninstall pkg". ( I have not pushed the changes for tests as it will fail).

But for executing the tests, I need a binary which will generate the metalink file based on file input. As repomd.xml file in test is generated using createrepo and the shasum might differ.  I see a github proj on that "https://github.com/metalink-dev/command". Checking other possibilities.

The one I am using for test is written manually.

Also few things which i am planning to do in next PR:
1. Add ability to check if the URL with highest preference is reachable or not, else return the next reachable URL with next highest preference.
2. Process multiple files in the metalink.

client/remoterepo.c
ppadmavilasom: :)

ppadmavilasom:
@tapakund we do tests here: https://github.com/vmware/tdnf/tree/dev/pytests/tests
can you please create your tests in this PR following the examples above? @cbsiddharth can help if you need pointers for writing tests.
manual metalink file should be fine. we dont need extra deps for generation.

tapakund:
> @tapakund we do tests here: https://github.com/vmware/tdnf/tree/dev/pytests/tests
> can you please create your tests in this PR following the examples above? @cbsiddharth can help if you need pointers for writing tests.
> manual metalink file should be fine. we dont need extra deps for generation.

Done. Thanks.

tapakund:
> Can't we compare with a integer here, comparing with string is kinda makes it clumsy don't you think?

If we have multiple repos, and we don't have baseurl and metalink for a particular repo. By default tdnf will disable that repo and traverse towards other repos. So, i thought its better to check with stderr. Also the error we set in will not be reflected out as dwError in the parent function will be reset to 0 while parsing next repo.

client/remoterepo.c
```diff
@@ -106,14 +106,94 @@ _handle_curl_cb(
...
+    goto cleanup;
+}
 uint32_t
 TDNFDownloadFile(
     PTDNF pTdnf,
     const char *pszRepo,
     const char *pszFileUrl,
     const char *pszFile,
-    const char *pszProgressData
+    const char *pszProgressData,
```
sshedi, line 209: Check for null for this parameter missing

client/remoterepo.c
```diff
@@ -200,6 +281,19 @@ TDNFDownloadFile(
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
+    else
+    {
+        if(is_metalink)
+        {
+            if(fp)
+            {
+                fclose(fp);
```
sshedi, line 306: Cleanup is doing it anyway, we need this?

client/repo.c
```diff
@@ -494,18 +498,30 @@ TDNFGetRepoMD(
...
+    if((pRepoData->pszBaseUrl == NULL) && (pRepoData->pszMetaLink == NULL))
+    {
+        fprintf(stderr, "Error: Cannot find a valid base URL for repo: %s\n", pRepoData->pszName);
+        dwError = ERROR_TDNF_BASEURL_DOES_NOT_EXISTS;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if(pRepoData->pszMetaLink)
+    {
+        metalink = 1;
```
sshedi, line 510: Can we avoid if condition by using below code, your wish :)
metalink = !!pRepoData->pszMetaLink;

client/remoterepo.c
```diff
@@ -106,14 +106,94 @@ _handle_curl_cb(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    fd = open(pszFile, O_RDONLY);
+    if(fd == -1)
+    {
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+    while((length = read(fd, buf, sizeof(buf))) != 0)
```
sshedi: read call ret val validation is wrong, use while((length = read(fd, buf, sizeof(buf))) > 0)

From man page
On error, -1 is returned, and errno is set appropriately.  In this case, it is left unspecified whether the file position (if any) changes.

sshedi: Also, this is just a hunch, buf is of size BUFSIZ which 1024 I guess, if pszFile has more than BUFSIZ characters, read won't null terminate buf. So it will crash, you need to revisit this part of code.
Read sizeof(buf)-1 bytes every time and do a memset buf to 0 before every iteration is my suggestion.

sshedi (changes requested):

client/remoterepo.c
tapakund: We are opening the same file for reading metalink in read mode. So closing this before call.

client/remoterepo.c
tapakund: already made > 0 ,pls check latest change :)

client/remoterepo.c
sshedi: I think there is no need of it, we can open again without closing it

client/remoterepo.c
tapakund: Added. Actually this was missing from before.

client/remoterepo.c
tapakund: Yes, but both will be pointing to same file. There is no problem as such but preferred to close the handle as its not required anymore. 

If metalink is not enabled, cleanup will take care of this.

client/repo.c
tapakund: I think this is better in means of readability. So, lets keep it like that.

sshedi (approved):

sidcha (approved):
LGTM.

ppadmavilasom (approved):

PR #129, opened 2020-04-13
Improvement for metalink support

tapakund:
-Added checks for files and resources.
-Added verifying checksum for the files to be downloaded using metalink file
-If verifying checksum fails, we will fail the discard the download.
-Added checks for resource type to:
  - http
  - https
  - ftp
  - ftps
  - rsync


tapakund:
cc @ppadmavilasom @keerthanakalyan @cbsiddharth @sshedi

ci/Dockerfile.photon
```diff
@@ -8,7 +8,8 @@ RUN tdnf install -q -y --enablerepo=photon-debuginfo \
                        build-essential cmake curl-devel rpm-build libsolv-devel \
                        popt-devel sed createrepo_c glib libxml2 findutils \
                        python3 python3-pip python3-setuptools python3-devel \
-                       valgrind gpgme-devel glibc-debuginfo libmetalink-devel
+                       valgrind gpgme-devel glibc-debuginfo libmetalink-devel \
```
sshedi, line 11: Tapas, I'm not sure about the feasibility of this request but can we have a single config file with all these package lists sorted alphabetically? This list is really growing and need to be handled better. Also please put a comment in the common file mentioning that items need to be added alphabetically.

client/includes.h
```diff
@@ -68,6 +68,13 @@
 #include "prototypes.h"
 
 #include <metalink/metalink_parser.h>
+#include <openssl/sha.h>
+#include <openssl/md5.h>
+typedef struct metalink_digest_alias{
```
sshedi: New line after includes before typedef

client/includes.h
```diff
@@ -68,6 +68,13 @@
...
+    const char *alias_name;
+    uint32_t (*func)();
```
sshedi: Use appropriate name for function pointer

client/prototypes.h
```diff
@@ -163,11 +170,61 @@ TDNFRepoApplyProxySettings(
...
+    );
+
+uint32_t
+TDNF_New_Metalink_Checksum_From_Hex_Digest
+    (const char *hex_digest,
+    char **ppdigest
+    );
+
+char
+TDNF_HexToUint
```
sshedi: Some style mismatch here?

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+        close(fd);
+    }
+    return dwError;
+error:
+    goto cleanup;
+}
+
+
+uint32_t
+TDNFGetSHA1Sum(
```
sshedi: There is a lot of code repetition in above 3 functions, we can remove it.
Create a function
```c
TDNFGetHASH(filname, ppresult_hash, enum hash_type)
{
    open file
    calloc to file size(assuming that file will be few KBs)

    switch (hash_type) {
        hash_func(buf, file size);
    }
}

File: client/remoterepo.c [hunk 6]
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+{
+    uint32_t dwError = 0;
+    if(IsNullOrEmptyString(pszFile) ||
+       !ml_file)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    fprintf(stdout, "Validating metalink (%s)...\n", pszFile);
```
sshedi: Use printf, fprintf(stdout, ...) and printf are same

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+        fprintf(stderr,
+                "Validating metalink (%s) FAILED (digest missing)\n", pszFile);
+        dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    dwError = TDNFCheckHash(pszFile, ml_file->digest, ml_file->type);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+cleanup:
+    if(ml_file->digest)
```
sshedi: Check for NULL before geting struct member. We might hit this part maybe because ml_file is NULL
And dow we really need these if conditions ? TDNF_SAFE_FREE_MEMORY is doing it anyway right?

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+       IsNullOrEmptyString(digest)   ||
+       IsNullOrEmptyString(type))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    for(digest_alias = digest_aliases; digest_alias->alias_name;
+        ++digest_alias)
+    {
+        if(!strcmp(digest_alias->alias_name, type))
```
sshedi: ignore case please while comparing

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    metalink_digest_alias *digest_alias;
+
+    if(IsNullOrEmptyString(filename) ||
+       IsNullOrEmptyString(digest)   ||
+       IsNullOrEmptyString(type))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    for(digest_alias = digest_aliases; digest_alias->alias_name;
```
sshedi: Check for digest_alias is NULL

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+        ++digest_alias)
+    {
+        if(!strcmp(digest_alias->alias_name, type))
+        {
+            digest_alias->func(filename, &result);
+            if(!memcmp(result, digest,
+                       digest_alias->length))
+            {
+                /* sha*sum style verdict output */
+                fprintf(stdout, "Validating metalink (%s) OK\n", filename);
```
sshedi: printf

client/remoterepo.c
```diff
@@ -17,8 +17,15 @@
  *
  * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
  */
-
 #include "includes.h"
+static metalink_digest_alias digest_aliases[] = {
```
sshedi: new line

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    for(digest_alias = digest_aliases; digest_alias->alias_name;
+        ++digest_alias)
+    {
+        if(!strcmp(digest_alias->alias_name, type))
+        {
+            digest_alias->func(filename, &result);
```
sshedi: Check retval before comparing

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+
+uint32_t
+TDNFCheckHash(
+    const char *filename,
+    const char *digest,
+    const char *type
+    )
+{
+    uint32_t dwError = 0;
+    unsigned char *result;
```
sshedi: init to NULL

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    int i = 0;
+    if(!hex_digest ||
+       !digest_length)
+    {
+        return 0;
+    }
+    for(i = 0; hex_digest[i]; ++i)
+    {
+        char c = hex_digest[i];
+        if(!(('0' <= c && c <= '9') || ('a' <= c && c <= 'z') ||
```
sshedi: Use isalnum(hex_digest[i]) from ctype.h

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+error:
+    goto cleanup;
+}
+
+/* Returns nonzero if hex_digest is properly formatted; that is each
+   letter is in [0-9A-Za-z] and the length of the string equals to the
+   result length of digest * 2. */
+uint32_t
+TDNF_Check_Hex_Digest
+    (const char *hex_digest,
```
sshedi: Style mismatch

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+        }
+    }
+    return digest_length * 2 == i;
+}
+
+char 
+TDNF_HexToUint
+    (const char *s)
+{
+    char buf[3];
```
sshedi: init to 0, check for null of function param and style mismatch in function braces

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    unsigned long val;
+
+    buf[0] = s[0];
+    buf[1] = s[1];
+    buf[2] = 0;
```
sshedi: if you init buf to 0, no need of this

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    val = strtoul(buf, NULL, 16);
```
sshedi, line 421: validate retval of strtoul

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    (const char *hex_digest,
+    char **ppdigest
+    )
+{
+    uint32_t dwError = 0;
+    char *digest;
+    size_t i;
+    size_t len = 0;
+
+    if(!hex_digest ||
```
sshedi: IsNullOrEmpty of hex_digest

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+}
+
+uint32_t
+TDNF_New_Metalink_Checksum_From_Hex_Digest
+    (const char *hex_digest,
+    char **ppdigest
+    )
+{
+    uint32_t dwError = 0;
+    char *digest;
```
sshedi: init variables

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    }
+
+    len = strlen(hex_digest);
+
+    dwError = TDNFAllocateMemory(1, len/2, (void **)&digest);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    for(i = 0; i < len; i += 2)
+    {
+        digest[i/2] = TDNF_HexToUint(hex_digest + i);
```
sshedi: TDNF_New_Metalink_Checksum_From_Hex_Digest function needs  a bit of rework, if TDNF_HexToUint fails, we have no way of knowing it

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    goto cleanup;
+}
+
+uint32_t
+TDNF_New_Metalinkfile
+    (metalink_file_t *fileinfo,
+    metalinkfile ****ml_file
+    )
+{
+    metalinkfile *metalink_file = NULL;
```
sshedi, line 524: validate function args

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    uint32_t dwError = 0;
+    dwError = TDNFAllocateMemory(1, sizeof(metalinkfile), (void **)&metalink_file);
```
sshedi, line 539: new line

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFAllocateString(fileinfo->name, &(metalink_file->filename));
+    BAIL_ON_TDNF_ERROR(dwError); 
+    
+    metalink_file->digest = NULL;
+    if(fileinfo->checksums)
+    {
+        metalink_digest_alias *digest_alias;
+        for(digest_alias = digest_aliases; digest_alias->alias_name; ++digest_alias)
```
sshedi: check for digest_alias NULL

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+        {
+            metalink_checksum_t **metalink_checksum;
+            for(metalink_checksum = fileinfo->checksums; *metalink_checksum; ++metalink_checksum)
```
sshedi: check metalink_checksum for null

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    }
+    if(fileinfo->resources)
+    {
+        metalink_resource_t **metalink_resource = fileinfo->resources;
+        /* Filter by type if it is non-NULL. In Metalink v3, type
+           includes the type of the resource. In repo, we are only
+           interested in HTTP, HTTPS ,FTP and rsync.
+        */
+        if((*metalink_resource)->type == NULL ||
+           !strcmp((*metalink_resource)->type, "http") ||
```
sshedi: Keep all types (http, https ...) in an array and compare in a loop and ignore case

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+        {
+            dwError = ERROR_TDNF_INVALID_REPO_FILE;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+    }
+
+cleanup:
+    return dwError;
+error:
+    if(metalink_file->filename)
```
sshedi: what if metalink_file itself is NULL?
and TDNF_SAFE_FREE_MEMORY checks for null anyway

client/remoterepo.c
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    }
+    goto cleanup;  
+}
 uint32_t
 TDNFParseAndGetURLFromMetalink(
     PTDNF pTdnf,
     const char *pszRepo,
-    const char *pszFile
+    const char *pszFile,
+    metalinkfile ***ml_file
```
sshedi: validate ml_file for null

client/remoterepo.c
```diff
@@ -176,13 +662,27 @@ TDNFParseAndGetURLFromMetalink(
...
+    {
+        fprintf(stderr, "Metalink does not contain any valid file.\n");
+        dwError = ERROR_TDNF_INVALID_REPO_FILE;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
     for(files = metalink->files; files && *files; ++files)
     {
         resources = (*files)->resources;
-        if(!resources)
+        if(!resources || (*resources == NULL))
```
sshedi: Use IsNullOrEmpty macro instead

client/remoterepo.c
```diff
@@ -306,7 +807,7 @@ TDNFDownloadFile(
                 fclose(fp);
                 fp = NULL;
             }
-            dwError = TDNFParseAndGetURLFromMetalink(pTdnf, pszRepo, pszFile);
+            dwError = TDNFParseAndGetURLFromMetalink(pTdnf, pszRepo, pszFile, &ml_file);
```
sshedi: Why do we need to keep passing address of ml_file and use double and triple pointers? Why can we use single pointer throughout

pytests/repo/setup-repo.sh
```diff
@@ -80,13 +80,23 @@ repodir=${TEST_REPO_DIR}/yum.repos.d
 cachedir=${TEST_REPO_DIR}/cache/tdnf
 EOF
 
+SHA1SUM=$(sha1sum ${PUBLISH_PATH}/repodata/repomd.xml | cut -d' ' -f1)
+SHA256SUM=$(sha256sum ${PUBLISH_PATH}/repodata/repomd.xml | cut -d' ' -f1)
+MD5SUM=$(md5sum ${PUBLISH_PATH}/repodata/repomd.xml | cut -d' ' -f1)
```
sshedi: Add some more negative test cases for failures, like invalid digest name, invalid protocol, invalid md5sum and so on

sshedi (changes requested):
Please address these changes, I'll review again

client/includes.h
tapakund: Done, Thanks!

client/includes.h
tapakund: Done, Thanks!

client/prototypes.h
tapakund: Done, Thanks!

client/remoterepo.c
tapakund: Done, Thanks!

client/remoterepo.c
tapakund: There are different function calls here for each digest, which can be managed using switch. But depending on type of hash, we have different digest length as well. Let me try passing length as well using function params. I thought about this.. but seeing so many variables.. thought to keep separate functions.

client/remoterepo.c
tapakund: I see fprintf used before as well, so followed the same here.

client/remoterepo.c
tapakund: yeah, i think we don't need this condition. But we might need to check if ml_file is not NULL? What do you think?

client/remoterepo.c
tapakund: Done.

client/remoterepo.c
tapakund: Added changes. Thanks!

client/remoterepo.c
tapakund: Done, thanks.

client/remoterepo.c
tapakund: sorry, did not get your point?

client/remoterepo.c
tapakund: Will check.

client/remoterepo.c
tapakund: Sure.

client/remoterepo.c
tapakund: agree

client/remoterepo.c
tapakund: Yes, right. Done.

client/remoterepo.c
tapakund: Changed it to right check.

client/remoterepo.c
tapakund: In case of just parsing metalink, we will not download anything.. Just parse the existing metalink to set the base url. In that case ml_file is expected to be null.

In this function, before calling  TDNF_New_Metalinkfile we are checking if ml_file is null or not. As the only usage for ml_file is in  TDNF_New_Metalinkfile.

client/remoterepo.c
tapakund: I thought this is better to track, let me know your thoughts?

pytests/repo/setup-repo.sh
tapakund: Yeah, will add some more test cases including negative test case for testing the failures.

tapakund (commented):
Did most of the changes, few of them pending along with addition of negative test cases. Will post it in sometime.

ci/Dockerfile.photon
tapakund: Shall we do it in separate PR?

client/remoterepo.c
tapakund: done.

client/remoterepo.c
tapakund: Done.

client/remoterepo.c
tapakund: Done.

client/remoterepo.c
tapakund: done.

client/remoterepo.c
tapakund: added, nice catch

client/remoterepo.c
tapakund: done.

ci/Dockerfile.photon
sshedi: Sure, np

client/remoterepo.c
sshedi: Sure, you can use the same call back functions in generic function as well

client/remoterepo.c
sshedi: Yes, fprintf(stderr, ...) is okay because stderr is not buffered, it'll flush everything to tty the moment it has data. fprintf(stdout, ...) = printf

client/remoterepo.c
sshedi: Definitely

client/remoterepo.c
sshedi: Okay, understood. But it'll be good if we have some kind of check to ensure that ml_file is non null when we are not parsing metalink file

client/remoterepo.c
sshedi: I say, it makes life difficult. We should avoid using double and triple pointers unless it's absolutely necessary.

client/remoterepo.c
sshedi: tdnf functions look like

```C
return_type
func_name(
    arg1,
    arg2,
    ...
    )

File: client/remoterepo.c (hunk 28 above)
Comment (tapakund) [reply client/remoterepo.c]: thats taken care.

File: client/remoterepo.c (hunk 6 above)
Comment (tapakund) [reply client/remoterepo.c]: Ok, agreed.

File: client/remoterepo.c (hunk 10 above)
Comment (tapakund) [reply client/remoterepo.c]: done.

File: client/remoterepo.c (hunk 15 above)
Comment (tapakund) [reply client/remoterepo.c]: oops.. corrected.

File: client/remoterepo.c (hunk 21 above)
Comment (tapakund) [reply client/remoterepo.c]: agree, i am working on it. We need to handle error.

File: client/remoterepo.c [hunk 32]
```diff
@@ -106,11 +113,490 @@ _handle_curl_cb(
...
+    {
+        TDNF_SAFE_FREE_MEMORY(digest);
+    }
+    goto cleanup;
+}
+
+uint32_t
+TDNF_New_Metalinkfile
+    (metalink_file_t *fileinfo,
+    metalinkfile ****ml_file
```
sidcha: This '****' construct is strong indicator for a design change/refactor. One possible, first glance fix would be to keep a `metalinkfile *` in `metalink_file_t` but you'd know better.

Generally going above a double pointers is accepted only upto triple pointers with very good reasoning :).

sidcha (changes requested):
Needs a minor rework.

client/remoterepo.c
tapakund: I am doing the changes to keep it upto double pointer.

client/remoterepo.c
tapakund: Done.

client/remoterepo.c
tapakund: done

client/remoterepo.c
tapakund: taken care.

client/remoterepo.c
tapakund: I feel we anyways need to keep different functions for getting the digest as everyone have different function calls. So, lets not add a new wrapper function for calling them as we will need to pass lot of things like, fd, length, etc..  

Also, we are using pointers to map to respective digest type so its fine.

pytests/repo/setup-repo.sh
tapakund: Added.

tapakund:
Test is failing for:

Error: 404 when downloading https://packages.vmware.com/photon/3.0/photon_updates_3.0_x86_64/x86_64/file-libs-5.38-1.ph3.x86_64.rpm

I see we have:

https://packages.vmware.com/photon/3.0/photon_updates_3.0_x86_64/x86_64/file-libs-5.34-2.ph3.x86_64.rpm

We do not have file-libs-5.38 in the packages.vmware.com. Fedora check passed.

client/remoterepo.c
sidcha: Actually, @sshedi's point - duplicate code - is still valid. If you don't like his API suggestion, you could still have it as an unexported method and then create wrappers around that method like:

```
TDNFGetSHA1Sum(file, **res) {
    return __TDNFGetHASH(file, res, TDNF_HASH_SHA1)
}
TDNFGetMD5Sum(file, **res) {
    return __TDNFGetHASH(file, res, TDNF_HASH_MD5)
}
```
This way, `__TDNFGetHASH` will be the only method we need to maintain while still preserving your external contract.

Thoughts?

client/repo.c
```diff
@@ -662,9 +664,18 @@ TDNFGetRepoMD(
...
                           pszTmpRepoMDFile,
                           pRepoData->pszId,
-                          0);
+                          0,
+                          NULL);
         BAIL_ON_TDNF_ERROR(dwError);
 
+        if(metalink)
+        {
+            if(ml_file)
```
sshedi: if (metalink && ml_file) looks better here instead of nested ifs

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+    int i = 0;
+    if(IsNullOrEmptyString(hex_digest) ||
+       !digest_length)
+    {
+        return 0;
+    }
+    for(i = 0; hex_digest[i]; ++i)
+    {
+        char c = hex_digest[i];
+        if(!isalnum(c))
```
sidcha: This check is incorrect. You must check for `[0-9A-Fa-f]`. So, `s/isalnum/isxdigit/` and update the comment above accordingly.

client/remoterepo.c
sshedi: Okay, np. Maybe we can revisit this in future.

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+   result length of digest * 2. */
+uint32_t
+TDNF_Check_Hex_Digest(
+    const char *hex_digest,
+    int digest_length
+    )
+{
+    int i = 0;
+    if(IsNullOrEmptyString(hex_digest) ||
+       !digest_length)
```
sshedi: Check for digest_length <= 0

client/remoterepo.c
sshedi: if(!isalnum(hex_digest[i])) would suffice

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+    return digest_length * 2 == i;
+}
+
+uint32_t
+TDNF_HexToUint(
+    const char *s,
+    unsigned char *return_value
+    )
+{
+    uint32_t dwError = 0;
```
sshedi, line 406: Validate function params

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+}
+
+uint32_t
+TDNF_New_Metalink_Checksum_From_Hex_Digest(
+    const char *hex_digest,
+    char **ppdigest
+    )
+{
+    uint32_t dwError = 0;
+    char *pdigest;
```
sshedi: Init pointers to NULL and variables as needed

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+
+    dwError = TDNFAllocateMemory(1, len/2, (void **)&pdigest);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    for(i = 0; i < len; i += 2)
+    {
+        dwError = TDNF_HexToUint(hex_digest + i, &ret);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        pdigest[i/2] = ret;
```
sshedi: pdigest[i >> 1] ? :)

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+    metalinkfile *metalink_file = NULL;
+    uint32_t dwError = 0;
+
+    if(!fileinfo)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(1, sizeof(metalinkfile), (void **)&metalink_file);
```
sshedi, line 539: Do we need to cast to (void **) ?

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+        */
+        if((*metalink_resource)->url == NULL)
+        {
+            dwError = ERROR_TDNF_INVALID_REPO_FILE;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+        else
+        {
+            if((*metalink_resource)->type == NULL ||
+               !strcasecmp((*metalink_resource)->type, "http") ||
```
sshedi: keep the supported protocols in an array (possibly a structure). In case if we need to compare few more attributes in future, it'll scale well

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+        TDNF_SAFE_FREE_MEMORY(metalink_file);
+    }
+    goto cleanup;  
+}
 uint32_t
 TDNFParseAndGetURLFromMetalink(
     PTDNF pTdnf,
     const char *pszRepo,
-    const char *pszFile
+    const char *pszFile,
```
sshedi, line 629: Validate function params

client/remoterepo.c
```diff
@@ -207,7 +725,8 @@ TDNFDownloadFile(
     const char *pszFileUrl,
     const char *pszFile,
     const char *pszProgressData,
-    int is_metalink
+    int is_metalink,
+    metalinkfile **ml_file
```
sshedi, line 743: validate function args

sshedi (changes requested):

client/remoterepo.c
```diff
@@ -106,11 +115,506 @@ _handle_curl_cb(
...
+    *return_value = (unsigned char)(val&0xff);
+
+cleanup:
+    return dwError;
+error:
+    goto cleanup;
+}
+
+uint32_t
+TDNF_New_Metalink_Checksum_From_Hex_Digest(
```
sidcha: This method (and its children) looks needlessly complicated. You need to have a utility method that converts hex string to byte array. Then just check that byte array everywhere. Here is a sample:

```c
static inline int hex2int(char ch)
{
	if (ch >= '0' && ch <= '9')
		return ch - '0';
	if (ch >= 'A' && ch <= 'F')
		return ch - 'A' + 10;
	if (ch >= 'a' && ch <= 'f')
		return ch - 'a' + 10;
	return -1;
}

/* hex_string_to_bytes:
 * hstr: "AABB12" becomes arr: { 0xAA, 0xBB, 0x12 }
 * Note: arr must be at least strlen(hstr)/2  */
int hex_string_to_bytes(const char *hstr, uint8_t *arr)
{
	int i, len, nib1, nib2;

	len = strlen(hstr);
	if ((len & 1) == 1 || len == 0) {
		return -1; // Must have even number of chars
	}
	len = len / 2;
	for (i = 0; i < len; i++) {
		nib1 = hex2int(hstr[(i * 2) + 0]);
		nib2 = hex2int(hstr[(i * 2) + 1]);
		if (nib1 == -1 || nib2 == -1) {
			return -1;
		}
		arr[i] = nib1 << 4 | nib2;
	}
	return len;
}
```
As this method does check and convert, you can get rid of `TDNF_Check_Hex_Digest` altogether and work only with the converted byte array.

sidcha (commented):
Some code flow changes.

client/remoterepo.c
tapakund: Yes.

client/remoterepo.c
tapakund: created a common function to manage it. Thanks @cbsiddharth @sshedi

client/remoterepo.c
tapakund: If we get NULL, we pass it to TDNFParseAndGetURLFromMetalink and this will take care.

client/remoterepo.c
tapakund: if  ml_file is not null, then we will call.

if(ml_file)
        {
            dwError = TDNF_New_Metalinkfile((*files), ml_file);
            BAIL_ON_TDNF_ERROR(dwError);
        }

It can be NULL depending on use case, so havent added check at the beginning of function.

client/remoterepo.c
tapakund: Thanks used isxdigit instead.

tapakund: agree.

client/remoterepo.c
tapakund: done.

client/remoterepo.c
tapakund: done.

client/remoterepo.c
tapakund: :)

client/remoterepo.c
tapakund: added

client/repo.c
tapakund: yes.

client/prototypes.h
```diff
@@ -21,6 +21,16 @@
 #ifndef __CLIENT_PROTOTYPES_H__
 #define __CLIENT_PROTOTYPES_H__
 
+#include <openssl/sha.h>
+#include <openssl/md5.h>
+
+typedef struct metalinkfile{
+    char *filename;
+    int   type;
```
sidcha, line 30: change the type from int to enum TDNF_HashType_e.

client/includes.h
```diff
@@ -69,5 +70,12 @@
 
 #include <metalink/metalink_parser.h>
 
+// Enum in order of preference
+enum {
```
sidcha, line 74: Name this enum TDNF_HashType_e

client/remoterepo.c
```diff
@@ -17,9 +17,109 @@
...
 #include "includes.h"
 
+typedef union {
+    MD5_CTX md5_ctx;
+    SHA_CTX sha1_ctx;
+    SHA256_CTX sha256_ctx;
+} hash_ctx_t;
+
+int
+TDNF_SHA1_Init(
```
sidcha, line 30: A single line comment above each method, indicating that these methods are just wrappers would help readers.

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+
+uint32_t
+TDNFCheckHash(
+    const char *filename,
+    unsigned char *digest,
+    int type
+    )
+{
+    uint32_t dwError = 0;
+    uint8_t result[64];
```
sshedi: Init buffer to 0

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (type  < TDNF_HASH_MD5 || type >= TDNF_HASH_SENTINEL)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    hash_op *hash = hash_ops + type;
```
sshedi: declare at top

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+        dwError = ERROR_TDNF_INVALID_REPO_FILE;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+
+    printf("Validating metalink (%s) OK\n", filename);
+
+cleanup:
+    return dwError;
+error:
+    fprintf(stderr, "Error: Validating metalink (%s) FAILED (digest mismatch)\n", filename);
```
sshedi: Check filename for NULL before printing

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+
+
+uint32_t
+TDNFMetalinkCheckHash(
+    char *pszFile,
+    metalinkfile *ml_file
+    )
+{
+    uint32_t dwError = 0;
+    if(IsNullOrEmptyString(pszFile) ||
```
sshedi, line 346: new line

client/remoterepo.c
sshedi: Reminder for the above comment

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+    uint32_t dwError = 0;
+
+    if(IsNullOrEmptyString(resource_type) ||
+       !type)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if(!strcasecmp(resource_type, "sha256") || !strcasecmp(resource_type, "sha-256"))
```
sidcha: Break these down to next line. Try to maintain 80 char width where possible.

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+    uint8_t *result
+    )
+{
+    uint32_t dwError = 0;
+    int fd = -1;
+    hash_ctx_t ctx;
+    char buf[BUFSIZ] = {0};
+    int length = 0;
+
+    if(IsNullOrEmptyString(filename) ||
```
sidcha: Put these in one line. Looks a little odd.

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+    metalinkfile **ml_file
+    )
+{
+    metalinkfile *metalink_file = NULL;
+    uint32_t dwError = 0;
+    int i = 0;
+    int offset = -1;
+    int type = 0;
+    char *resource_type [] = { "https", "http", "ftp", "ftps", "rsync", NULL };
+    if(!fileinfo)
```
sshedi, line 533: new line

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+    metalink_file->type = -1;
+    //set digest to NULL
+    metalink_file->digest = NULL;
+    
+    if(!fileinfo->checksums)
+    {
+        dwError = ERROR_TDNF_INVALID_REPO_FILE;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    metalink_checksum_t **metalink_checksum = fileinfo->checksums;
```
sshedi: declare at top

client/prototypes.h
```diff
@@ -177,7 +194,26 @@ TDNFDownloadFile(
...
+uint32_t
+TDNFGetSHA256Sum(
+    const char *filename,
+    char **type
+    );
+
+uint32_t
+TDNFGetMD5Sum(
+    const char *filename,
+    char **type
```
sidcha: Are these exports still needed?

sidcha (commented):
Minor comments. Otherwise LGTM.

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+    {
+        dwError = ERROR_TDNF_INVALID_REPO_FILE;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    metalink_resource_t **metalink_resource = fileinfo->resources;
+    /* Filter by type if it is non-NULL. In Metalink v3, type
+       includes the type of the resource. In repo, we are only
+       interested in HTTP, HTTPS ,FTP and rsync.
+    */
+    if((*metalink_resource)->url == NULL)
```
sshedi, line 593: We need to check metalink_resource for NULL, just in case?

client/remoterepo.c
```diff
@@ -106,11 +206,409 @@ _handle_curl_cb(
...
+    {
+        dwError = ERROR_TDNF_INVALID_REPO_FILE;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    if(!fileinfo->resources)
+    {
+        dwError = ERROR_TDNF_INVALID_REPO_FILE;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    metalink_resource_t **metalink_resource = fileinfo->resources;
```
sshedi: Declare at top

client/remoterepo.c
sshedi: Reminder for above comment

client/remoterepo.c
sshedi: Reminder

client/remoterepo.c
sshedi: Can we have a 1-1 mapping for these hash methods and types by using a struct? You've declared an enum already,
So we can do it inside a loop instead of an if else ladder.
I suggest putting everything in _hash_op structure and using it everywhere.

sshedi: It'll be nice if we have a cutomised version of strcasecmp :) I think it's not too much to ask. It's your wish to do it or not :)

```C
/*
 * s1 - as in stracasecmp
 * s2 - as in strcasecmp
 * ignore_str - ignore the characters in this string while comparing
*/
 
int custom_strcasecmp(const char *s1, const char *s2, const char *ignore_str);

in our case we can call, custom_strcasecmp(resource_type, "sha-256", "-");

File: client/remoterepo.c (hunk 49 above)
Comment (tapakund) [reply client/remoterepo.c]: not required. We have null check at the beginning of fucntion.

File: client/includes.h (hunk 45 above)
Comment (tapakund) [reply client/includes.h:74]: I am keeping it like that for now.

File: client/prototypes.h (hunk 44 above)
Comment (tapakund) [reply client/prototypes.h:30]: Not addressing this right now.

File: client/prototypes.h (hunk 55 above)
Comment (tapakund) [reply client/prototypes.h]: thanks for pointing out. removed.

File: client/remoterepo.c (hunk 46 above)
Comment (tapakund) [reply client/remoterepo.c:30]: added comments, those are tdnf wrappers.

File: client/remoterepo.c (hunk 52 above)
Comment (tapakund) [reply client/remoterepo.c]: done.

File: client/remoterepo.c (hunk 47 above)
Comment (tapakund) [reply client/remoterepo.c]: done.

File: client/remoterepo.c (hunk 48 above)
Comment (tapakund) [reply client/remoterepo.c]: done.

File: client/remoterepo.c (hunk 50 above)
Comment (tapakund) [reply client/remoterepo.c:346]: done.

File: client/remoterepo.c (hunk 51 above)
Comment (tapakund) [reply client/remoterepo.c]: sure.

Comment (tapakund) [reply client/remoterepo.c]: if we add that, it will be again a double loop.. trying to avoid it. Also beside "sha1" and "sha1-1" there is no other standard representation.

File: client/remoterepo.c (hunk 53 above)
Comment (tapakund) [reply client/remoterepo.c:533]: done.

File: client/remoterepo.c (hunk 54 above)
Comment (tapakund) [reply client/remoterepo.c]: done.

File: client/remoterepo.c (hunk 57 above)
Comment (tapakund) [reply client/remoterepo.c]: done.

File: client/remoterepo.c (hunk 56 above)
Comment (tapakund) [reply client/remoterepo.c:593]: added check.

File: client/remoterepo.c (hunk 49 above)
Comment (sshedi) [reply client/remoterepo.c]: Yes but what if we reach here because filename is  NULL?

File: client/remoterepo.c [hunk 58]
```diff
@@ -106,11 +215,412 @@ _handle_curl_cb(
...
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    for(i = 0; i < len; i += 2)
+    {
+        dwError = TDNF_HexToUint(hex_digest + i, &ret);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        pdigest[i>>1] = ret;
+    }
+    memcpy(ppdigest, pdigest, len/2);
```
sshedi: len >> 1 :)

client/remoterepo.c
```diff
@@ -106,11 +215,412 @@ _handle_curl_cb(
...
+uint32_t
+TDNF_Check_Digest(
+    const char *filename,
+    hash_op *hash,
+    uint8_t *result
+    )
+{
+    uint32_t dwError = 0;
+    int fd = -1;
+    hash_ctx_t ctx;
```
sshedi, line 229: I think we should memset ctx to 0 before passing it's address. Just a good practice

sshedi (commented):
Almost there, small changes.

client/remoterepo.c
tapakund: this is a pointer which will be allocated by sha/md5 init function. So not required to memset this.

sidcha (approved):
LGTM.

sshedi (approved):
Looks good.

client/remoterepo.c
ppadmavilasom: please init all variables. @sshedi is right.

ppadmavilasom (commented):
@keerthanakalyan did we not add shasum check elsewhere ? can you please point @tapakund to it so we can reuse existing?

tapakund:
> @keerthanakalyan did we not add shasum check elsewhere ? can you please point @tapakund to it so we can reuse existing?

@ppadmavilasom @keerthanakalyan i see we are using solv_chksum_create API for checksum for repomd file while we refresh. I also checked the implementation of this in solv code base. They are internally using openssl APIs similar we we used for calculating the shasum. I will use the same for this.

sidcha:
> I will use the same for this.

Its cleaner to directly call openssl methods (which are well documented) from TDNF if this going to be a common requirement. Now that we have a okay-ish hashing API, we can modify the other instance to use this API instead. If needed, I can send a clean up patch to enhance this hashing API.

@ppadmavilasom  Thoughts?

client/remoterepo.c
```diff
@@ -106,11 +215,415 @@ _handle_curl_cb(
...
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(1, sizeof(metalinkfile), (void **)&metalink_file);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFAllocateString(fileinfo->name, &(metalink_file->filename));
+    BAIL_ON_TDNF_ERROR(dwError);
+    //set type to 0
```
keerthanakalyan: Comment needs to be changed.

client/remoterepo.c
```diff
@@ -106,11 +215,415 @@ _handle_curl_cb(
...
+            return 0;
+        }
+    }
+    return digest_length * 2 == i;
+}
+
+uint32_t
+TDNF_HexToUint(
+    const char *hex_digest,
+    unsigned char *return_value
```
keerthanakalyan: It would be good to have a descriptive names for the function params instead of common name.

client/remoterepo.c
```diff
@@ -106,11 +215,415 @@ _handle_curl_cb(
...
+    }
+
+    len = strlen(hex_digest);
+
+    dwError = TDNFAllocateMemory(1, len/2, (void **)&pdigest);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    for(i = 0; i < len; i += 2)
+    {
+        dwError = TDNF_HexToUint(hex_digest + i, &ret);
```
keerthanakalyan: It would be good if the variables (ret) are named descriptively.

keerthanakalyan (commented):
General comment: 
1. Some of the functions are written with underscore and some are with MixedCase, since TDNF already follows MixedCase coding style it would be good if we use the same.
2. It would be good if we have descriptive variable names and function params for better readability and understanding.

client/remoterepo.c
tapakund: Thanks, Done.

client/remoterepo.c
tapakund: Done.

client/remoterepo.c
tapakund: Done.

client/remoterepo.c
tapakund: Added memset for the union variable :)

client/includes.h
tapakund: Done.

ppadmavilasom (approved):

PR #131, opened 2020-04-25
Bug fixes and code improvements

sshedi:
Fixed a potential buffer overflow with fread
Use getchar instead of scanf

### DIFF
--- a/client/defines.h
+++ b/client/defines.h
@@ -74,20 +74,6 @@ typedef enum
         }                                                          \
     } while(0)
 
-#define TDNF_SAFE_FREE_MEMORY(pMemory) \
-    do {                                                           \
-        if (pMemory) {                                             \
-            TDNFFreeMemory(pMemory);                               \
-        }                                                          \
-    } while(0)
-
-#define TDNF_SAFE_FREE_STRINGARRAY(ppArray) \
-    do {                                                           \
-        if (ppArray) {                                             \
-            TDNFFreeStringArray(ppArray);                          \
-        }                                                          \
-    } while(0)
-
 //Misc
 #define TDNF_RPM_EXT                      ".rpm"
 #define TDNF_NAME                         "tdnf"
--- a/common/defines.h
+++ b/common/defines.h
@@ -40,11 +40,11 @@
         }                                                          \
     } while(0)
 
-
 #define TDNF_SAFE_FREE_MEMORY(pMemory) \
     do {                                                           \
         if (pMemory) {                                             \
             TDNFFreeMemory(pMemory);                               \
+            pMemory = NULL;                                        \
         }                                                          \
     } while(0)
 
@@ -52,6 +52,7 @@
     do {                                                           \
         if (ppArray) {                                             \
             TDNFFreeStringArray(ppArray);                          \
+            ppArray = NULL;                                        \
         }                                                          \
     } while(0)
 
--- a/common/setopt.c
+++ b/common/setopt.c
@@ -68,8 +68,8 @@ uint32_t
 AddSetOptWithValues(
     PTDNF_CMD_ARGS pCmdArgs,
     int nType,
-    const char* pszOptArg,
-    const char* pszOptValue
+    const char *pszOptArg,
+    const char *pszOptValue
     )
 {
     uint32_t dwError = 0;
@@ -80,11 +80,10 @@ AddSetOptWithValues(
        IsNullOrEmptyString(pszOptArg) ||
        IsNullOrEmptyString(pszOptValue) || nType == CMDOPT_CURL_INIT_CB)
     {
-        dwError = ERROR_TDNF_INVALID_PARAMETER;
-        BAIL_ON_TDNF_ERROR(dwError);
+        BAIL_ON_TDNF_ERROR((dwError = ERROR_TDNF_INVALID_PARAMETER));
     }
 
-    dwError = TDNFAllocateMemory(1, sizeof(TDNF_CMD_OPT), (void**)&pCmdOpt);
+    dwError = TDNFAllocateMemory(1, sizeof(TDNF_CMD_OPT), (void **)&pCmdOpt);
     BAIL_ON_TDNF_ERROR(dwError);
 
     pCmdOpt->nType = nType;
@@ -96,24 +95,24 @@ AddSetOptWithValues(
     BAIL_ON_TDNF_ERROR(dwError);
 
     pSetOptTemp = pCmdArgs->pSetOpt;
-    if(!pSetOptTemp)
+    if (pSetOptTemp)
     {
-        pCmdArgs->pSetOpt = pCmdOpt;
-    }
-    else
-    {
-        while(pSetOptTemp->pNext)
+        while (pSetOptTemp->pNext)
         {
             pSetOptTemp = pSetOptTemp->pNext;
         }
         pSetOptTemp->pNext = pCmdOpt;
+    }
+    else
+    {
+        pCmdArgs->pSetOpt = pCmdOpt;
     }
 
 cleanup:
     return dwError;
 
 error:
-    if(pCmdOpt)
+    if (pCmdOpt)
     {
         TDNFFreeCmdOpt(pCmdOpt);
     }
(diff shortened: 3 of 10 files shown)


tools/cli/lib/installcmd.c
```diff
@@ -218,16 +217,20 @@ TDNFCliAlterCommand(
 
     if(pSolvedPkgInfo->nNeedAction)
     {
+        int32_t opt = 0;
```
tapakund, line 220: shall we use uint8?

tools/cli/lib/installcmd.c
```diff
@@ -218,16 +217,20 @@ TDNFCliAlterCommand(
...
+
         if(!pCmdArgs->nAssumeYes && !pCmdArgs->nAssumeNo)
         {
-            printf("Is this ok [y/N]:");
-            if (scanf("%c", &chChoice) != 1)
+            printf("Is this ok [y/N]: ");
+
+            opt = getchar();
+            if (tolower(opt) != 'y' && tolower(opt) != 'n')
```
tapakund, line 227: shall we check for opt != EOF? As i am not sure the behaviour of tolower if we pass EOF as an input?

tapakund (commented):
Minor comments, otherwise LGTM.

tools/cli/lib/installcmd.c
sshedi: From - man getchar

int getchar(void);

Return Value
fgetc(), getc() and getchar() return the character read as an unsigned char cast to an int or EOF on end of file or error.

tools/cli/lib/installcmd.c
sshedi: EOF is CTRL-D so it's still an invalid input so we must fail. So no need to check again

tapakund (approved):
LGTM

tools/cli/lib/parseargs.c
```diff
@@ -335,64 +322,31 @@ ParseOption(
...
-        BAIL_ON_CLI_ERROR(dwError);
     }
-    else if(!strcasecmp(pszName, "skipconflicts"))
+    else if(!strcasecmp(pszName, "skipconflicts") ||
+            !strcasecmp(pszName, "skipobsoletes") ||
+            !strcasecmp(pszName, "skipsignature") ||
+            !strcasecmp(pszName, "skipdigest") ||
+            !strcasecmp(pszName, "noplugins") ||
+            !strcasecmp(pszName, "security") ||
+            !strcasecmp(pszName, "reboot-required"))
```
ppadmavilasom: would it be better if we make a local struct with args that needs to be passed to AddSetOptWithValues 
like 
struct setopt_args
{
TDNF_CMDOPT_TYPE type;
char *optarg;
char *optval;
} optvaltable[] =
{
{CMDOPT_KEYVALUE, "sec-severity",  optarg},
{CMDOPT_KEYVALUE, "security",  "1"},
};

etc and rewrite this function to loop through the table and just handle the extra validation of comma separated strings (in exclude for eg) as needed?

sshedi: Good idea but how about: https://github.com/vmware/tdnf/blob/ae364836529ec020d66ac1af86590c70f2711332/common/setopt.c#L352?

sshedi: It can be done, ignore my previous comment. Was thinking about some different solution of passing a struct as function parameter

tools/cli/lib/parseargs.c
```diff
@@ -251,193 +268,108 @@ TDNFCopyOptions(
...
+        uint32_t i = 0;
+
+        for (i = 0; i < ARRAY_SIZE(OptValTable); i++)
+        {
+            dwError = TDNFAllocateString(OptValTable[i].OptName, &pszCopyArgs);
+            BAIL_ON_CLI_ERROR(dwError);
+
+            ToFree = pszCopyArgs;
+
+            while ((pszToken = strsep(&pszCopyArgs, " ")))
```
sshedi: strtok is obsoleted by strsep

tools/cli/lib/parseargs.c
```diff
@@ -251,193 +268,108 @@ TDNFCopyOptions(
...
-                                      optarg);
-        BAIL_ON_CLI_ERROR(dwError);
+        uint32_t i = 0;
+
+        for (i = 0; i < ARRAY_SIZE(OptValTable); i++)
+        {
+            dwError = TDNFAllocateString(OptValTable[i].OptName, &pszCopyArgs);
+            BAIL_ON_CLI_ERROR(dwError);
+
+            ToFree = pszCopyArgs;
```
sshedi: We need to retain the starting address of pszCopyArgs because strsep will modify it's address

client/defines.h
```diff
@@ -74,20 +74,6 @@ typedef enum
         }                                                          \
     } while(0)
 
-#define TDNF_SAFE_FREE_MEMORY(pMemory) \
```
sshedi, line 77: Duplicates

common/defines.h
```diff
@@ -40,18 +40,19 @@
         }                                                          \
     } while(0)
 
-
 #define TDNF_SAFE_FREE_MEMORY(pMemory) \
     do {                                                           \
         if (pMemory) {                                             \
             TDNFFreeMemory(pMemory);                               \
+            pMemory = NULL;                                        \
```
sshedi, line 47: Let's be safe than sorry. This helps if we are using this macro inside a loop

common/setopt.c
```diff
@@ -96,24 +95,24 @@ AddSetOptWithValues(
     BAIL_ON_TDNF_ERROR(dwError);
 
     pSetOptTemp = pCmdArgs->pSetOpt;
-    if(!pSetOptTemp)
-    {
-        pCmdArgs->pSetOpt = pCmdOpt;
-    }
-    else
+    if (pSetOptTemp)
```
sshedi, line 98: Let's keep the case with more hit rate first

tools/cli/lib/parseargs.c
```diff
@@ -21,6 +21,22 @@
...
+    const char *OptVal;
+} SetOptArgs;
+
+static SetOptArgs OptValTable[] = {
+    {CMDOPT_KEYVALUE, "sec-severity", NULL},
+    {CMDOPT_ENABLEREPO, "enablerepo", NULL},
+    {CMDOPT_DISABLEREPO, "disablerepo", NULL},
+    {CMDOPT_ENABLEPLUGIN, "enableplugin", NULL},
+    {CMDOPT_DISABLEPLUGIN, "disableplugin", NULL},
+    {CMDOPT_KEYVALUE, "skipconflicts skipobsoletes skipsignature skipdigest noplugins reboot-required security", "1"}
```
ppadmavilasom: if you add each with a delimiter (space okay) including the last one, you can do strcasecmp after strcat the delimiter to the search input. avoid loop

sshedi: Didn't get your point, please explain :)

ppadmavilasom: ah i see i meant strstr and wrote strcasecmp. 
this is what i meant for eg.
suppose CMDOPT_KEYVALUE, "skipconflicts;skipobsoletes;"
then, instead of strsep and loop to match, you could
say you have to_match="skipobsoletes", make delimited_tomatch="skipobsoletes;"
then you can do if (strstr("skipconflicts;skipobsoletes;", "skipobsoletes;")) add set opt values
no need for strsep and loop.

sshedi: I thought of it but it's not robust enough. If my input is:
tdnf check --blablabla-skipobsoletes, strstr will accept it :) that's why I went for the hard way of using strsep :)

sshedi: Or I got confused with haystack needle of strstr, time to check man page :)

ppadmavilasom: ParseOption is not called with raw input. if you check the calls, it is always matched from the options array or pre-determined constant strings like config. getopt might match aggressively but ParseOption is called with controlled input

PR #132, opened 2020-05-03
Use mmap instead of reading chunks

sshedi:
Use the TDNFFileReadAllText function instead of doing whole open-read-close in few places.
Removed some duplicate macros
Minor code improvements in few places

### DIFF
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -216,7 +216,6 @@ error:
     return dwError;
 }
 
-
 uint32_t
 TDNFCheckDigest(
     const char *filename,
@@ -225,64 +224,50 @@ TDNFCheckDigest(
     )
 {
     uint32_t dwError = 0;
-    int fd = -1;
-    hash_ctx_t ctx;
-    char buf[BUFSIZ] = {0};
-    int length = 0;
+    hash_ctx_t ctx = {0};
+    FileMapInfo fMap = {0};
 
-    if(IsNullOrEmptyString(filename) || !hash || !digest)
+    if (IsNullOrEmptyString(filename) || !hash || !digest)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    memset(&ctx, 0, sizeof(hash_ctx_t));
-
-    fd = open(filename, O_RDONLY);
-    if(fd == -1)
+    dwError = TDNFMapFile(filename, &fMap);
+    if (dwError)
     {
         fprintf(stderr, "Metalink: validating (%s) FAILED\n", filename);
         dwError = errno;
         BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
     }
-    dwError = hash->init_fn(&ctx);
 
-    if(!dwError)
+    dwError = hash->init_fn(&ctx);
+    if (!dwError)
     {
         fprintf(stderr, "Hash Init Failed\n");
         dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
         BAIL_ON_TDNF_ERROR(dwError);
     }
-    while((length = read(fd, buf, (sizeof(buf)-1))) > 0)
+
+    dwError = hash->update_fn(&ctx, fMap.fData, fMap.fSize);
+    if (!dwError)
     {
-        dwError = hash->update_fn(&ctx, buf, length);
-        if(!dwError)
-        {
-            fprintf(stderr, "Hash Update Failed\n");
-            dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
-            BAIL_ON_TDNF_ERROR(dwError);
-        }
-        memset(buf, 0, BUFSIZ);
+        fprintf(stderr, "Hash Update Failed\n");
+        BAIL_ON_TDNF_ERROR((dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED));
     }
-    if(length == -1)
-    {
-        fprintf(stderr, "Metalink: validating (%s) FAILED\n", filename);
-        dwError = errno;
-        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
-    }
+
     dwError = hash->final_fn(digest, &ctx);
-    if(!dwError)
+    if (!dwError)
     {
         dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
         BAIL_ON_TDNF_ERROR(dwError);
     }
     dwError = 0;
+
 cleanup:
-    if(fd != -1)
-    {
-        close(fd);
-    }
+    TDNFUnMapFile(&fMap);
     return dwError;
+
 error:
     goto cleanup;
 }
@@ -294,29 +279,27 @@ TDNFCheckHash(
     int type
     )
 {
-
+    hash_op *hash = NULL;
     uint32_t dwError = 0;
     uint8_t digest_from_file[MAX_DIGEST_LENGTH] = {0};
-    hash_op *hash = NULL;
 
-    if(IsNullOrEmptyString(filename) ||
-       !digest)
+    if (IsNullOrEmptyString(filename) || !digest)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    if (type  < TDNF_HASH_MD5 || type >= TDNF_HASH_SENTINEL)
+    if (type < TDNF_HASH_MD5 || type >= TDNF_HASH_SENTINEL)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
     hash = hash_ops + type;
-
     dwError = TDNFCheckDigest(filename, hash, digest_from_file);
     BAIL_ON_TDNF_ERROR(dwError);
-    if(memcmp(digest_from_file, digest, hash->length))
+
+    if (memcmp(digest_from_file, digest, hash->length))
     {
         dwError = ERROR_TDNF_INVALID_REPO_FILE;
         BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
@@ -326,14 +309,15 @@ TDNFCheckHash(
 
 cleanup:
     return dwError;
+
 error:
-    if(!IsNullOrEmptyString(filename))
+    if (!IsNullOrEmptyString(filename))
     {
         fprintf(stderr, "Error: Validating metalink (%s) FAILED (digest mismatch)\n", filename);
     }
+
     goto cleanup;
 }
-
 
 uint32_t
 TDNFMetalinkCheckHash(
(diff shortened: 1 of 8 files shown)


pytests/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ configure_file(
 )
 
 add_custom_target(check
-    COMMAND pytest
+    COMMAND pytest -x
```
sshedi: Let's stop after first failure, saves a lot of time while dev testing

common/utils.c
```diff
@@ -14,53 +14,50 @@ TDNFFileReadAllText(
...
-    {
-        dwError = ENOENT;
-        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
-    }
-    fseek(fp, 0, SEEK_END);
-    nLength = ftell(fp);
+    dwError = stat(pszFileName, &st);
+    BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+
+    nLength = st.st_size;
```
sshedi, line 101: Using stat here is cleaner than doing fseeks

client/remoterepo.c
```diff
@@ -650,56 +636,44 @@ TDNFParseAndGetURLFromMetalink(
...
     metalink_error = metalink_parse_final(metalink_context, NULL, 0, &metalink);
-    if((metalink_error != 0) || (metalink == NULL))
+    if (metalink_error || !metalink)
     {
         fprintf(stderr, "metalink_parse_final failed, ERROR: code=%d\n", metalink_error);
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
-    if(metalink->files == NULL)
+    metalink_context = NULL;
```
sshedi, line 667: If metalink_context not set to NULL, cleanup will fail

client/remoterepo.c
```diff
@@ -225,64 +224,50 @@ TDNFCheckDigest(
...
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    memset(&ctx, 0, sizeof(hash_ctx_t));
-
-    fd = open(filename, O_RDONLY);
-    if(fd == -1)
+    dwError = TDNFFileReadAllText(filename, &readbuf);
```
sidcha: This read, pulls the entire contents of the file into memory. The preconditions when implementing this class of methods is that the target file could be very large and we read fixed size buffers into memory and then call the upadte_fn on each of those chunks.

Although the current files we are targeting may be small, we'd still like to keep this API free from such assumptions. Please revert this change.

client/remoterepo.c
sidcha: Does, `metalink_parse_final()` also get rid of the `metalink_context`? I'm not familiar with this API, but this looks buggy from an API standpoint as`metalink_parser_context_delete()`must be called.

pytests/CMakeLists.txt
sidcha: This is indeed very handy during development, but we would like to see all failures of a current commit in CI logs. Can you add add another target, `check_all` that doesn't pass this option and update the CI scripts to use that target instead?

Or better, can you work your way such that this flag isn't passed in CI based on some docker ENV? Let's not merge this as is.

sidcha (changes requested):

client/remoterepo.c
keerthanakalyan: I agree with Siddharth's comment. We should not read complete contents file into memory.

pytests/CMakeLists.txt
sshedi: Never mind, I'll remove it. I thought what's the point of continuing when we hit an error, we need to fix all failures anyway. Like compilation errors, fix one by one.

client/remoterepo.c
sshedi: Yes, metalink_parse_final does metalink_parser_context_delete. Check man page: https://www.systutorials.com/docs/linux/man/3-metalink_parse_final/

client/remoterepo.c
sshedi: Yes, I'm aware of the fact that it could be a large file.

I have made this change considering our use case.

At max what might the metalink file size? 16K ? 256K? we can afford that with the kind of RAM we have.
This is ultimately a space time complexity decision.

Let's say we are reading a 12K file in 1K chunks. On every read context switch happens from user to kernel space and before that from our program we have to go through lots of glibc wrappers and same things on the way back with lots of error checking, and we may end up getting page fault upon reading
every 4K and that has to be taken care every time.

Instead with a single context switch we can get all this done with a little bit of extra memory, I think it should be okay. If I'm not wrong the update functions in hash will be using a realloc upon every call(once we cross the default alloc size), so we end up using (file size * 2) bytes of RAM and by saving lot of time and context switches. All we need to sacrifice is a little bit more memory for fraction of a second.

And FYI, if we use huge files(in terms of GBs) we are screwed anyway, hash APIs will use more RAM and we can't help it. End of the day, we need to use the tools based on our requirement. And in this case that's the whole point of existence for TDNFFileReadAllText function.

@tapakund  & @ppadmavilasom - do we have any special case where we process huge metalink files?
Correct me if I'm wrong. If this doesn't convince you guys, I'll revert the changes.

client/remoterepo.c
tapakund: @cbsiddharth yes @sshedi is right.

client/remoterepo.c
tapakund: @sshedi this API is to get the shasum for the file path passed as an argument.
This doesn't get the shasum for the metalink file, but the file which is downloaded using metalink file(repomd.xml as of now) and verify with the shasum mentioned in the metalink file.
I hope this API can be used in future for getting the shasum in a generic way and not limited to metalink file downloads. So, i think it would be more generic to keep reading the file in chunks as this will avoid our assumptions on the file size before calling this API. Let me know your thoughts as well.

sshedi: I agree with your point, IMO we should do some simple decision making before reading file.
Can we just check for file size before reading it and based on it's size we can decide whether we need to read in chunks or we can read whole file. What say you?

sidcha: > hash APIs will use more RAM and we can't help it.

No, hashing is a primarily CPU intensive task. In Fact, they try to be religious in not using unnecessary space.

> This is ultimately a space time complexity decision.

What are you _trading off_ on actually? few system calls? and yet you choose to do a user space malloc (instead of fixed buffers) that have no deterministic return times what-so-ever. As Knuth rightly said, "premature optimization is the root of all evil".

> At max what might the metalink file size? 16K ? 256K?

These are the assumptions you are making (1. metalink file only; 2. file has predictable sizes). When creating methods that provide functionality, you try to be free from such assumptions. Again that is why the openssl methods have the update_fn concept.

> update functions in hash will be using a realloc upon every call

I'm not sure what makes you think this. One-way-functions typically don't need do any look-ahead (or behind for that matter). I can even be so bold and claim that the update_fn operates in constant space because there is only so-much to it.

> And FYI, if we use huge files(in terms of GBs) we are screwed anyway

Ofcourse not. Hope that settles it.

sshedi: You're right about hash algorithms not using more memory. That's an excellent point to learn at this point. Thanks for that.

What are you trading off on actually? few system calls?
- Time is my main concern here, I'm not strongly advocating for malloc. Yes, it's my bad that I assumed file size will be small so I felt malloc would be suffice. My major concern here is how can avoid loop and repeated context switching. Now I'm back with data.

For a 6.5G file, in a Ubuntu 18.04 host with 32G RAM and 64 core CPU here are the results:

**Using buffer and update method(sha1):**
➜ $ time ./sha1.out
17aa2b1ee20e977fb4f8f8391563f57c3e456361

real	0m10.410s
user 0m9.251s
sys	0m1.156s

**Using mmap and calling a single update(sha1):**
➜ $ time ./sha1.out
17aa2b1ee20e977fb4f8f8391563f57c3e456361

real	0m9.631s
user 0m9.310s
sys	0m0.320s

Yes, we may have to modify TDNFFileReadAllText to use mmap instead of calloc.
But results show a performance gain of 1 second which is obviously good.
Here is my test code: https://controlc.com/d1b98e12
If you want to quote Knuth, quote him fully :)
**We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil. Yet we should not pass up our opportunities in that critical 3%.**
With 1 second performance improvement it's up to you whether it falls in that 3% category or not :)
I'm okay with reverting changes but I really wanted to know this can be optimized or not. I rest my case here.

sidcha: First off, really appreciate your efforts to prove your point and this discussion in general.

About the test cases, No one is arguing that single chunk read is worse than multiple chunk reads. Obviously, if you **can** read the entire file into memory, then that would be faster than reading it partially. The question really is: **what if the machine has only 4GB of ram?** The chunked read would perform just the same way while the mmap/malloc method would fail. And this is genuine use case, not every machine what we run tdnf on has 32GB of memory and its very common these days to have files to hash that are larger than 4GB.

I also saw your suggestion of doing both, (switching between approach based on file size). Here is my reasoning for not doing that:
* Why not increase the chunk size to some reasonable number (based on usage patterns) and settle with that? it achieves best of both worlds. Trust me a big part of the internet is built this way (HTTP, TCP, and a bunch of other protocols).
* For people who are already familiar with with such methods, we are introducing a "what-is-going-here" moment. Then you'd have to put in a comment to explain why you are doing things this way. This is called cognitive load on readers, which we generally try to avoid.

In conclusion, I'd still suggest you do not spend your optimization efforts here. And I strongly believe this falls well within the 97% that you quoted (just do a tdnf install XXX and see how long it takes to get back with the y/n prompt, clearly there are more performance bottlenecks than this?) I'd like to have a keep it simple approach to this. But then again, that's just my thoughts. I'll let you guys decide.

client/remoterepo.c
sidcha: Ok, Thanks.

client/remoterepo.c
tapakund: @sshedi i agree reading the file at once reduces the time taken for this API but if we consider few more aspect like, 1) this API is not called all the time we run tdnf binary, it will be only called when we do "tdnf makecache" or "refresh" the metadata. 2) We will have to check for the size and still be unsure of the machine memory at that point. Also, coming up with a file size number is not a generic way to deal with. So, i feel its better to read in chunks. Let me know your thoughts as well.

sshedi: @tapakund @cbsiddharth please check my latest change set, which uses mmap.

sidcha: @sshedi, How does this address the large file - low memory concern we all are raised?

sshedi: mmap doesn't do any memory allocation and it creates a page-wise mapping and it's much faster than read/fread and it won't increase memory usage

sidcha: @sshedi, I do understand how mmap works. That's not what we are asking. Now instead of favouring smaller files, you are favouring larger files. The question still stands open.

In the first approach where you were reading the entire file into memory, they kernel swaps out pages that are not needed in the process's virtual memory to make do with you request. Here, the kernel will page fault and switch page. Both the approaches favour one kind of datasets. 

The one which malloc-ed the entire file into memory had very high performance for smaller files as you did only one read. The mmap solution favous the larger files as the setup/teardown time for small files are much higher now. The amortized expense of both the approach cannot be predicted unless we have some hints on what sort of data we are dealing with (which we don't have/don't want to make).

What you doing here, in comparison to what @tapakund did is different. Not better. I still vote to revert and leave it as is.

sshedi: `The one which malloc-ed the entire file into memory had very high performance for smaller files as you did only one read.`
It did not. It ended up with same page faults and lots of context switches. mmap definitely improves the performance whether it's big/small whatever file it is. mmap solves this problem as a whole. Just write a simple app to do read and do the same using mmap, you'll see the difference(doesn't matter small or big file). I have given the statistics already and it holds good for small and big file both. **this indeed does a better job for this task**
if you're still not seeing it, I don't know what to say. I'll delete this PR.

PR #134, opened 2020-05-06
Fix stale solv cache issue

keerthanakalyan:
This fix is to remove the old repomd , solv cache and lastrefresh
marker when there is a new repomd file downloaded. This fixes the issue
https://github.com/vmware/photon/issues/1008 

Signed-off-by: Keerthana K <keerthanak@vmware.com>

### DIFF
--- a/client/repoutils.c
+++ b/client/repoutils.c
@@ -482,22 +482,49 @@ error:
 
 uint32_t
 TDNFRemoveTmpRepodata(
-    const char* pszTmpRepodataDir,
-    const char* pszTmpRepoMDFile
+    const char* pszTmpRepodataDir
     )
 {
     uint32_t dwError = 0;
+    char* pszFilePath = NULL;
+    DIR *pDir = NULL;
+    struct dirent *pEnt = NULL;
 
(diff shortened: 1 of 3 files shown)


client/repoutils.c
```diff
@@ -482,22 +482,49 @@ TDNFRemoveRpmCache(
...
+                    pEnt->d_name);
+        if(pszFilePath)
+        {
+            if(unlink(pszFilePath))
+            {
+                dwError = errno;
+                BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+            }
+            TDNF_SAFE_FREE_MEMORY(pszFilePath);
+            pszFilePath = NULL;
```
tapakund, line 523: TDNF_SAFE_FREE_MEMORY will set pszFilePath to NULL.

#define TDNF_SAFE_FREE_MEMORY(pMemory) \
    do {                                                           \
        if (pMemory) {                                             \
            TDNFFreeMemory(pMemory);                               \
            pMemory = NULL;                                        \
    } while(0)

tapakund (approved):
Changes looks good to me. One minor comment.

PR #136, opened 2020-05-24
Remove duplicate code

sshedi:
Cleaned up redundant macros
Use common/includes.h everywhere
Fixed few function prototypes
Added few missing validations
Use ARRAY_SIZE, UNUSED from a common header file
Added and rectified few test scripts


sshedi:
I ended up touching a lot of files, added few more test scripts, other than that there is nothing tricky/difficult to understand things in this PR

pytests/tests/test_downgrade.py
```diff
@@ -0,0 +1,55 @@
...
+    teardown_test(utils)
+
+
+def teardown_test(utils):
+    utils.run(['tdnf', 'erase', '-y', 'tdnf-test-two', 'tdnf-test-multiversion'])
+
+
+def test_downgrade_no_arg(utils):
+    ret = utils.run(['tdnf', 'downgrade'])
+    assert(ret['retval'] == 1011)
```
keerthanakalyan: why do we expect a non zero return code here? Downgrade command should pass.

pytests/tests/test_history.py
```diff
@@ -0,0 +1,69 @@
+#
+# Copyright (C) 2019 VMware, Inc. All Rights Reserved.
```
keerthanakalyan: Please remove this test as there is no history command as of now in tdnf.

tools/cli/lib/help.c
```diff
@@ -80,20 +82,23 @@ TDNFCliShowHelp(
...
+uint32_t
 TDNFCliShowNoSuchCommand(
-    const char* pszCmd
+    const char *pszCmd
     )
 {
-    printf("No such command: %s. Please use /usr/bin/tdnf --help\n", pszCmd);
+    printf("No such command: %s. Please use /usr/bin/tdnf --help\n",
+            pszCmd ? pszCmd : "");
+    return ERROR_TDNF_CLI_NO_SUCH_CMD;
```
keerthanakalyan: If there is a change in the return code, please take care of the harness tests before this change is consumed in photon.

tapakund (commented):
Apart from comments from @keerthanakalyan  Changes looks good.

pytests/tests/test_downgrade.py
sshedi: I ended up fixing the test tear down part of various scripts, still in progress, will push the proper code soon.

pytests/tests/test_check_skip.py
```diff
@@ -8,56 +8,46 @@
...
 
-def run_cmd(utils, cmd, retval):
+def run_cmd(utils, opt, retval):
+    cmd = ['tdnf', '--config', '/etc/tdnf/tdnf.conf', 'check']
+    # DIST will be set in docker environment
+    # To run locallly DIST should be exported manually
+    if os.environ['DIST'] != 'photon':
+        cmd.pop(1)
+        cmd.pop(1)
+    cmd.extend(opt)
```
sidcha: What are we trying to do here? Do you intend to override the settings passed from `config['repo_path']/tdnf.conf`? If so why?

pytests/tests/test_check_skip.py
```diff
@@ -8,56 +8,46 @@
...
-isNotPhoton = -1
-try:
-    isNotPhoton = subprocess.check_call("uname -a | grep -i photon > /dev/null", shell = True)
-except Exception as e:
-    isNotPhoton = 1
 
-def run_cmd(utils, cmd, retval):
+def run_cmd(utils, opt, retval):
+    cmd = ['tdnf', '--config', '/etc/tdnf/tdnf.conf', 'check']
+    # DIST will be set in docker environment
```
sidcha: If "DIST" is not set, default to photon.

pytests/tests/test_downgrade.py
sshedi: Everything is taken care now and all tests are passed.

sidcha (approved):
@sshedi, nice work. Some suggestions,

- Please try to keep PRs small. Split them into multiple smaller PRs if you must. I know it's not very complicated but still, when 42 files are modified, we may miss something.
- Do one thing in one commit. This is actually super critical and lot of upstreams are very picky about this. The reason being we can revert individual commits when they cause an issue. Here, this commit has some C header re-org, code clean up, test additions, and test clean up.

So far, this patch looks good. @keerthanakalyan please merge this soon as I need some of these changes for another work.

pytests/tests/test_check_skip.py
sshedi: made it right

pytests/tests/test_check_skip.py
sshedi: Done

sshedi:
> @sshedi, nice work. Some suggestions,
> 
>     * Please try to keep PRs small. Split them into multiple smaller PRs if you must. I know it's not very complicated but still, when 42 files are modified, we may miss something.
> 
>     * Do one thing in one commit. This is actually super critical and lot of upstreams are very picky about this. The reason being we can revert individual commits when they cause an issue. Here, this commit has some C header re-org, code clean up, test additions, and test clean up.
> 
> 
> So far, this patch looks good. @keerthanakalyan please merge this soon as I need some of these changes for another work.

I started with something, ended up doing lots of other things, so result was 42 files. I don't like such PRs either

sidcha (approved):

sidcha:
@keerthanakalyan, can we merge this? Has some test fix I need.

keerthanakalyan (approved):

PR #137, opened 2020-05-24
Always run single instance of tdnf

sshedi:

### DIFF
--- a/tools/cli/defines.h
+++ b/tools/cli/defines.h
@@ -21,6 +21,8 @@
 
 #pragma once
 
+#define TDNF_INSTANCE_LOCK_FILE     "/var/run/lock/.tdnf-instace-lockfile"
+
 #define BAIL_ON_CLI_ERROR(unError) \
     do {                                                           \
         if (unError)                                               \
--- a/tools/cli/main.c
+++ b/tools/cli/main.c
@@ -21,8 +21,14 @@
 
 #include "includes.h"
 
-TDNF_CLI_CONTEXT _context = {0};
 int nQuiet = 0;
+TDNF_CLI_CONTEXT _context = {0};
+
+/* glockfd used in exit handler nad inst*/
+static int32_t glockfd = 0;
+
+static void TdnfExitHandler(void);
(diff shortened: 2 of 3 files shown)


tools/cli/defines.h
```diff
@@ -21,6 +21,8 @@
 
 #pragma once
 
+#define TDNF_INSTANCE_LOCK_FILE     "/tmp/.tdnf-instace-lockfile"
```
sidcha: Can we stick to conventions put this inside inside `/var/lock/` or `/var/run/lock/`

tools/cli/main.c
```diff
@@ -63,13 +69,26 @@ int main(int argc, char* argv[])
...
+    /*
+     * granular permissions for non root users are pending.
+     * blocking all operations for non root and show the
+     * right error to avoid confusion
+     */
+    if (geteuid())
     {
-        dwError = ERROR_TDNF_PERM;
-        BAIL_ON_CLI_ERROR(dwError);
+        BAIL_ON_CLI_ERROR((dwError = ERROR_TDNF_PERM));
```
sidcha: This looks little aggressive :). Even if its not as granular as it must be, we must to this at a per-command level at the very least. Some commands like tdnf list should be runnable by any user.

tools/cli/main.c
```diff
@@ -63,13 +69,26 @@ int main(int argc, char* argv[])
...
+    }
+
+    if (IsTdnfAlreadyRunning())
```
sidcha, line 83: Again, only modify operations needs to take this lock. We must defer this call further down, maybe at per-comand bassis.

sidcha (changes requested):
Some more work is needed. But a step in the good direction. :)

Thanks!

tools/cli/main.c
sshedi: This should be a separate enhancement. Let's do it later

ppadmavilasom (approved):
LGTM

keerthanakalyan (approved):

PR #139, opened 2020-05-26
Add cmake check for openssl-devel

tapakund:
We should check if system have openssl-devel installed
before building tdnf.


keerthanakalyan (commented):
LGTM

sshedi (approved):

sidcha (approved):

PR #144, opened 2020-06-05
Serve test-repo RPMs over a super simple http server

sidcha:
Now, test-repo is served at http://localhost:8080/ on the test/dev
machine. Hopefully, this get tests closer to actual usage scenario.

Also, fix tests/test_downgrade.py: test_downgrade_no_arg() and add
new key "distribution" to utils.config so tests can perform some
distribution specifc tests (or outcome asserts).

### DIFF
--- a/pytests/repo/setup-repo.sh
+++ b/pytests/repo/setup-repo.sh
@@ -64,8 +64,8 @@ echo ${GPG_PASS} | gpg --batch --passphrase-fd 0 \
 cat << EOF > ${TEST_REPO_DIR}/yum.repos.d/photon-test.repo
 [photon-test]
 name=basic
-baseurl=file://${PUBLISH_PATH}
-#metalink=file://${PUBLISH_PATH}/metalink
+baseurl=http://localhost:8080/photon-test
+#metalink=http://localhost:8080/photon-test/metalink
 gpgkey=file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
 gpgcheck=0
 enabled=1
@@ -88,7 +88,7 @@ cat << EOF > ${PUBLISH_PATH}/metalink
    <verification>
    </verification>
    <resources maxconnections="1">
-    <url protocol="file" type="file" location="IN" preference="100">file://${PUBLISH_PATH}/repodata/repomd.xml</url>
+    <url protocol="http" type="file" location="IN" preference="100">http://localhost:8080/photon-test/repodata/repomd.xml</url>
    </resources>
   </file>
  </files>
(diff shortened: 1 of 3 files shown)


pytests/repo/setup-repo.sh
```diff
@@ -88,7 +88,7 @@ cat << EOF > ${PUBLISH_PATH}/metalink
    <verification>
    </verification>
    <resources maxconnections="1">
-    <url protocol="file" type="file" location="IN" preference="100">file://${PUBLISH_PATH}/repodata/repomd.xml</url>
+    <url protocol="file" type="file" location="IN" preference="100">http://localhost:8080/photon-test/repodata/repomd.xml</url>
```
tapakund: Sid, pls change the protocol and type to http.

tapakund (changes requested):

pytests/tests/test_downgrade.py
```diff
@@ -22,15 +22,8 @@ def teardown_test(utils):
...
-    ret = utils.run(['tdnf', '--config', '/etc/tdnf/tdnf.conf', '--enablerepo=photon-debuginfo', '--assumeno', 'downgrade'])
-    try:
-        if os.environ['DIST'] != 'photon':
-            assert(ret['retval'] == 1602)
-        else:
-            assert(ret['retval'] == 8)
-    except Exception as e:
-        # Default to Photon
-        assert(ret['retval'] == 8)
+    ret = utils.run(['tdnf', 'downgrade'])
```
sshedi, line 25: The reason I used different config for Photon was, we install glibc-debuginfo in test environment and with that 'tdnf downgrade' should try to downgrade everything and with assume assumeno we should get 1602 as retval.
TBH, I didn't knew what 'tdnf downgrade' without args will do and @keerthanakalyan gave me insight on how it is supposed to work so I tried to achieve it.

sidcha: Ohh, I see. Then we must probably call it `test_downgrade_everything` and create some new dummy spec files that simulate this situation. For now, let the `test_downgrade_no_args` be consistent with it's naming (with other such tests everywhere). We'll put in a TODO to add the other test.

pytests/repo/setup-repo.sh
sidcha: Done. Thanks!

tapakund: I don't see the changes reflected in the commit.

keerthanakalyan (approved):

tapakund (approved):

keerthanakalyan (approved):

PR #145, opened 2020-06-05
Don't check for providers when achieving goal

sshedi:
tdnf used to report false positives on failure to get providers while
using 'tdnf check', the fix for that issue didn't took care of the issue
while achieving goal. This fixes that issue buy reporting missing
providers while insatlling a package.

To put it in simple terms, if we are trying to install a package and a
provider is missing, it should be reported.


sshedi:
I feel we should have a test case for this one, please provide some inputs. @cbsiddharth can we add a new spec file to regenerate this issue?
But I'm a bit concerned about this one, sometimes this may be a spec file issue and we need to fix the spec file. How can we detect such mistakes?

sidcha:
@sshedi, you are right, we must create a spec file that introduces this issue and test if we catch it. As to spec file having issues, so long as we are in the right side (from a TDNF contract perspective) of things, we don't (and can't) have to worry about it.

Commit Message:

> using 'tdnf check', the fix for that issue didn't took care of the issue
> while achieving goal. This fixes that issue buy reporting missing
> providers while insatlling a package.

s/took/take/
s/buy/by/
s/insatlling/installing/

include/tdnftypes.h
```diff
@@ -144,9 +144,10 @@ typedef enum
...
 typedef enum
 {
-    SKIPPROBLEM_NONE=0,
-    SKIPPROBLEM_CONFLICTS,
-    SKIPPROBLEM_OBSOLETES
+    SKIPPROBLEM_NONE = 0,
+    SKIPPROBLEM_CONFLICTS = 0x1,
+    SKIPPROBLEM_OBSOLETES = 0x2,
+    SKIPPROBLEM_IGNORE = 0x4
 }TDNF_SKIPPROBLEM_TYPE;
```
sidcha: While we are at this, I'd also like a SPACE char between `}` and `TYPE_NAME`. This instance is fine, no need to fix it but just saying.

solv/tdnfpackage.c
```diff
@@ -1683,10 +1681,6 @@ SolvReportProblems(
     {
         fprintf(stderr, "Found %u problem(s) while resolving\n", total_prblms);
     }
-    else
-    {
-        printf("No problems found while resolving\n");
```
sidcha, line 1688: This is changing the observable behaviour of TDNF in general. Can you please make sure nothing is depending on this text output? for instance, out internal test suite does a lot of checks based on the output strings.

sidcha (approved):
LGTM!

keerthanakalyan:
> @sshedi, you are right, we must create a spec file that introduces this issue and test if we catch it. As to spec file having issues, so long as we are in the right side (from a TDNF contract perspective) of things, we don't (and can't) have to worry about it.
> 
> Commit Message:
> 
> > using 'tdnf check', the fix for that issue didn't took care of the issue
> > while achieving goal. This fixes that issue buy reporting missing
> > providers while insatlling a package.
> 
> s/took/take/
> s/buy/by/
> s/insatlling/installing/

@sshedi : some of the testcases that I could think of would be : 
1. You can create a spec file with dummy dependency (in Requires:) and do not have the dummy package in the repo. TDNF should report a dependency missed error. 
2. Other positive tests include test for conflicting files provided from 2 packages and trying to install an obsolete package.

keerthanakalyan (approved):
LGTM.

solv/tdnfpackage.c
```diff
@@ -1655,24 +1660,22 @@ SolvReportProblems(
...
+            SkipBasedOnType(type, dwSkipProblem))
         {
             continue;
         }
 
         pszProblem = solver_problemruleinfo2str(pSolv, type, dwSource,
                                                 dwTarget, dwDep);
 
-        if (type == SOLVER_RULE_PKG_REQUIRES)
+        if (dwSkipProblem != SKIPPROBLEM_NONE &&
```
jaankit, line 1672: How does SOLVER_RULE_PKG_REQUIRES type is related to SKIPPROBLEM_CONFLICTS and OBSOLETES?
Why are we adding dwSkipProblem != SKIPPROBLEM_NONE check?
As per my understanding, If --skip conflicts and obsolete are not provided then dwSkipProblem should be SKIPPROBLEM_NONE and if type is SOLVER_RULE_PKG_REQUIRES, in that case also, flow should go inside if condition.

jaankit (commented):
Just a small query.

solv/tdnfpackage.c
sshedi: This is not related to skipping conflicts or obsoletes. TDNF reports few false positives saying 'none of the providers can be installed'. For that I added one more check to confirm whether a provider is really not available.

Yes, with 'tdnf check' dwSkipProblem value will be NONE and we don't check for providers. That way we will report everything (including false positives), if there are any spec file issues we can address them with this method.
Or else we will end up blocking everything related to providers.

```
Check: pytests/tests/test_check_skip.py
def test_dummy_check(utils):
    ret = utils.run(['tdnf', 'check'])
    assert ' obsoletes ' in ret['stderr'][0]
    assert ' conflicts ' in ret['stderr'][1]
    assert ' provides ' in ret['stderr'][2]
```

tapakund (approved):
LGTM! Just one general comment, if we can do cleanup(typos fixes, space indentation) in separate PR that would be great.

sshedi:
@tapakund Agreed but I saw it and did it anyway. I'll follow it from next time on-wards.

solv/tdnfpackage.c
jaankit: As discussed, This is just a work around to skip check_for_providers().
Actual issue need to be fixed  in  check_for_providers().

jaankit (approved):

PR #152, opened 2020-07-23
Fixed an issue related to signature validation

sshedi:
gpgme_op_verify_result can succeed but returns a list of zero signatures
Enabled TDNF_PLUGIN_EVENT_TYPE_REPO_MD event for repogpgcheck plugin

### DIFF
--- a/plugins/repogpgcheck/repogpgcheck.c
+++ b/plugins/repogpgcheck/repogpgcheck.c
@@ -16,18 +16,18 @@ _TDNFVerifyResult(
     )
 {
     uint32_t dwError = 0;
-    gpgme_verify_result_t pResult = NULL;
     gpgme_signature_t pSig = NULL;
+    gpgme_verify_result_t pResult = NULL;
 
     /* pContext release will free pResult. do not free otherwise */
     pResult = gpgme_op_verify_result(pContext);
-    if (!pResult)
+    if (!pResult || !pResult->signatures)
     {
         dwError = ERROR_TDNF_GPG_VERIFY_RESULT;
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    for(pSig = pResult->signatures; pSig; pSig = pSig->next)
+    for (pSig = pResult->signatures; pSig; pSig = pSig->next)
     {
         if (pSig->status)
         {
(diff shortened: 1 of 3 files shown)


plugins/repogpgcheck/repogpgcheck.c
```diff
@@ -21,13 +21,13 @@ _TDNFVerifyResult(
 
     /* pContext release will free pResult. do not free otherwise */
     pResult = gpgme_op_verify_result(pContext);
-    if (!pResult)
+    if (!pResult || !(pSig = pResult->signatures))
```
oliverkurth: I think it would look cleaner to use `if (!pResult || !pResult->signatures)` and then leave the `for` loop unchanged.

pytests/tests/test_repogpgcheck.py
```diff
@@ -67,5 +67,8 @@ def test_tdnfrepogpgcheck_plugin_validatesignature(utils):
     ret = utils.run([ 'tdnf', 'repolist', '--refresh'])
     #we should load the plugin
     assert(ret['stdout'][0].startswith('Loaded plugin: tdnfrepogpgcheck')) #nosec
-    #we should also pass the repolist command
-    assert(ret['retval'] == 0) #nosec
+    dist = os.environ.get('DIST')
+    if dist == 'photon':
+        assert(ret['retval'] == 0)
+    elif dist == 'fedora':
+        assert(ret['retval'] == 86)
```
ppadmavilasom, line 77: try and avoid magic numbers or please add comment. if its something we need to fix, please open an issue to track

plugins/repogpgcheck/repogpgcheck.c
sshedi: Wanted to save a read :) fixed it

pytests/tests/test_repogpgcheck.py
sshedi: 'tdnf repolist --refresh' fails with "Unable to parse metalink, ERROR: code=201", I have informed @tapakund  about it already. Left a comment in the source file & totally agree with your comment.

keerthanakalyan: @sshedi : Could you please raise it under open issues?

keerthanakalyan: Issue has been raised for this failure https://github.com/vmware/tdnf/issues/165

keerthanakalyan (approved):

PR #156, opened 2020-08-11
Allow alter type ALTER_UPGRADE to call SolvAddExcludes()

sidcha:
`TDNFPrepareAllPackages()` forces `nAlterType` to `ALTER_UPGRADE` for both `ALTER_UPGRADE` and `ALTER_UPGRADEALL` when `--security` is passed. Subsequently, when `TDNFGoal()` is called, the excludes are skipped from the job queue in `TDNFAddGoal()` but are not marked in solver map as excludes with a call to `SolvAddExcludes()` as it's gating condition only check for `nAlterType == ALTER_UPGRADEALL`.

Fix this by allowing `nAlterType == ALTER_UPGRADE` to also call `SolvAddExcludes()`.

### DIFF
--- a/client/goal.c
+++ b/client/goal.c
@@ -464,7 +464,6 @@ TDNFAddGoal(
     char** ppszPackagesTemp = NULL;
     char* pszName = NULL;
     uint32_t dwCount = 0;
-    Queue queueJob = {0};
 
     if(!pQueueJobs || dwId == 0 || !pTdnf->pSack || !pTdnf->pSack->pPool)
     {
@@ -472,7 +471,6 @@ TDNFAddGoal(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    queue_init(&queueJob);
     if (nAlterType == ALTER_UPGRADE)
     {
         dwError = TDNFPkgsToExclude(pTdnf, &dwCount, &ppszExcludes);
@@ -500,6 +498,7 @@ TDNFAddGoal(
             }
         }
     }
+
     switch(nAlterType)
     {
         case ALTER_DOWNGRADEALL:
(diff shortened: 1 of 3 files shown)


sidcha:
@keerthanakalyan, I want these commits merged as 2 separate commits for ease of blaming at a later time.

CC: @sshedi, @tapakund, @jaankit

ppadmavilasom (approved):
LGTM

client/goal.c
```diff
@@ -356,7 +356,8 @@ TDNFGoal(
     dwError = SolvAddFlagsToJobs(&queueJobs, nFlags);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    if (nAlterType == ALTER_UPGRADEALL)
+    if (nAlterType == ALTER_UPGRADEALL ||
+        nAlterType == ALTER_UPGRADE)
```
keerthanakalyan, line 360: This change fails for the packages which includes subpackages. For example if we provide --exclude=curl we get an error like below, since it is not able to exlcude curl-libs from the list.
```
]# ./tdnf update --security --exclude=curl
1. package curl-7.61.1-2.ph3.x86_64 is disabled
Found 1 problem(s) while resolving
Error(1301) : Solv general runtime error
```

sidcha: That is was a valid error. We must skip this error where there are excludes.

keerthanakalyan (approved):

solv/tdnfpackage.c
```diff
@@ -1534,38 +1520,45 @@ SolvSplitEvr(
...
+                 type == SOLVER_RULE_PKG_SELF_CONFLICT;
     }
 
-    if (dwSkipProblem == SKIPPROBLEM_OBSOLETES)
+    if (dwSkipProblem & SKIPPROBLEM_OBSOLETES)
     {
-        return (type == SOLVER_RULE_PKG_OBSOLETES ||
-                type == SOLVER_RULE_PKG_IMPLICIT_OBSOLETES ||
-                type == SOLVER_RULE_PKG_INSTALLED_OBSOLETES);
+        result = result || type == SOLVER_RULE_PKG_OBSOLETES ||
```
sshedi, line 1540: Cool 👍

sshedi (approved):
LGTM

PR #157, opened 2020-08-14
Add support for excludepkgs option in tdnf.conf

sidcha:
This patch adds the config option "excludepkgs" from tdnf.conf and some
utils to make this possible neatly.

This also adds a new TDNFReAllocateMemory() that mimics realloc() but
also guarantees that the newly reallocated memory is also zero-ed to
meet the general tdnf codebase expectation.

### DIFF
--- a/client/client.c
+++ b/client/client.c
@@ -77,7 +77,7 @@ TDNFPkgsToExclude(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    if (pTdnf->pConf->ppszExcludes)
+    if (!pTdnf->pArgs->nDisableExcludes && pTdnf->pConf->ppszExcludes)
     {
         if (!pTdnf->pArgs->nQuiet)
         {
@@ -115,7 +115,7 @@ TDNFPkgsToExclude(
         BAIL_ON_TDNF_ERROR(dwError);
 
         nIndex = 0;
-        if (pTdnf->pConf->ppszExcludes)
+        if (!pTdnf->pArgs->nDisableExcludes && pTdnf->pConf->ppszExcludes)
         {
             while (pTdnf->pConf->ppszExcludes[nIndex])
             {
--- a/client/client.c
+++ b/client/client.c
@@ -77,6 +77,24 @@ TDNFPkgsToExclude(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    if (pTdnf->pConf->ppszExcludes)
+    {
+        if (!pTdnf->pArgs->nQuiet)
+        {
+            printf("Warning: The following packages are excluded "
+                   "from tdnf.conf:\n");
+        }
+        while (pTdnf->pConf->ppszExcludes[nIndex])
+        {
+            if (!pTdnf->pArgs->nQuiet)
+            {
+                printf("  %s\n", pTdnf->pConf->ppszExcludes[nIndex]);
+            }
+            dwCount++;
+            nIndex++;
+        }
+    }
+
     pSetOpt = pTdnf->pArgs->pSetOpt;
     while(pSetOpt)
     {
@@ -95,6 +113,20 @@ TDNFPkgsToExclude(
                       sizeof(char*),
                       (void**)&ppszExcludes);
         BAIL_ON_TDNF_ERROR(dwError);
+
+        nIndex = 0;
+        if (pTdnf->pConf->ppszExcludes)
+        {
+            while (pTdnf->pConf->ppszExcludes[nIndex])
+            {
+                dwError = TDNFAllocateString(pTdnf->pConf->ppszExcludes[nIndex],
+                                             &ppszExcludes[nIndex]);
+                BAIL_ON_TDNF_ERROR(dwError);
+                dwCount++;
+                nIndex++;
+            }
+        }
+
         pSetOpt = pTdnf->pArgs->pSetOpt;
         while(pSetOpt)
         {
@@ -105,7 +137,6 @@ TDNFPkgsToExclude(
                       pSetOpt->pszOptValue,
                       &ppszExcludes[nIndex++]);
                 BAIL_ON_TDNF_ERROR(dwError);
-
             }
             pSetOpt = pSetOpt->pNext;
         }
--- a/client/client.c
+++ b/client/client.c
@@ -71,6 +71,7 @@ TDNFPkgsToExclude(
     uint32_t dwCount = 0;
     char**   ppszExcludes = NULL;
     int nIndex = 0;
+
     if(!pTdnf || !pTdnf->pArgs || !pdwCount || !pppszExcludes)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
--- a/client/defines.h
+++ b/client/defines.h
@@ -79,6 +79,7 @@ typedef enum
 #define TDNF_CONF_KEY_PLUGIN_PATH         "pluginpath"
 #define TDNF_CONF_KEY_PLUGIN_CONF_PATH    "pluginconfpath"
 #define TDNF_PLUGIN_CONF_KEY_ENABLED      "enabled"
+#define TDNF_CONF_KEY_EXCLUDE             "excludepkgs"
 //Repo file key names
 #define TDNF_REPO_KEY_BASEURL             "baseurl"
 #define TDNF_REPO_KEY_ENABLED             "enabled"
--- a/common/configreader.c
+++ b/common/configreader.c
@@ -594,6 +594,47 @@ error:
 }
 
 uint32_t
+TDNFReadKeyValueStringArray(
+    PCONF_SECTION pSection,
+    const char* pszKeyName,
+    char*** pppszValueList
+    )
+{
+    uint32_t dwError = 0;
+    char** ppszValList = NULL;
+    PKEYVALUE pKeyValues = NULL;
+
+    if (!pSection || !pszKeyName || !pppszValueList)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pKeyValues = pSection->pKeyValues;
+    for (; pKeyValues; pKeyValues = pKeyValues->pNext)
+    {
+        if (strcmp(pszKeyName, pKeyValues->pszKey) == 0)
+        {
+            dwError = TDNFSplitStringToArray(pKeyValues->pszValue,
+                                             " ", &ppszValList);
+            BAIL_ON_TDNF_ERROR(dwError);
+            *pppszValueList = ppszValList;
+            break;
+        }
+    }
+
+cleanup:
+    return dwError;
+
+error:
+    if(pppszValueList)
+    {
+        *pppszValueList = NULL;
+    }
+    goto cleanup;
+}
+
+uint32_t
 TDNFReadKeyValue(
     PCONF_SECTION pSection,
     const char* pszKeyName,
@@ -653,4 +694,3 @@ error:
     TDNF_SAFE_FREE_MEMORY(pszValue);
     goto cleanup;
 }
-
--- a/common/strings.c
+++ b/common/strings.c
@@ -88,23 +88,21 @@ error:
     goto cleanup;
 }
 
-size_t
+uint32_t
 TDNFStringSepCount(
     char *pszBuf,
-    char *pszSep
+    char *pszSep,
+    size_t *nSepCount
     )
 {
     size_t nCount = 0;
-    const char *pszTemp;
+    uint32_t dwError = 0;
+    const char *pszTemp = NULL;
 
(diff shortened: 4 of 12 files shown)


sidcha:
CC: @keerthanakalyan, @sshedi, @jaankit, @tapakund, @oliverkurth

client/defines.h
```diff
@@ -79,6 +79,7 @@ typedef enum
 #define TDNF_CONF_KEY_PLUGIN_PATH         "pluginpath"
 #define TDNF_CONF_KEY_PLUGIN_CONF_PATH    "pluginconfpath"
 #define TDNF_PLUGIN_CONF_KEY_ENABLED      "enabled"
+#define TDNF_CONF_KEY_EXCLUDE             "exclude"
```
oliverkurth: Should this be `excludepkgs`? I see `excludepkgs` but not `exclude` in https://dnf.readthedocs.io/en/latest/conf_ref.html .

client/client.c
```diff
@@ -77,6 +77,16 @@ TDNFPkgsToExclude(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    if (pTdnf->pConf->ppszExcludes)
+    {
+        printf("Warning: The following packages are excluded from tdnf.conf:\n");
```
ppadmavilasom: you might have to check if `quiet` flag is on before printing this

common/strings.c
```diff
@@ -88,6 +88,57 @@ TDNFSafeAllocateString(
     goto cleanup;
 }
 
+uint32_t
```
ppadmavilasom, line 132: if split into a count and collect, can avoid realloc

client/client.c
sidcha: Will fix this, thanks. Should we have a method for logging in TDNF (that takes pTdnf)? that can handle these nits instead of this Ad-Hoc prints? I can introduce one if the idea sounds okay.

common/strings.c
sidcha: Here is my justification for using `realloc()` in this instance. I'm open for suggestions/corrections:

- The number of realloc() calls are logarithmic to the target size and we start off with a large chunk (16 in this case) and double each time.
- `strtok()` (or other tokenizing methods `strsep()` etc) cobblers the input string so I'd have to make two copies (`strdup()`-ed) to achieve this effect. This makes the code look ugly :).
- The talingling resize realloc() does no copy or dealloc most of the time.
- This progressive growth pattern is easy understand and looks good.

client/defines.h
sidcha: This was passed down to me in the JIRA, I'll change this to match dnf. I'll also add a new flag `--disableexcludes` to exclude those from config file as this looks useful keep around.

common/strings.c
ppadmavilasom: you dont have to use strtok. you have some separator - space in this case. just count space instances using an advancing ptr and strchr. i think its much cleaner to know what you need to do and then allocate precisely for that.
again, just my opinion.

client/client.c
ppadmavilasom: sounds good.

common/strings.c
sidcha: Switched to count-and-alloc

client/client.c
sidcha: Checked for quite in this instance. Will raise a separate PR for logging.

client/defines.h
sidcha: This is now fixed.

common/strings.c
```diff
@@ -88,6 +88,91 @@ TDNFSafeAllocateString(
     goto cleanup;
 }
 
+size_t
```
oliverkurth: Common style in tdnf code is to return dwError, and return the value in an argument.

common/strings.c
```diff
@@ -88,6 +88,91 @@ TDNFSafeAllocateString(
...
+    const char *pszTemp;
+
+    if (IsNullOrEmptyString(pszBuf))
+    {
+        BAIL_ON_TDNF_ERROR(1);
+    }
+    if (IsNullOrEmptyString(pszSep))
+    {
+        nCount = 1;
+        BAIL_ON_TDNF_ERROR(1);
```
oliverkurth: Here we should use a macros as defined in `include/tdnferror.h `.

Like:
```
dwError = ERROR_TDNF_INVALID_PARAMETER
BAIL_ON_TDNF_ERROR(dwError);
```

common/strings.c
```diff
@@ -88,6 +88,91 @@ TDNFSafeAllocateString(
...
+        }
+        nCount++;
+        while(*pszTemp && !strchr(pszSep, *pszTemp))
+        {
+            pszTemp++;
+        }
+    }
+
+error:
+    return nCount;
```
oliverkurth: return dwError here.

oliverkurth (changes requested):

common/strings.c
sidcha: This method has no errors. An error here is application/usage dependant. For instance someone can pass a string and  check if it does not have any of the give separator chars by check `count == 1`.

The mandate of using `BAIL_ON_TDNF_ERROR` is causing methods that don't need to return anything also return a `uint32_t` just so the caller can also call `BAIL_ON_TDNF_ERROR`. It's fast becoming an anti-pattern IMO. But do let me know if this is a "must be followed".

sidcha:
> Is this option required? Can the user remove it from conf file instead of having this option?

DNF has it so I added it for sake of completeness. And yes it might be useful when you want to install excluded packages in a more controlled manner (without having to modify tdnf.conf). I can think of kernel and other core packages that might fall under this category.

common/strings.c
ppadmavilasom: https://github.com/vmware/tdnf/blob/dev/docs/coding-guidelines.md
Credit to @snambakam

sidcha: Yes, I've seen this. Specifically:

> If a function has the possibility to fail, it must return a 32 bit error code

common/memory.c
```diff
@@ -62,6 +62,53 @@ TDNFAllocateMemory(
     goto cleanup;
 }
 
+uint32_t
+TDNFReAllocateMemory(
```
ppadmavilasom: unused after your recent refactors.

common/strings.c
```diff
@@ -88,6 +88,91 @@ TDNFSafeAllocateString(
...
+        BAIL_ON_TDNF_ERROR(1);
+    }
+    if (IsNullOrEmptyString(pszSep))
+    {
+        nCount = 1;
+        BAIL_ON_TDNF_ERROR(1);
+    }
+
+    pszTemp = pszBuf;
+    while (*pszTemp)
```
ppadmavilasom, line 109: offset = 0;
while((next = strstr(str + offset, sep)))
{
      int len = next - (str + offset);
      //allocate array[arrayindex] element with the len above
      memcpy(array[arrayindex], str + offset, len);
      offset = next - str + strlen(sep);
      ++arrayindex;
}
//do an extra allocatestring with source as str+offset

sidcha: The code posted is copying sections too, but I'm doing this in `TDNFSplitStringToArray()`. This method only gives me a leading count on number of sections that is used to allocate memory upfront (to address your review comment earlier).

BTW, the above code does not handle whitespace/multiple-delimiters in input string correctly while mine does.

common/memory.c
sidcha: Removed. Thanks.

sshedi:
Is it possible to wait for a while? I want to understand this better.

keerthanakalyan (approved):

tools/cli/lib/help.c
```diff
@@ -57,6 +57,7 @@ TDNFCliShowHelp(
     printf("           [--reboot-required]\n");
     printf("           [--skipsignature]\n");
     printf("           [--skipdigest]\n");
+    printf("           [--disableexcludes]");
```
keerthanakalyan: Needs a newline after --disableexcludes.

keerthanakalyan (changes requested):

tools/cli/lib/help.c
sidcha: Good catch. Thanks.

sidcha:
@sshedi, have you had a chance to look into this?

common/configreader.c
```diff
@@ -593,6 +593,47 @@ TDNFReadKeyValueInt(
...
+    PCONF_SECTION pSection,
+    const char* pszKeyName,
+    char*** pppszValueList
+    )
+{
+    uint32_t dwError = 0;
+    char** ppszValList = NULL;
+    PKEYVALUE pKeyValues = NULL;
+
+    if(!pSection || !pszKeyName || !pppszValueList)
```
tapakund: maintain space between if (

common/configreader.c
```diff
@@ -593,6 +593,47 @@ TDNFReadKeyValueInt(
...
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pKeyValues = pSection->pKeyValues;
+    for(; pKeyValues; pKeyValues = pKeyValues->pNext)
+    {
+        if(strcmp(pszKeyName, pKeyValues->pszKey) == 0)
```
tapakund: same here.

common/configreader.c
```diff
@@ -593,6 +593,47 @@ TDNFReadKeyValueInt(
...
+    PKEYVALUE pKeyValues = NULL;
+
+    if(!pSection || !pszKeyName || !pppszValueList)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pKeyValues = pSection->pKeyValues;
+    for(; pKeyValues; pKeyValues = pKeyValues->pNext)
```
tapakund: here

common/strings.c
tapakund: Is this resolved? Current changes still shows we passing hardcoded value.

common/strings.c
```diff
@@ -88,6 +88,91 @@ TDNFSafeAllocateString(
...
+    )
+{
+    uint32_t dwError = 0;
+    char *pszTok = NULL;
+    char *pszState = NULL;
+    char **ppszToks = NULL;
+    size_t nCount = 0;
+    size_t nLength = 0;
+
+    nCount = TDNFStringSepCount(pszBuf, pszSep);
```
tapakund: Better to NULL check the parameters before passing it on to a function.

common/strings.c
```diff
@@ -88,6 +88,91 @@ TDNFSafeAllocateString(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(nCount + 1, sizeof(char *), (void**)&ppszToks);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    pszTok = strtok_r(pszBuf, pszSep, &pszState);
+    while (nLength < nCount && pszTok != NULL)
+    {
+        dwError = TDNFAllocateString(pszTok, &ppszToks[nLength]);
```
tapakund: i think renaming nLength to nIndex makes more sense while going through the code. its just a suggestion, you can take a call.

common/strings.c
sidcha: Actually `BAIL_ON_TDNF_ERROR` here is misleading. It must just be `goto exit;` as this method does not return have an erroring scenario (see thread above).

```c
if (IsNullOrEmptyString(pszBuf))
{
    goto exit;
}
if (IsNullOrEmptyString(pszSep))
{
    nCount = 1;
    goto exit;
}
```
I am not sure if tDNF allows bare word goto to be placed instead of a `BAIL_ON_XXX()`. If it does, I'll go ahead with that.

common/configreader.c
sidcha: Done.

common/configreader.c
sidcha: Done.

common/strings.c
sidcha: Done.

common/configreader.c
sidcha: Done.

sshedi (approved):
LGTM

keerthanakalyan (approved):
LGTM

common/strings.c
oliverkurth: I don't really like his, because it's different to the behaviors of similar functions and therefore maybe confusing. Caller may not know that this function cannot fail, and maybe this will change some time in the future.

I won't insist on the change.

oliverkurth:
LGTM.

common/strings.c
sidcha: Changed.

A caller _must_ know know a function's contract before invoking it.

keerthanakalyan (approved):

PR #158, opened 2020-08-17
Fix gpgkey configured in repo config 

oliverkurth:
When a key is configured for the repo with the "gpgkey" option, the key is ignored. Only keys import with rpm are used.

To reproduce - the step `rpm -e gpg-pubkey-66fd4949-4803fe57` is crucial:
```
root [ /etc/yum.repos.d ]# cat photon.repo 
[photon]
name=VMware Photon Linux $releasever ($basearch)
baseurl=https://dl.bintray.com/vmware/photon_release_$releasever_$basearch
gpgkey=file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
gpgcheck=1
enabled=1
skip_if_unavailable=True
root [ /etc/yum.repos.d ]# ls -l /etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
-rw-r--r-- 1 root root 755 2020-02-15 03:16 /etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
root [ /etc/yum.repos.d ]# rpm -q gpg-pubkey
gpg-pubkey-66fd4949-4803fe57
root [ /etc/yum.repos.d ]# rpm -e gpg-pubkey-66fd4949-4803fe57
root [ /etc/yum.repos.d ]# tdnf install less
Installing:
less                                      x86_64               530-1.ph3                   photon               229.65k 235163
Total installed size: 229.65k 235163
Is this ok [y/N]:y
Downloading:
less                                    121878    100%
Error(1474) : public key is unavailable. install public key using rpm --import or use --nogpgcheck to ignore.
root [ /etc/yum.repos.d ]# 
```

This change first tests the rpm keyring, and if that fails with a signature error adds the configured to a new keyring. The keyring will be discarded so it's not going to be used for other packages from another repo that may have other keys configured.

### DIFF
--- a/client/gpgcheck.c
+++ b/client/gpgcheck.c
@@ -264,7 +264,6 @@ AddKeyToKeyRing(
         }
         BAIL_ON_TDNF_ERROR(dwError);
     }
-
 cleanup:
     TDNF_SAFE_FREE_MEMORY(pszKeyData);
     return dwError;
@@ -376,3 +375,36 @@ error:
     }
     goto cleanup;
 }
+
+uint32_t
+TDNFImportGPGKey(
+    rpmts pTS,
+    const char* pszFile
+    )
+{
+    uint32_t dwError = 0;
+    pgpArmor nArmor = PGPARMOR_NONE;
+    uint8_t* pPkt = NULL;
+    size_t nPktLen = 0;
+    char* pszKeyData = NULL;
+
+    dwError = ReadGPGKey(pszFile, &pszKeyData);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    nArmor = pgpParsePkts(pszKeyData, &pPkt, &nPktLen);
+    if(nArmor != PGPARMOR_PUBKEY)
+    {
+        dwError = ERROR_TDNF_INVALID_PUBKEY_FILE;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = rpmtsImportPubkey(pTS, pPkt, nPktLen);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+cleanup:
+    TDNF_SAFE_FREE_MEMORY(pszKeyData);
+    return dwError;
+error:
+    goto cleanup;
+}
+
--- a/client/gpgcheck.c
+++ b/client/gpgcheck.c
@@ -388,6 +388,12 @@ TDNFImportGPGKey(
     size_t nPktLen = 0;
     char* pszKeyData = NULL;
 
+    if(pTS == NULL || IsNullOrEmptyString(pszFile))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
     dwError = ReadGPGKey(pszFile, &pszKeyData);
     BAIL_ON_TDNF_ERROR(dwError);
 
--- a/client/repo.c
+++ b/client/repo.c
@@ -318,7 +318,7 @@ TDNFGetGPGSignatureCheck(
     uint32_t dwSkipSignature = 0;
     char* pszUrlGPGKey = NULL;
 
-    if(!pTdnf || !ppszUrlGPGKey || IsNullOrEmptyString(pszRepo) || !pnGPGSigCheck)
+    if(!pTdnf || IsNullOrEmptyString(pszRepo) || !pnGPGSigCheck)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
@@ -340,21 +340,25 @@ TDNFGetGPGSignatureCheck(
             nGPGSigCheck = pRepo->nGPGCheck;
             if(nGPGSigCheck)
             {
-                if (IsNullOrEmptyString(pRepo->pszUrlGPGKey))
-                {
-                    dwError = ERROR_TDNF_NO_GPGKEY_CONF_ENTRY;
+                if (ppszUrlGPGKey != NULL) {
+                    if (IsNullOrEmptyString(pRepo->pszUrlGPGKey))
+                    {
+                        dwError = ERROR_TDNF_NO_GPGKEY_CONF_ENTRY;
+                        BAIL_ON_TDNF_ERROR(dwError);
+                    }
+                    dwError = TDNFAllocateString(
+                                 pRepo->pszUrlGPGKey,
+                                 &pszUrlGPGKey);
                     BAIL_ON_TDNF_ERROR(dwError);
                 }
-                dwError = TDNFAllocateString(
-                             pRepo->pszUrlGPGKey,
-                             &pszUrlGPGKey);
-                BAIL_ON_TDNF_ERROR(dwError);
             }
         }
     }
 
     *pnGPGSigCheck = nGPGSigCheck;
-    *ppszUrlGPGKey = pszUrlGPGKey;
+    if (ppszUrlGPGKey) {
+        *ppszUrlGPGKey = pszUrlGPGKey;
+    }
 
 cleanup:
     return dwError;
--- a/client/repo.c
+++ b/client/repo.c
@@ -335,28 +335,27 @@ TDNFGetGPGSignatureCheck(
             dwError = 0;
         }
         BAIL_ON_TDNF_ERROR(dwError);
-        if(pRepo)
+        if(pRepo && pRepo->nGPGCheck)
         {
-            nGPGSigCheck = pRepo->nGPGCheck;
-            if(nGPGSigCheck)
+            nGPGSigCheck = 1;
+            if (ppszUrlGPGKey != NULL)
             {
-                if (ppszUrlGPGKey != NULL) {
-                    if (IsNullOrEmptyString(pRepo->pszUrlGPGKey))
-                    {
-                        dwError = ERROR_TDNF_NO_GPGKEY_CONF_ENTRY;
-                        BAIL_ON_TDNF_ERROR(dwError);
-                    }
-                    dwError = TDNFAllocateString(
-                                 pRepo->pszUrlGPGKey,
-                                 &pszUrlGPGKey);
+                if (IsNullOrEmptyString(pRepo->pszUrlGPGKey))
+                {
+                    dwError = ERROR_TDNF_NO_GPGKEY_CONF_ENTRY;
                     BAIL_ON_TDNF_ERROR(dwError);
                 }
+                dwError = TDNFAllocateString(
+                             pRepo->pszUrlGPGKey,
+                             &pszUrlGPGKey);
+                BAIL_ON_TDNF_ERROR(dwError);
             }
         }
     }
 
     *pnGPGSigCheck = nGPGSigCheck;
-    if (ppszUrlGPGKey) {
+    if (ppszUrlGPGKey)
+    {
         *ppszUrlGPGKey = pszUrlGPGKey;
     }
 
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -451,6 +451,9 @@ TDNFTransAddInstallPkg(
         BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
     }
 
+    dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, NULL);
+    BAIL_ON_TDNF_ERROR(dwError);
+
     fp = Fopen (pszFilePath, "r.ufdio");
     if(!fp)
     {
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -469,7 +469,11 @@ TDNFTransAddInstallPkg(
     Fclose(fp);
     fp = NULL;
 
-    if(nGPGSigCheck && (dwError == RPMRC_NOTTRUSTED || dwError == RPMRC_NOKEY))
+    if (dwError != RPMRC_NOTTRUSTED && dwError != RPMRC_NOKEY)
+    {
+        BAIL_ON_TDNF_RPM_ERROR(dwError);
+    }
+    else if(nGPGSigCheck)
     {
         dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, &pszUrlGPGKey);
         BAIL_ON_TDNF_ERROR(dwError);
@@ -478,8 +482,13 @@ TDNFTransAddInstallPkg(
         dwError = TDNFYesOrNo(pTdnf->pArgs, "Is this ok [y/N]: ", &nAnswer);
         BAIL_ON_TDNF_ERROR(dwError);
 
-        if(nAnswer)
+        if(!nAnswer)
         {
+            dwError = ERROR_TDNF_OPERATION_ABORTED;
+            BAIL_ON_TDNF_ERROR(dwError);
+	}
+	else
+	{
             pKeyring = rpmtsGetKeyring(pTS->pTS, 0);
 
             dwError = TDNFImportGPGKey(pTS->pTS, pszUrlGPGKey);
@@ -505,14 +514,8 @@ TDNFTransAddInstallPkg(
 
             Fclose(fp);
             fp = NULL;
-        } else {
-            dwError = ERROR_TDNF_OPERATION_ABORTED;
-            BAIL_ON_TDNF_ERROR(dwError);
 	}
-    } else if (!nGPGSigCheck && (dwError == RPMRC_NOTTRUSTED || dwError == RPMRC_NOKEY)) {
-        dwError = 0;
     }
-    BAIL_ON_TDNF_RPM_ERROR(dwError);
 
     dwError = TDNFGetGPGCheck(pTdnf, pszRepoName, &nGPGCheck, &pszUrlGPGKey);
     BAIL_ON_TDNF_ERROR(dwError);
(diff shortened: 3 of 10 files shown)


ppadmavilasom:
just in time use of keys without user approval is not good. the error is clear enough for user to take action. is there a specific case that requires this automatic just in time handling?

oliverkurth:
@ppadmavilasom I am not sure I follow. I can add a question to the user if they want to use the key - would that address your concern? Otherwise please elaborate.

oliverkurth:
As @keerthanakalyan noticed, dnf asks if it is okay to import the key, then adds it to the rpm keyring permanently. Thus it could be used for al packages, also from other repos, not just the one it was configured for. This PR as of now would use the key only for the configured repo. I think that makes more sense - why set the key for just one repo when it's used globally? But it maybe be better to follow dnf behavior, so I can change it. Any opinions on that?

Another question - the question would need to be added to `client/rpmtrans.c` in `TDNFTransAddInstallPkg()`. There is no precedence to have interactive questions outside of `tools/cli/`. So would this still be a good place to add it?

keerthanakalyan:
> As @keerthanakalyan noticed, dnf asks if it is okay to import the key, then adds it to the rpm keyring permanently. Thus it could be used for al packages, also from other repos, not just the one it was configured for. This PR as of now would use the key only for the configured repo. I think that makes more sense - why set the key for just one repo when it's used globally? But it maybe be better to follow dnf behavior, so I can change it. Any opinions on that?
> 
> Another question - the question would need to be added to `client/rpmtrans.c` in `TDNFTransAddInstallPkg()`. There is no precedence to have interactive questions outside of `tools/cli/`. So would this still be a good place to add it?

Hi @oliverkurth ,
IMO it's better to add the keys permanently to rpm keyring by prompting to the user instead of temporarily adding and deleting for each transaction.
Regarding the question on interactive questions on rpmtrans.c, I guess it should be okay to add interactive prompt here. Others please feel free to give your suggestions.

client/utils.c
```diff
@@ -649,3 +649,36 @@ TDNFGetCmdOpt(
...
+{
+    uint32_t dwError = 0;
+    int nAnswer = 0;
+    int32_t opt = 0;
+
+    if(!pArgs->nAssumeYes && !pArgs->nAssumeNo)
+    {
+        printf("%s ", pszQuestion);
+        while ((getchar()) != '\n');
+        opt = getchar();
```
sshedi: Do
```
opt = tolower(getchar());
```
Saves bunch of function calls over next lines

Also please call this method from:
tools/cli/lib/installcmd.c:TDNFCliAlterCommand() line 233

include/tdnferror.h
```diff
@@ -65,6 +65,8 @@ extern "C" {
 #define ERROR_TDNF_ERASE_NEEDS_INSTALL      1031
 #define ERROR_TDNF_OPERATION_ABORTED        1032
 
+#define ERROR_TDNF_INVALID_INPUT            1033
```
sshedi, line 69: please add a corresponding description in client/defines.h

client/rpmtrans.c
```diff
@@ -480,9 +465,51 @@ TDNFTransAddInstallPkg(
                   fp,
                   pszFilePath,
                   &rpmHeader);
-    //If not checking gpg sigs, ignore signature errors
-    if(!nGPGSigCheck && (dwError == RPMRC_NOTTRUSTED || dwError == RPMRC_NOKEY))
+
+    Fclose(fp);
+    fp = NULL;
+
+    if(nGPGSigCheck && (dwError == RPMRC_NOTTRUSTED || dwError == RPMRC_NOKEY))
```
sshedi: Please check for negative condition first, that saves a level of indentation and no need of else block

client/rpmtrans.c
```diff
@@ -480,9 +465,51 @@ TDNFTransAddInstallPkg(
...
     {
+        dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, &pszUrlGPGKey);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        printf("importing key from %s\n", pszUrlGPGKey);
+        dwError = TDNFYesOrNo(pTdnf->pArgs, "Is this ok [y/N]: ", &nAnswer);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        if(nAnswer)
```
sshedi: check for negative  case first, here as well :)

client/utils.c
```diff
@@ -649,3 +649,36 @@ TDNFGetCmdOpt(
...
+TDNFYesOrNo(
+    PTDNF_CMD_ARGS pArgs,
+    const char *pszQuestion,
+    int *pAnswer
+    )
+{
+    uint32_t dwError = 0;
+    int nAnswer = 0;
+    int32_t opt = 0;
+
```
sshedi: Validate function parameters first and initialize *pAnswer = 0 before proceeding

client/rpmtrans.c
```diff
@@ -480,9 +465,51 @@ TDNFTransAddInstallPkg(
...
+        {
+            pKeyring = rpmtsGetKeyring(pTS->pTS, 0);
```
sshedi: Validate pKeyring value?

client/rpmtrans.c
```diff
@@ -480,9 +465,51 @@ TDNFTransAddInstallPkg(
...
+
+            dwError = TDNFImportGPGKey(pTS->pTS, pszUrlGPGKey);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            dwError = TDNFGPGCheck(pKeyring, pszUrlGPGKey, pszFilePath);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            fp = Fopen (pszFilePath, "r.ufdio");
```
sshedi: I think we should make use of a macro for "r.ufdio"

client/rpmtrans.c
```diff
@@ -480,9 +465,51 @@ TDNFTransAddInstallPkg(
...
+            dwError = rpmReadPackageFile(
+                          pTS->pTS,
+                          fp,
+                          pszFilePath,
+                          &rpmHeader);
+
+            BAIL_ON_TDNF_RPM_ERROR(dwError);
+
+            Fclose(fp);
+            fp = NULL;
```
sshedi: is this assignment really needed? There is one more above, I feel it won't make a difference

client/repo.c
```diff
@@ -340,21 +340,25 @@ TDNFGetGPGSignatureCheck(
             nGPGSigCheck = pRepo->nGPGCheck;
             if(nGPGSigCheck)
             {
-                if (IsNullOrEmptyString(pRepo->pszUrlGPGKey))
-                {
-                    dwError = ERROR_TDNF_NO_GPGKEY_CONF_ENTRY;
+                if (ppszUrlGPGKey != NULL) {
```
sshedi: Use
```
if ()
{
   ...
}
```

client/repo.c
```diff
@@ -340,21 +340,25 @@ TDNFGetGPGSignatureCheck(
...
-                             pRepo->pszUrlGPGKey,
-                             &pszUrlGPGKey);
-                BAIL_ON_TDNF_ERROR(dwError);
             }
         }
     }
 
     *pnGPGSigCheck = nGPGSigCheck;
-    *ppszUrlGPGKey = pszUrlGPGKey;
+    if (ppszUrlGPGKey) {
```
sshedi: Use
```
if ()
{
   ...
}
```

client/utils.c
```diff
@@ -649,3 +649,36 @@ TDNFGetCmdOpt(
...
+        {
+            printf("Invalid input\n");
+            dwError = ERROR_TDNF_INVALID_INPUT;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+    }
+
+    if(pArgs->nAssumeYes || (tolower(opt) == 'y'))
+    {
+        nAnswer = 1;
```
sshedi: I think we can do
```
*pAnswer = 1;
```
and avoid using nAnswer variable here

client/repo.c
sshedi: request here, this function has like 5 indentation levels, can you please rewrite this one and not exceed 3 levels?
I kinda agree with kernel coding guide line:
```
if you need more than 3 levels of indentation, you’re screwed anyway, and should fix your program.
```

client/gpgcheck.c
```diff
@@ -376,3 +375,36 @@ VerifyRpmSig(
...
+    rpmts pTS,
+    const char* pszFile
+    )
+{
+    uint32_t dwError = 0;
+    pgpArmor nArmor = PGPARMOR_NONE;
+    uint8_t* pPkt = NULL;
+    size_t nPktLen = 0;
+    char* pszKeyData = NULL;
+
```
sshedi, line 390: validate function params

sshedi (changes requested):
Please address the comments and I'll review again

sshedi:
FYI, the travis ci tests are failing because of tdnf single instance merge.
```
#define TDNF_INSTANCE_LOCK_FILE     "/var/run/lock/.tdnf-instace-lockfile"
```
The location /var/run/lock goes missing during run time.
I have added a fix to it in https://github.com/vmware/tdnf/pull/160
If someone know the reason, please share it

keerthanakalyan:
> /var/run/lock/

Please address the fix in a separate commit instead of adding to existing PRs.

sshedi:
> > /var/run/lock/
> 
> Please address the fix in a separate commit instead of adding to existing PRs.

Will do but I'm not sure about the reason why /var/run/lock/.tdnf-instace-lockfile O_CREAT call is failing in fedora docker image

client/rpmtrans.c
oliverkurth: After closing it should be set to NULL, so it's not closed on cleanup.

sshedi: Agreed, ignore my comment about null assignment

client/repo.c
oliverkurth: I reduced it by one level. I don't think trying more would improve readability.

oliverkurth:
Thanks @sshedi for your feedback. I think I addressed all comments.

client/rpmtrans.c
```diff
@@ -480,12 +465,57 @@ TDNFTransAddInstallPkg(
...
+        printf("importing key from %s\n", pszUrlGPGKey);
+        dwError = TDNFYesOrNo(pTdnf->pArgs, "Is this ok [y/N]: ", &nAnswer);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        if(!nAnswer)
+        {
+            dwError = ERROR_TDNF_OPERATION_ABORTED;
+            BAIL_ON_TDNF_ERROR(dwError);
+	}
+	else
```
keerthanakalyan: Please correct the indentation.

client/rpmtrans.c
```diff
@@ -458,15 +450,8 @@ TDNFTransAddInstallPkg(
         BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
     }
 
-    //Check override, then repo config and launch
-    //gpg check if needed
-    dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, &pszUrlGPGKey);
+    dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, NULL);
```
keerthanakalyan, line 453: What is the significance of calling TDNFGetGPGSignatureCheck twice (one with NULL and one with URLKey)? Can we get the URLgpgkey here itself?

client/gpgcheck.c
```diff
@@ -376,3 +375,42 @@ VerifyRpmSig(
...
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    nArmor = pgpParsePkts(pszKeyData, &pPkt, &nPktLen);
+    if(nArmor != PGPARMOR_PUBKEY)
+    {
+        dwError = ERROR_TDNF_INVALID_PUBKEY_FILE;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = rpmtsImportPubkey(pTS, pPkt, nPktLen);
```
keerthanakalyan, line 407: Can't we add rpmtsImportPubkey() to AddKeyToKeyRing() instead of handling it in a separate function? Is this separately handled only because we do not have pTS as a parameter in TDNFGPGCheck() function?

client/rpmtrans.c
```diff
@@ -480,12 +465,57 @@ TDNFTransAddInstallPkg(
...
+    fp = NULL;
+
+    if (dwError != RPMRC_NOTTRUSTED && dwError != RPMRC_NOKEY)
     {
-        dwError = 0;
+        BAIL_ON_TDNF_RPM_ERROR(dwError);
+    }
+    else if(nGPGSigCheck)
+    {
+        dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, &pszUrlGPGKey);
```
keerthanakalyan, line 479: This check is done after downloading first rpm from the transaction and user is prompted and then proceeded to download next set of dependencies. Is it possible to make this check separately after downloading all the RPMs for the transaction because the right now prompt is confusing to the user as it is in the middle of download for a single RPM but it uses the same imported key for all the RPMs in the transaction.
```
root@photon-3-dev [ ~/tdnf-keerthanak-github/tdnf ]# ./build/bin/tdnf upgrade curl

Upgrading:
curl-devel                                               x86_64                       7.61.1-8.ph3                          photon-updates                   766.02k 784405
curl-libs                                                x86_64                       7.61.1-8.ph3                          photon-updates                   528.69k 541376
curl                                                     x86_64                       7.61.1-8.ph3                          photon-updates                   228.12k 233595

Total installed size:   1.49M 1559376
Is this ok [y/N] y

Downloading:
curl-devel                              703935   100%
importing key from file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
Is this ok [y/N]:  y
curl-libs                               268559   100%
curl                                    129034   100%
Testing transaction
Running transaction
Installing/Updating: curl-libs-7.61.1-8.ph3.x86_64
Installing/Updating: curl-7.61.1-8.ph3.x86_64
Installing/Updating: curl-devel-7.61.1-8.ph3.x86_64
Removing: curl-devel-7.61.1-4.ph3.x86_64
Removing: curl-7.61.1-4.ph3.x86_64
Removing: curl-libs-7.61.1-4.ph3.x86_64

Complete!
```

client/gpgcheck.c
oliverkurth: I think instead we can eliminate the `AddKeyToKeyRing()` function. It looks like rpmtsImportPubkey() already adds the key to he keyring (see https://github.com/rpm-software-management/rpm/blob/853c48ba6468ce1a516621a2fa6d1fc51e4f7410/lib/rpmts.c#L557).

How about doing this in another change?

client/rpmtrans.c
oliverkurth: The first call wouldn't bail if there is no key configured in the repo. It's not required at that point because we may already have the key in the rpm keyring, and at this point we only need to know if need to check the signature (by checking `nogpgcheck` or `skipsignature`. Only if the required key is not in the keyring do we need it configured, in the second call.

client/rpmtrans.c
oliverkurth: It is different from dnf or yum behavior, but I do not think it is confusing. Note that this isn't really caused by this change, signatures were already verified directly after download. It's only now apparent because of the added question.

We can change this, but this could be in a separate PR.

keerthanakalyan: Yes, I agree this is the original tdnf behavior. Since the previous code does not have the prompt, we wouldn't know when the check is done. We can take it in separate PR.

client/gpgcheck.c
keerthanakalyan: Yes, we can take it in a separate PR.

client/rpmtrans.c
```diff
@@ -480,12 +465,57 @@ TDNFTransAddInstallPkg(
...
+
+    Fclose(fp);
+    fp = NULL;
+
+    if (dwError != RPMRC_NOTTRUSTED && dwError != RPMRC_NOKEY)
     {
-        dwError = 0;
+        BAIL_ON_TDNF_RPM_ERROR(dwError);
+    }
+    else if(nGPGSigCheck)
```
sshedi: nit: No need of else here. just if will suffice

client/rpmtrans.c
sshedi: ?

client/rpmtrans.c
sshedi: ?

tools/cli/lib/installcmd.c
```diff
@@ -217,20 +217,12 @@ TDNFCliAlterCommand(
...
-        {
-            printf("Is this ok [y/N]: ");
-
-            opt = getchar();
-            if (tolower(opt) != 'y' && tolower(opt) != 'n')
-            {
-                printf("Invalid input\n");
-            }
-        }
+        dwError = TDNFYesOrNo(pCmdArgs, "Is this ok [y/N]", &nAnswer);
```
sshedi: Please use the exact prompt string as it was before, there may be some test case which might rely on it.

include/tdnferror.h
sshedi: is it added?

sshedi (changes requested):

client/rpmtrans.c
oliverkurth: Good point.

sshedi (approved):
Thanks for the fix. LGTM

PR #160, opened 2020-08-26
tdnf-automatic

sshedi:
This is something weird, PR shows 2 commits, don't know why


etc/systemd/tdnf-automatic-install.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=tdnf automatic install updates
+# On systems managed by either rpm-ostree/ostree, tdnf is read-only;
+# while someone might theoretically want the cache updated, in practice
+# anyone who wants that could override this via a file in /etc.
+ConditionPathExists=!/run/ostree-booted
```
keerthanakalyan, line 3: Is ConditionPathExists=!/run/ostree-booted specific to dnf behaviour with ostree? I don't see its usage in Photon, as we know that tdnf is not packaged in ostree based deployments.

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.0 FATAL_ERROR)
 
 project(tdnf VERSION 2.1.1 LANGUAGES C)
 set(PROJECT_YEAR 2019)
+set(VERSION_STRING ${PROJECT_VERSION})
```
sidcha: I see `VERSION_STRING` is being used in 2 places already but added only now. Was this broken somewhere down the line?

Also, please used `${PROJECT_VERSION}` everywhere to be consistent with the other `PROJECT_XXX` variables that cmake exports.

CMakeLists.txt
```diff
@@ -28,7 +29,10 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 # Ensure that we don't accidentally install into /usr/etc
-set(CMAKE_INSTALL_FULL_SYSCONDIR "/etc")
```
sidcha, line 31: The variables `CMAKE_INSTALL_XXX` and `CMAKE_INSTALL_FULL_XXX` are part of [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) cmake module (included above). They are is used produce interoperable install destinations please have a look.

sidcha (changes requested):
Minor build related comments.

etc/systemd/tdnf-automatic-install.service
```diff
@@ -0,0 +1,15 @@
...
+After=network-online.target
+
+[Service]
+Type=oneshot
+Nice=19
+IOSchedulingClass=2
+IOSchedulingPriority=7
+Environment="ABRT_IGNORE_PYTHON=1"
```
keerthanakalyan: What is the use of ABRT_IGNORE_PYTHON in tdnf automatic?

etc/systemd/tdnf-automatic-notifyonly.service
```diff
@@ -0,0 +1,15 @@
...
+# anyone who wants that could override this via a file in /etc.
+ConditionPathExists=!/run/ostree-booted
+After=network-online.target
+
+[Service]
+Type=oneshot
+Nice=19
+IOSchedulingClass=2
+IOSchedulingPriority=7
+Environment="ABRT_IGNORE_PYTHON=1"
```
keerthanakalyan: What is the use of ABRT_IGNORE_PYTHON in tdnf automatic?

etc/systemd/tdnf-automatic.service
```diff
@@ -0,0 +1,15 @@
...
+# anyone who wants that could override this via a file in /etc.
+ConditionPathExists=!/run/ostree-booted
+After=network-online.target
+
+[Service]
+Type=oneshot
+Nice=19
+IOSchedulingClass=2
+IOSchedulingPriority=7
+Environment="ABRT_IGNORE_PYTHON=1"
```
keerthanakalyan: What is the use of ABRT_IGNORE_PYTHON in tdnf automatic?

scripts/tdnf-automatic
```diff
@@ -0,0 +1,566 @@
...
+
+  echo -e "\ntdnf-automatic - started at $(date)...\n"
+
+  EchoDbg "[${FUNCNAME}:${LINENO}] Parsing arg(s): '$@'"
+  # Parse arguements
+  while [[ $# -gt 0 ]]; do
+    key="$1"
+
+    case ${key} in
+      -c|--auto_conf)
```
keerthanakalyan: --auto_conf option supported? I see only -c and --conf options from help.

CMakeLists.txt
sshedi: @cbsiddharth Yes, was trying few things with cmake and got this issue of versioning. 
What does your second sentence mean? is there any action item pending here?

CMakeLists.txt
sshedi: You want to keep CMAKE_INSTALL_FULL_SYSCONDIR?

etc/systemd/tdnf-automatic-install.service
sshedi: Okay, I'll remove it. Thanks

etc/systemd/tdnf-automatic-install.service
sshedi: ABRT_IGNORE_PYTHON is a python environment variable to ignore python program crashes, not needed for us, will remove it.

scripts/tdnf-automatic
sshedi: Typo, thanks for finding this one :) I used '-c' all the time while testing

CMakeLists.txt
sidcha: Yes. Instead of `set(VERSION_STRING ${PROJECT_VERSION})` you can directly use `${PROJECT_VERSION}` everywhere.

There are 2 other places which refer to `${VERSION_STRING}` in tdnf which must be replaced with `${PROJECT_VERSION}`. Since you already have an open PR and identified the issue, please fix this also.

CMakeLists.txt
sidcha: You must set `CMAKE_INSTALL_FULL_SYSCONDIR` as earlier. I'm okay with creating new variables if you need shorthands.

etc/systemd/tdnf-automatic-install.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=tdnf automatic install updates
+After=network-online.target
+
+[Service]
+Type=oneshot
```
keerthanakalyan, line 7: I just had a discussion with Ankit regarding co-existence of rpm-ostree and tdnf. It looks like, anyone can install tdnf in a rpm ostree based deployment, tdnf wont work since its read only. It is better to have "ConditionPathExists" in all the automatic services to avoid starting these services. So, please add these check to all the services.

CMakeLists.txt
sshedi: Agree. Done. Thanks

CMakeLists.txt
```diff
@@ -62,12 +66,17 @@ set(LIB_TDNF "tdnf")
 set(LIB_TDNF_SOLV "tdnfsolv")
 set(LIB_TDNF_COMMON "common")
 set(LIB_TDNF_CLI tdnfcli)
+set(SHEBANG "#!/usr/bin/env bash")
```
sidcha: Why is this needed?

sshedi: I'm using this in bin/tdnf-automatic.in and bin/tdnf-cache-updateinfo.in

I prefer using #!/usr/bin/env bash over #!/bin/bash after I faced some trouble in MacOS

SHEBANG here is a provision for developers, if they need it, they can use it.

Also, in bin/tdnf-cache-updateinfo.in, I wanted the script to depend on cmake for some variable at least instead of straight away pushing the script to build output. It also helps in case if we want to make use of few more cmake variables in script and we can recognize easily that this is dependent on cmake. Hope this clarifies question.

sidcha: Yes it does thank you. Two points/suggestions,

1. Can you not hardcode `#!/usr/bin/env bash` in script? I think it would work both places. The reason is, having a bunch on .in files is messy (starting with syntax recognition to language type categorization). If we start doing configure for SHEBANGs then we'd end up having all shell scripts with .in suffix just so it uses the SHEBANG.
1. Would it be better to put all such configuration variables in a separate file `tdnf-automatic-conf.sh.in` and then source them in your script at runtime? Again, just to keep the number congure_files low.

sidcha (approved):
I have not reviewed the bash scripts in depth, my bash skills are rather limited. Apart from that, LGTM.

.gitignore
```diff
@@ -85,7 +85,7 @@ __pycache__/
 
 # build directory
 build*/
-bin/
+bin/tdnf
```
oliverkurth, line 88: Curious - why?

oliverkurth:
This looks good to me, I haven't found any obvious issues.

keerthanakalyan (approved):
LGTM

PR #161, opened 2020-08-28
Create lock file under /var/run

sshedi:
/var/run/lock doesn't exist in fedora docker image and ci fails as a
result. So let's create it under /var/run


sshedi:
@keerthanakalyan please merge this asap

keerthanakalyan (approved):

PR #162, opened 2020-09-01
Detect RPM pretrans errors and suggest solution

keerthanakalyan:
RPM transaction errors are reported when Requires(pretrans) dependencies
are added to the same transaction. This fix detects pretrans dependency
failures and suggests solution to resolve these errors.

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -211,6 +211,103 @@ error:
     goto cleanup;
 }
 
+uint32_t
+TDNFDetectPreTransFailure(
+    rpmts pTS,
+    char *pszError
+    )
+{
+    uint32_t dwError = 0;
+    rpmtsi pi = NULL;
+    rpmte pte = NULL;
+    char *pszToken;
+    int i = 0;
+    char *pszErrorStr = NULL;
+    char *pszPkgName = NULL;
+    char *pszSymbol = NULL;
+    char *pszVersion = NULL;
+    char *pszEVR = NULL;
+    char *pszCachePkgName = NULL;
+    char *pszCachePkgEVR = NULL;
+
+    if (!pTS || IsNullOrEmptyString(pszError))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    // Error Str has the format: <Pkg-Name> <Symbol> <version-release>
+    dwError = TDNFAllocateString(pszError, &pszErrorStr);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    pszToken = strtok(pszErrorStr, " ");
+    while (pszToken != NULL)
+    {
+        switch(i)
+        {
+            case 0:
+                pszPkgName = pszToken;
+                break;
+            case 1:
+                pszSymbol = pszToken;
+                break;
+            case 2:
+                pszVersion = pszToken;
+                break;
+            default:
+                fprintf(stderr, "RPM problem string format unsupported\n");
+                dwError = ERROR_TDNF_INVALID_PARAMETER;
+                BAIL_ON_TDNF_ERROR(dwError);
+        }
+        i++;
+        pszToken = strtok(NULL, " ");
+    }
+
+    if (IsNullOrEmptyString(pszPkgName) || IsNullOrEmptyString(pszSymbol) || IsNullOrEmptyString(pszVersion))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pi = rpmtsiInit(pTS);
+    while ((pte = rpmtsiNext(pi, 0)) != NULL)
+    {
+        dwError = TDNFAllocateString(rpmteN(pte), &pszCachePkgName);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFAllocateString(rpmteEVR(pte), &pszCachePkgEVR);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        pszEVR = strtok(pszCachePkgEVR, "ph");
+        if (IsNullOrEmptyString(pszEVR))
+        {
+            dwError = ERROR_TDNF_INVALID_PARAMETER;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+
+        if (strcmp(pszCachePkgName, pszPkgName) == 0)
+        {
+            if ((strchr(pszSymbol, '>') && (rpmvercmp(pszEVR, pszVersion) > 0)) ||
+                (strchr(pszSymbol, '<') && (rpmvercmp(pszEVR, pszVersion) < 0)) ||
+                (strchr(pszSymbol, '=') && (rpmvercmp(pszEVR, pszVersion) == 0)))
+            {
+                fprintf(stderr, "Detected rpm pre-transaction dependency errors. "
+                        "Install %s %s %s first to resolve this failure.\n",
+                        pszPkgName, pszSymbol, pszVersion);
+                break;
+            }
+        }
+    }
+
+cleanup:
+    rpmtsiFree(pi);
+    TDNF_SAFE_FREE_MEMORY(pszErrorStr);
+    TDNF_SAFE_FREE_MEMORY(pszCachePkgName);
+    TDNF_SAFE_FREE_MEMORY(pszCachePkgEVR);
+    return dwError;
+error:
+    goto cleanup;
+}
+
 int
 doCheck(PTDNFRPMTS pTS)
 {
@@ -218,6 +315,8 @@ doCheck(PTDNFRPMTS pTS)
     rpmpsi psi = NULL;
     rpmProblem prob = NULL;
     nResult = rpmtsCheck(pTS->pTS);
+    char *pErrorStr = NULL;
+    uint32_t dwError = 0;
 
     rpmps ps = rpmtsProblems(pTS->pTS);
     if(ps)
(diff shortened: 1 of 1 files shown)


client/rpmtrans.c
```diff
@@ -211,13 +211,104 @@ TDNFPopulateTransaction(
...
+    rpmte pte = NULL;
+    char *pszToken;
+    int i = 0;
+    char *pszPkgName = NULL;
+    char *pszSymbol = NULL;
+    char *pszVersion = NULL;
+    char *pszEVR = NULL;
+    char *pszCachePkgName = NULL;
+    char *pszCachePkgEVR = NULL;
+
```
tapakund, line 232: better to check for pszError before we use it? If it is NULL or empty string

client/rpmtrans.c
```diff
@@ -211,13 +211,104 @@ TDNFPopulateTransaction(
...
+        pszToken = strtok(NULL, " ");
+    }
+
+    if (!pszPkgName || !pszSymbol || !pszVersion)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pi = rpmtsiInit(pTS);
```
tapakund, line 271: we haven't checked if pTS is valid as well? Also, once check if we should validate pi before passing it in rpmtsiNext.

client/rpmtrans.c
```diff
@@ -211,13 +211,104 @@ TDNFPopulateTransaction(
...
+    while ((pte = rpmtsiNext(pi, 0)) != NULL)
+    {
+        dwError = TDNFAllocateString(rpmteN(pte), &pszCachePkgName);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFAllocateString(rpmteEVR(pte), &pszCachePkgEVR);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        pszEVR = strtok(pszCachePkgEVR, "ph");
```
tapakund, line 280: check for pszEVR is NULL?

tapakund (commented):
Few comments on some error checks.

client/rpmtrans.c
keerthanakalyan: Done. Thanks.

client/rpmtrans.c
keerthanakalyan: Done. Thanks.

client/rpmtrans.c
keerthanakalyan: Done. Thanks.

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+    char *pszCachePkgName = NULL;
+    char *pszCachePkgEVR = NULL;
+
+    if (!pTS || !pszError)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    // Error Str has the format: <Pkg-Name> <Symbol> <version-release>
+    pszToken = strtok(pszError, " ");
```
sshedi: Please use strsep instead of strtok.
I think it's better if you do a strdup of pszError string before parsing it using strtok (strtok alters the input string)
So, if pszError is read only, we might end up in segfault.

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+        {
+            case 0:
+                pszPkgName = pszToken;
+                break;
+            case 1:
+                pszSymbol = pszToken;
+                break;
+            case 2:
+                pszVersion = pszToken;
+                break;
```
sshedi, line 255: Add a default case with a error log

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+        }
+        i++;
+        pszToken = strtok(NULL, " ");
+    }
+
+    if (!pszPkgName || !pszSymbol || !pszVersion)
```
sshedi: All three should be IsNullOrEmptyString.

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pi = rpmtsiInit(pTS);
```
sshedi, line 271: Is there a return value by rpmtsiInit?

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+    while ((pte = rpmtsiNext(pi, 0)) != NULL)
+    {
+        dwError = TDNFAllocateString(rpmteN(pte), &pszCachePkgName);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFAllocateString(rpmteEVR(pte), &pszCachePkgEVR);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        pszEVR = strtok(pszCachePkgEVR, "ph");
```
sshedi, line 280: Use strsep :)

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+        if (!pszEVR)
+        {
+            dwError = ERROR_TDNF_INVALID_PARAMETER;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+
+        if (strcmp(pszCachePkgName, pszPkgName) == 0)
+        {
+            if (strstr(pszSymbol, ">") && (rpmvercmp(pszEVR, pszVersion) > 0))
```
sshedi: I think strchr is better suited here for checking existence of a character occurrence

Error messages are identical so you can just logical OR the conditions in a single if

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+            {
+                fprintf(stderr, "Detected rpm pre-transaction dependency errors."
+                        "Install %s %s %s separately first to resolve this failure.\n",
+                        pszPkgName, pszSymbol, pszVersion);
+                break;
+            }
+            else if (strstr(pszSymbol, "<") && (rpmvercmp(pszEVR, pszVersion) < 0))
+            {
+                fprintf(stderr, "Detected rpm pre-transaction dependency errors."
```
sshedi: Add a space:
```
fprintf(stderr, "Detected rpm pre-transaction dependency errors."
fprintf(stderr, "Detected rpm pre-transaction dependency errors. "
```

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+            else if (strstr(pszSymbol, "=") && (rpmvercmp(pszEVR, pszVersion) == 0))
+            {
+                fprintf(stderr, "Detected rpm pre-transaction dependency errors."
+                        "Install %s %s %s separately first to resolve this failure.\n",
+                        pszPkgName, pszSymbol, pszVersion);
+                break;
+            }
+        }
+    }
+    rpmtsiFree(pi);
```
sshedi: Does this need to be freed in case of error? I mean, it should be added in cleanup:?

client/rpmtrans.c
```diff
@@ -211,13 +211,115 @@ TDNFPopulateTransaction(
...
+            dwError = ERROR_TDNF_INVALID_PARAMETER;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+
+        if (strcmp(pszCachePkgName, pszPkgName) == 0)
+        {
+            if (strstr(pszSymbol, ">") && (rpmvercmp(pszEVR, pszVersion) > 0))
+            {
+                fprintf(stderr, "Detected rpm pre-transaction dependency errors."
+                        "Install %s %s %s separately first to resolve this failure.\n",
```
sshedi: IMO, "Install %s %s %s separately to resolve this failure." will suffice.

sshedi (changes requested):

client/rpmtrans.c
keerthanakalyan: This was a requirement, to show the pre-trans errors in the output. Appliance teams use their local repos built with pre-trans dependencies. This is helpful for them to identify the issue.

client/rpmtrans.c
keerthanakalyan: yes, will do that.

client/rpmtrans.c
keerthanakalyan: yes, it returns rpmtsi (transaction element iterator) which is been validated in rpmtsiNext().

client/rpmtrans.c
keerthanakalyan: will do that. Thanks

client/rpmtrans.c
keerthanakalyan: will do that. Thanks

client/rpmtrans.c
keerthanakalyan: will do that. Thanks

client/rpmtrans.c
keerthanakalyan: Since pszError is no longer used in this function as well as in the caller, I feel it should okay if strtok modifies it. what advantage do you see using strsep instead of strtok?

sshedi: Advantage is none in our case but strsep is better than strtok while handling empty fields: https://man7.org/linux/man-pages/man3/strsep.3.html

And concern about pszError is not just modifying, this function might grow in future and we may need the original error string and if in any library updates if error string becomes read only, tdnf will crash

client/rpmtrans.c
sshedi: No, sorry for the confusion, my comment was on second line of print. The word "separately" looks odd.

keerthanakalyan: Separately is added explicitly because they should not add the package to the same command tdnf command along with the other packages. This creates a single transaction.

client/rpmtrans.c
keerthanakalyan: Done. Thanks

sidcha (approved):
LGTM.

sshedi (approved):
Thank you. LGTM

PR #166, opened 2020-09-03
prevent re-adding keys even if signature check has passed with keys f…

oliverkurth:
Turns out that the `else` in `client/rpmtrans.c` is needed after all. I had it removed as suggested in a review comment but missed that it's needed for the case that the first `rpmReadPackageFile()` succeeds (with a key that's already in the keyring). tdnf would ask for every single package for confirmation. This PR fixes that.

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -576,8 +576,7 @@ TDNFTransAddInstallPkg(
     {
         BAIL_ON_TDNF_RPM_ERROR(dwError);
     }
-
-    if(nGPGSigCheck)
+    else if(nGPGSigCheck)
     {
         dwError = TDNFGetGPGSignatureCheck(pTdnf, pszRepoName, &nGPGSigCheck, &pszUrlGPGKey);
         BAIL_ON_TDNF_ERROR(dwError);


sidcha (changes requested):
The else is redundant here as the if part bails out. 

Also, please add a space char after `if` and before `(` wherever you are modifying.

sidcha:
Okay, I see the reasoning here. But even with it, this is beyond me :-D . `IF` fails, `ELSE-IF` is evaluated. `IF` fail as if it didn't, it must bail out. So `ELSE-IF == IF`. What am I missing?

Can you also add the reasoning to the commit message itself? When looking at the code changes and git logs, this reason was not available.

tapakund (approved):
LGTM

client/rpmtrans.c
```diff
@@ -576,8 +576,7 @@ TDNFTransAddInstallPkg(
     {
         BAIL_ON_TDNF_RPM_ERROR(dwError);
```
tapakund, line 577: The point here is this BAIL_ON_TDNF_RPM_ERROR(dwError); can return back if dwError is 0. Then code need not need go in else path.

keerthanakalyan:
> Okay, I see the reasoning here. But even with it, this is beyond me :-D . `IF` fails, `ELSE-IF` is evaluated. `IF` fail as if it didn't, it must bail out. So `ELSE-IF == IF`. What am I missing?
> 
> Can you also add the reasoning to the commit message itself? When looking at the code changes and git logs, this reason was not available.

If checks for only dwError of RPMRC_NOKEY and RPMRC_NOTTRUSTED, dwError can be zero as well (in success case). In that case it should not execute else if and proceed. with if (nGPGSigCheck), tdnf checks for adding gpgkey even for a success case as well.

keerthanakalyan (approved):

client/rpmtrans.c
sidcha: Ahh, I see, thanks. But this else was an odd way of catching that cass. Would it have been cleaner to say,
```c
if (dwError) {
  if (dwError != RPMRC_NOTTRUSTED && dwError != RPMRC_NOKEY) {
    goto error;
  }
  if (nGPGSigCheck) {
     ...
  }
}
```
This construct distinguishes errors path from main code flow. But that's behind us now :-)

PR #167, opened 2020-09-07
support sequential retry of lesser priority mirrors

tapakund:
Below are the modifications done with this PR:

1. Add sha512sum support for verifying the repomd file.
2. Cache the baseurl in case of metalink.
3. support sequential retry of lesser priority mirrors
4. Removed hardcoded support of https, http, ftp, ftps, file resource types.
5. With this change, curl will decide on the protocol supported/unsupported.
6. Minor code cleanup and some improvements.


tapakund:
cc @keerthanakalyan @sshedi @cbsiddharth @oliverkurth

client/prototypes.h
```diff
@@ -25,11 +25,18 @@
 #include <openssl/md5.h>
 
 #define MAX_DIGEST_LENGTH 64
-typedef struct metalinkfile{
+
+typedef struct metalinkurls_t {
```
sidcha: the `_t` suffix convention is to indicate that a literal is a type in C. So please change as,

```
typedef struct metalinkfile {
    ...
} metalinkfile_t;
```

And use metalinkfile_t as your typename.

client/prototypes.h
```diff
@@ -25,11 +25,18 @@
...
+    char *url;
+    struct metalinkurls_t *next;
+} metalinkurls;
+
+typedef struct metalinkfile {
```
sidcha: Also  same here.

client/prototypes.h
```diff
@@ -117,12 +124,14 @@ TDNFRepoGetBaseUrl(
     const char* pszRepo,
     char** ppszBaseUrl
     );
+
 uint32_t
 TDNFRepoSetBaseUrl(
     PTDNF pTdnf,
-    const char* pszRepo,
-    char* pszBaseUrl
+    PTDNF_REPO_DATA_INTERNAL pszRepo,
```
sidcha, line 136: But if you see, TDNF uses the all CAPs notation for custom types (eg., `PTDNF_REPO_DATA_INTERNAL`). So metalinkfile_t and metalinkurs_t should actually be `TDNF_METALINK_FILES` and `TDNF_METALINK_URLS` etc.,

sidcha (changes requested):
I think this patch needs to be rebased on top of your other openssl patch. I'll review once the other one is merged.

client/prototypes.h
tapakund: Thanks, addressed.

client/prototypes.h
tapakund: Thanks, I followed same for the new structures.

client/remoterepo.c
```diff
@@ -551,12 +539,55 @@ TDNFNewMetalinkfile(
...
+    new_metalink_url->next = NULL;
+
+    if (*metalink_tail == NULL)
+    {
+        *metalink_tail = new_metalink_url;
+    }
+    else
+    {
+        (*metalink_tail)->next = new_metalink_url;
+    }
```
sidcha: This method's contract is not clear. It is called `AppendResourceURlToList` but is not doing a list walk. OOH, if you assume that the caller will always pass tail, then this IF block should not be there.

From what I'm seeing, this method could be named `TDNFAllocateResourceURl` and then in the caller should build the list with the node items it returns.

client/remoterepo.c
```diff
@@ -607,100 +639,79 @@ TDNFParseAndGetURLFromMetalink(
...
-                }
-                i++;
+                metalink_urls_head = metalink_urls_tail;
             }
-            if(resource_type[i] != NULL)
+            else
             {
-                break;
+                metalink_urls_tail = metalink_urls_tail->next;
             }
```
sidcha: Suggested:
```c
dwError = TDNFAllocateResourceURL(&current, (*resources)->url);
BAIL_ON_TDNF_ERROR(dwError);
if (!metalink_urls_head) {
    metalink_urls_head = current;
} else {
    prev->next = current;
}
prev = current;
```

common/strings.c
```diff
@@ -331,3 +331,43 @@ TDNFTrimSuffix(
...
+TDNFStringEndsWith(
+    char* pszSource,
+    const char* pszSuffix
+    )
+{
+    int nSourceStrLen = 0, nSuffixStrLen = 0;
+    uint32_t dwError = 0;
+    int ret = 0;
+
+    if (!pszSource || !pszSuffix)
```
sshedi: should this be IsNullOrEmptyString?

client/prototypes.h
```diff
@@ -24,12 +24,19 @@
 #include <openssl/sha.h>
 #include <openssl/md5.h>
 #include <openssl/evp.h>
-typedef struct metalinkfile{
+
+typedef struct _TDNF_METALINK_URLS_ {
+    char *url;
```
sshedi, line 29: url should be const?

client/repo.c
```diff
@@ -467,6 +467,155 @@ TDNFEventRepoMDDownloadEnd(
...
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    if (!pTdnf->pRepos)
+    {
+        dwError = ERROR_TDNF_NO_REPOS;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    pRepos = pTdnf->pRepos;
```
sshedi: nit: move this assignment above the previous if and modify the check accordingly.

client/repo.c
```diff
@@ -467,6 +467,155 @@ TDNFEventRepoMDDownloadEnd(
...
+
+    while (pRepos)
```
sshedi: nit: IMO, for loop looks clean while traversing SLL, in fact this whole thing can be done in one line :)

client/repo.c
```diff
@@ -467,6 +467,155 @@ TDNFEventRepoMDDownloadEnd(
...
+    {
+        if (!strcmp(pszRepo, pRepos->pszId))
+        {
+            break;
+        }
+        pRepos = pRepos->pNext;
+    }
+
```
sshedi, line 503: check pRepos for NULL after loop

client/repo.c
```diff
@@ -467,6 +467,155 @@ TDNFEventRepoMDDownloadEnd(
...
+        {
+            urls = urls->next;
+            if (urls)
+            {
+                continue;
+            }
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+        strncpy(buf, urls->url, BUFSIZ-1);
+        buf[BUFSIZ-1] = '\0'; // force terminate
```
sshedi, line 566: memset buf to 0 on every iteration, we can't be sure that buf length is BUFSIZ

client/repo.c
```diff
@@ -467,6 +467,155 @@ TDNFEventRepoMDDownloadEnd(
...
+        dwError = TDNFStoreBaseURLFromMetalink(pTdnf, pszRepo, buf);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFAllocateStringPrintf(&pszRepoMDUrl,
+                                           "%s/%s",
+                                           buf,
+                                           TDNF_REPO_METADATA_FILE_PATH);
+        BAIL_ON_TDNF_ERROR(dwError);
+        *ppszRepoMDUrl = pszRepoMDUrl;
+        break;
```
sshedi, line 579: why are we breaking here? can break right after line 565 if condition ?

client/repo.c
```diff
@@ -467,6 +467,155 @@ TDNFEventRepoMDDownloadEnd(
...
+uint32_t
+TDNFFreeMetalinkUrlsList(
+    TDNF_METALINK_URLS *metalink_urls
+    )
+{
+    uint32_t dwError = 0;
+    TDNF_METALINK_URLS *metalink_urls_temp = NULL;
+    TDNF_METALINK_URLS *metalink_urls_temp_next = NULL;
+
+    if (!metalink_urls)
```
sshedi, line 596: metalink_urls list can be empty? which is a valid case. So returning 0 should be fine

sshedi (changes requested):

client/remoterepo.c
```diff
@@ -607,100 +639,79 @@ TDNFParseAndGetURLFromMetalink(
...
         {
             fprintf(stderr, "File %s does not have any resource.\n", (*files)->name);
             dwError = ERROR_TDNF_METALINK_RESOURCE_VALIDATION_FAILED;
             BAIL_ON_TDNF_ERROR(dwError);
         }
         while(*resources)
         {
-            i = 0;
-            while(resource_type[i] != NULL)
+            dwError = TDNFAppendResourceURlToList(&metalink_urls_tail, (*resources)->url);
```
sidcha: URL in TDNFAppendResourceURlToList must be all caps.

client/repo.c
```diff
@@ -467,6 +467,155 @@ TDNFEventRepoMDDownloadEnd(
...
+    )
+{
+    uint32_t dwError = 0;
+    char *pszBaseUrlFile = NULL;
+    PTDNF_REPO_DATA_INTERNAL pRepos = NULL;
+
+    if (!pTdnf ||
+       !pTdnf->pConf ||
+       IsNullOrEmptyString(pszRepo) ||
+       IsNullOrEmptyString(pszRepoMDURL))
```
sidcha: These lines have to move forward by one space char. Align folds **after** the opening '(' in IF.

client/repo.c
sidcha: strncpy is done above. This should be fine.

client/repo.c
sidcha: Only one URL is valid. Loop has a continue above.

client/repo.c
```diff
@@ -682,22 +862,27 @@ TDNFGetRepoMD(
...
-            dwError = TDNFParseAndGetURLFromMetalink(pTdnf, pRepoData->pszId, pszMetaLinkFile, NULL);
+            dwError = TDNFRepoSetBaseUrl(pTdnf, pRepoData, pszBaseUrlFile);
             BAIL_ON_TDNF_ERROR(dwError);
         }
     }
 
     if (nNewRepoMDFile)
     {
         nReplaceRepoMD = 1;
+        /*We need to be strict here, if user have switched from metalink to baseurl or vice versa
```
sidcha: s/have/has

client/remoterepo.c
tapakund: Done, renamed the function name to TDNFAllocateResourceURL

client/remoterepo.c
tapakund: Sure, done.

client/repo.c
tapakund: Taken care of indentation :)

client/repo.c
tapakund: Sure.

client/repo.c
tapakund: Sure, modified this to for loop.

client/repo.c
tapakund: Done.

client/repo.c
tapakund: Right, better if we free only if metalink_urls is not NULL. Else just return success.

client/repo.c
tapakund: done.

client/remoterepo.c
tapakund: Yeah, this looks more simple to understand. And all at a place. Modified it. Thanks!

client/repo.c
tapakund: buf is initialized with 0. And strncpy is used. So, i think it's taken care. Also, as per the logic of this function this will not be executed twice it will break from the loop.

client/repo.c
tapakund: Yeah, we continue until TDNFDownloadFile is successful. And once its successful we store the url. Let me know if any concern here

common/strings.c
tapakund: Yeah, modified this to IsNullOrEmptyString. Thats should be fine.

client/prototypes.h
tapakund: As discussed, keeping it as it is. Don't want to complicated things.

client/remoterepo.c
```diff
@@ -540,7 +540,7 @@ TDNFGetFileHashFromMetalink(
 }
 
 uint32_t
-TDNFAppendResourceURlToList(
+TDNFAllocateResourceURL(
     TDNF_METALINK_URLS **metalink_tail,
```
sidcha: This is just `metalink_url` please remove `_tail`.

sidcha (approved):
LGTM.

client/remoterepo.c
tapakund: Yeah, missed to change this. Thanks! modified it now.

sshedi (approved):
lgtm.

client/repo.c
```diff
@@ -467,6 +467,146 @@ TDNFEventRepoMDDownloadEnd(
...
+            TDNF_SAFE_FREE_MEMORY(metalink_urls_temp->url);
+            TDNF_SAFE_FREE_MEMORY(metalink_urls_temp);
+            metalink_urls_temp = metalink_urls_temp_next;
+        }
+    }
+    return;
+}
+
 uint32_t
 TDNFGetRepoMD(
```
keerthanakalyan, line 611: This check was missed in last PR. We need to check for IsNullOrEmptyString(pRepoData->pszBaseUrl) and IsNullOrEmptyString(pRepoData->pszMetaLink) instead of (pRepoData->pszBaseUrl == NULL) && (pRepoData->pszMetaLink == NULL). Please take care of this change in next PR.

client/repo.c
```diff
@@ -545,51 +689,67 @@ TDNFGetRepoMD(
...
+            //we should remove the metalink file now
+            if(remove(pszMetaLinkFile))
             {
                 dwError = errno;
                 BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
             }
+            /*setting need download true and base url might point to another mirror
+             than the mirror metalink was pointing to */
             nNeedDownload = 1;
+            nStrictlyReplace = 1;
```
keerthanakalyan, line 738: nStrictlyReplace need not be set here. nDownload handles even if the URL used before from metalink is different from the baseURL which is set now. shasum of repomd files will be different so nDownload will download the repodata again. We can add a separate PR to improvise this logic.

client/repo.c
```diff
@@ -636,41 +795,53 @@ TDNFGetRepoMD(
...
             BAIL_ON_TDNF_ERROR(dwError);
-            dwError = TDNFDownloadFile(
-                          pTdnf,
-                          pRepoData->pszId,
-                          pszRepoMetalink,
-                          pszTmpRepoMetalinkFile,
-                          pRepoData->pszId,
-                          metalink,
-                          &ml_file);
+            dwError = TDNFAllocateStringPrintf(&pszTempBaseUrlFile,
```
keerthanakalyan, line 803: nDownload flag can be improved by checking the shasum of metalink file and download the repomd and other xml files if the shasum of metalink differs. That way we can avoid the logic of downloading repomd xml and checking the shasum. We can add a separate PR to improvise this logic.

keerthanakalyan (approved):
We can add a separate PR to improve this logic of handling metalink and repomd files shasum which will bring down the refresh time.

PR #168, opened 2020-09-07
use high level openssl api for calculating hash

tapakund:
Use EVP api for hash calculation as low level openssl api's are not accessible in FIPS mode.


tapakund:
cc @keerthanakalyan @cbsiddharth @sshedi @oliverkurth

client/defines.h
```diff
@@ -188,4 +188,5 @@ typedef enum
     {ERROR_TDNF_BASEURL_DOES_NOT_EXISTS,             "ERROR_TDNF_BASEURL_DOES_NOT_EXISTS",             "Base URL and Metalink URL not found in the repo file"},\
     {ERROR_TDNF_CHECKSUM_VALIDATION_FAILED,          "ERROR_TDNF_CHECKSUM_VALIDATION_FAILED",          "Checksum Validation failed for the file downloaded using URL from metalink"},\
     {ERROR_TDNF_METALINK_RESOURCE_VALIDATION_FAILED, "ERROR_TDNF_METALINK_RESOURCE_VALIDATION_FAILED", "No Resource present in metalink file for file download"},\
+    {ERROR_TDNF_MD5_FORBIDDEN_FIPS_MODE,             "ERROR_TDNF_MD5_FORBIDDEN_FIPS_MODE",             "API call to digest MD5 forbidden in FIPS mode!"},\
```
sidcha: Can you make this as a more common `ERROR_TDNF_FIPS_FIPS_MODE_FORBIDDEN`? FIPS may forbid more algorithm/methods for which we can reuse this error code.

sidcha (approved):
LGTM. This a much better API 👍

client/defines.h
tapakund: Agree, added a commit to address this. Thanks!

keerthanakalyan (approved):

PR #171, opened 2020-09-09
Fetch remote gpg keys from repo config

oliverkurth:
This PR adds downloading gpg keys from remote locations (http, ftp, ..). The key is first downloaded into the cache area, then used like a local key. The cached key will be deleted after it has ben added to the key ring.

When merging the PR, please squash all changes.

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -492,7 +492,7 @@ TDNFFetchRemoteGPGKey(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    dwError = TDNFFileNameFromUri(pszUrlGPGKey, &pszKeyLocation);
+    dwError = TDNFPathFromUri(pszUrlGPGKey, &pszKeyLocation);
     if (dwError == ERROR_TDNF_URL_INVALID)
     {
         dwError = ERROR_TDNF_KEYURL_INVALID;
@@ -683,7 +683,7 @@ TDNFTransAddInstallPkg(
         }
         else
         {
-            dwError = TDNFFileNameFromUri(pszUrlGPGKey, &pszLocalGPGKey);
+            dwError = TDNFPathFromUri(pszUrlGPGKey, &pszLocalGPGKey);
             if (dwError == ERROR_TDNF_URL_INVALID)
             {
                 dwError = ERROR_TDNF_KEYURL_INVALID;
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -473,6 +473,82 @@ TDNFTransAddReInstallPkgs(
 }
 
 uint32_t
+TDNFFetchRemoteGPGKey(
+    PTDNF pTdnf,
+    const char* pszRepoName,
+    const char* pszUrlGPGKey,
+    char** ppszKeyLocation
+    )
+{
+    uint32_t dwError = 0;
+    char* pszFilePath = NULL;
+    char* pszFilePathCopy = NULL;
+    char* pszDownloadCacheDir = NULL;
+    char* pszKeyLocation = NULL;
+
+    if(!pTdnf || IsNullOrEmptyString(pszRepoName || IsNullOrEmptyString(pszUrlGPGKey)))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFFileNameFromUri(pszUrlGPGKey, &pszKeyLocation);
+    if (dwError == ERROR_TDNF_URL_INVALID)
+    {
+        dwError = ERROR_TDNF_KEYURL_INVALID;
+    }
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFAllocateStringPrintf(
+                  &pszFilePath,
+                  "%s/%s/keys/%s",
+                  pTdnf->pConf->pszCacheDir,
+                  pszRepoName,
+                  pszKeyLocation);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    // dirname() may modify the contents of path, so it may be desirable to
+    // pass a copy when calling this function.
+    dwError = TDNFAllocateString(pszFilePath, &pszFilePathCopy);
+    BAIL_ON_TDNF_ERROR(dwError);
+    pszDownloadCacheDir = dirname(pszFilePathCopy);
+    if(!pszDownloadCacheDir)
+    {
+        dwError = ENOENT;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+
+    if(access(pszDownloadCacheDir, F_OK))
+    {
+        if(errno != ENOENT)
+        {
+            dwError = errno;
+        }
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+
+        dwError = TDNFUtilsMakeDirs(pszDownloadCacheDir);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFDownloadFile(pTdnf, pszRepoName, pszUrlGPGKey, pszFilePath,
+                               NULL, 0, NULL);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    *ppszKeyLocation = pszFilePath;
+
+cleanup:
+    TDNF_SAFE_FREE_MEMORY(pszFilePathCopy);
+    TDNF_SAFE_FREE_MEMORY(pszKeyLocation);
+    return dwError;
+
+error:
+    fprintf(stderr, "Error processing key: %s\n", pszUrlGPGKey);
+    TDNF_SAFE_FREE_MEMORY(pszFilePath);
+    *ppszKeyLocation = NULL;
+    goto cleanup;
+}
+
+uint32_t
 TDNFTransAddInstallPkg(
     PTDNFRPMTS pTS,
     PTDNF pTdnf,
@@ -492,9 +568,11 @@ TDNFTransAddInstallPkg(
     FD_t fp = NULL;
     char* pszDownloadCacheDir = NULL;
     char* pszUrlGPGKey = NULL;
+    char* pszLocalGPGKey = NULL;
     PTDNF_CACHED_RPM_ENTRY pRpmCache = NULL;
     rpmKeyring pKeyring = NULL;
     int nAnswer = 0;
+    int nRemote = 0;
 
     dwError = TDNFAllocateStringPrintf(
                   &pszRpmCacheDir,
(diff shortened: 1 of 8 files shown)


client/rpmtrans.c
```diff
@@ -472,6 +472,78 @@ TDNFTransAddReInstallPkgs(
...
+    const char* pszUrlGPGKey,
+    char** ppszKeyLocation)
+{
+    uint32_t dwError = 0;
+    char* pszKeyCacheDir = NULL;
+    char* pszFilePath = NULL;
+    char* pszFilePathCopy = NULL;
+    char* pszDownloadCacheDir = NULL;
+    char* pszKeyLocation = NULL;
+
```
tapakund, line 488: please NULL check for all the arguments.

client/rpmtrans.c
```diff
@@ -472,6 +472,78 @@ TDNFTransAddReInstallPkgs(
...
+    dwError = TDNFAllocateStringPrintf(
+                  &pszKeyCacheDir,
+                  "%s/%s/%s",
+                  pTdnf->pConf->pszCacheDir,
+                  pszRepoName,
+                  "keys");
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFFileNameFromUri(pszUrlGPGKey, &pszKeyLocation);
+
```
tapakund, line 501: BAIL_ON_TDNF_ERROR(dwError); missing.

client/rpmtrans.c
```diff
@@ -591,14 +665,36 @@ TDNFTransAddInstallPkg(
...
 
-        dwError = TDNFImportGPGKey(pTS->pTS, pszUrlGPGKey);
+        dwError = TDNFImportGPGKey(pTS->pTS, pszLocalGPGKey);
         BAIL_ON_TDNF_ERROR(dwError);
 
-        dwError = TDNFGPGCheck(pKeyring, pszUrlGPGKey, pszFilePath);
+        dwError = TDNFGPGCheck(pKeyring, pszLocalGPGKey, pszFilePath);
         BAIL_ON_TDNF_ERROR(dwError);
 
+        if (nRemote) {
```
tapakund: please move the { for if to next line.

client/rpmtrans.c
```diff
@@ -591,14 +665,36 @@ TDNFTransAddInstallPkg(
...
+            if(unlink(pszLocalGPGKey))
```
tapakund: space if (

common/utils.c
```diff
@@ -416,3 +416,94 @@ TDNFYesOrNo(
...
 
+uint32_t TDNFUriIsRemote(
+    const char* pszKeyUrl,
+    int *pnRemote
+)
+{
+    uint32_t dwError = 0;
+    const char *szRemotes[] = {"http://", "https://", "ftp://",
+                               "ftps://", NULL};
+    int i = 0;
```
tapakund, line 458: check the arguments before using it.

common/utils.c
```diff
@@ -416,3 +416,94 @@ TDNFYesOrNo(
...
+
+    *pnRemote = 0;
+    for(i = 0; szRemotes[i]; i++) {
+        if (strncasecmp(pszKeyUrl, szRemotes[i], strlen(szRemotes[i])) == 0) {
+            *pnRemote = 1;
+            break;
+        }
+    }
+    if (szRemotes[i] == NULL && strncasecmp(pszKeyUrl, "file://", 7) != 0) {
+       dwError = ERROR_TDNF_INVALID_PARAMETER;
```
tapakund: let me know your thoughts if we set some specific error here other than invalid parameters?

common/utils.c
```diff
@@ -416,3 +416,94 @@ TDNFYesOrNo(
...
+uint32_t TDNFFileNameFromUri(
+    const char* pszKeyUrl,
+    char** ppszFile)
+{
+    uint32_t dwError = 0;
+    const char* pszPath = NULL;
+    char *pszFile = NULL;
+    size_t nOffset;
+    const char *szPrefixes[] = {"http://", "https://", "ftp://",
+                                "ftps://", "file://", NULL};
```
tapakund, line 493: check the arguments before using it.

common/utils.c
```diff
@@ -416,3 +416,94 @@ TDNFYesOrNo(
...
+    int i = 0;
+
+    for(i = 0; szPrefixes[i]; i++) {
+        if (strncasecmp(pszKeyUrl, szPrefixes[i], strlen(szPrefixes[i])) == 0) {
+            nOffset = strlen(szPrefixes[i]);
+            break;
+        }
+    }
+    if (szPrefixes[i] == NULL) {
+       dwError = ERROR_TDNF_INVALID_PARAMETER;
```
tapakund: same here : let me know your thoughts if we set some specific error here other than invalid parameters?

common/utils.c
```diff
@@ -416,3 +416,94 @@ TDNFYesOrNo(
...
+       BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pszPath = pszKeyUrl + nOffset;
+
+    if(!pszPath || *pszPath == '\0')
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
```
tapakund: here as well: let me know your thoughts if we set some specific error here other than invalid parameters?

common/utils.c
```diff
@@ -416,3 +416,94 @@ TDNFYesOrNo(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if(*pszPath != '/')
+    {
+        //skip hostname in the uri.
+        pszPath = strchr (pszPath, '/');
+        if(pszPath == NULL)
+        {
+            dwError = ERROR_TDNF_INVALID_PARAMETER;
```
tapakund: same here as well.

common/utils.c
```diff
@@ -416,3 +416,94 @@ TDNFYesOrNo(
...
+
+    if(!pszPath || *pszPath == '\0')
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (strchr (pszPath, '#') != NULL)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
```
tapakund: here as well.

client/rpmtrans.c
```diff
@@ -472,6 +472,78 @@ TDNFTransAddReInstallPkgs(
...
+cleanup:
+    TDNF_SAFE_FREE_MEMORY(pszFilePathCopy);
+    TDNF_SAFE_FREE_MEMORY(pszKeyCacheDir);
+    TDNF_SAFE_FREE_MEMORY(pszKeyLocation);
+    return dwError;
+
+error:
+    fprintf(stderr, "Error processing key: %s\n", pszUrlGPGKey);
+    TDNF_SAFE_FREE_MEMORY(pszFilePath);
+    goto cleanup;
```
tapakund, line 548: better to set 

*ppszKeyLocation = NULL;

common/utils.c
oliverkurth: Good point. I replaced the error codes with `ERROR_TDNF_KEYURL_INVALID`.

client/rpmtrans.c
```diff
@@ -472,6 +472,86 @@ TDNFTransAddReInstallPkgs(
     return TDNFTransAddInstallPkgs(pTS, pTdnf, pInfo);
 }
 
+uint32_t
+TDNFFetchRemoteGPGKey(
+    PTDNF pTdnf,
+    const char* pszRepoName,
+    const char* pszUrlGPGKey,
+    char** ppszKeyLocation)
```
sidcha: The closing ')' goese in a new line with one level of indentation.

client/rpmtrans.c
```diff
@@ -472,6 +472,86 @@ TDNFTransAddReInstallPkgs(
...
+                  "keys");
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFFileNameFromUri(pszUrlGPGKey, &pszKeyLocation);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFAllocateStringPrintf(
+                  &pszFilePath,
+                  "%s/%s",
+                  pszKeyCacheDir,
```
sidcha: Looks like this is the only place where `pszKeyCacheDir` is being used. Can we form `pszFilePath` directly?

common/utils.c
```diff
@@ -416,3 +416,106 @@ TDNFYesOrNo(
...
+
+    pszPath = pszKeyUrl + nOffset;
+
+    if(!pszPath || *pszPath == '\0')
+    {
+        dwError = ERROR_TDNF_KEYURL_INVALID;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (strchr (pszPath, '#') != NULL)
```
sidcha, line 521: This method must be a generic `get_file_name_from_url()`. So '#' is a valid URL char (for anchors) and must be treated accordingly.

So is a '?' (URL params) and other URL defined reserved chars. These are the two common URL additional info strings that I can remember. For starters handling these 2 would be sufficient.

common/utils.c
```diff
@@ -416,3 +416,106 @@ TDNFYesOrNo(
...
+
+uint32_t TDNFFileNameFromUri(
+    const char* pszKeyUrl,
+    char** ppszFile)
+{
+    uint32_t dwError = 0;
+    const char* pszPath = NULL;
+    char *pszFile = NULL;
+    size_t nOffset;
+    const char *szPrefixes[] = {"http://", "https://", "ftp://",
```
sidcha: Call this variable protocols to match the URL terminologies. Also, your error messages must be generic, no key related errors from here.

common/utils.c
```diff
@@ -416,3 +416,106 @@ TDNFYesOrNo(
...
+
+    if(*pszPath != '/')
+    {
+        //skip hostname in the uri.
+        pszPath = strchr (pszPath, '/');
+        if(pszPath == NULL)
+        {
+            dwError = ERROR_TDNF_KEYURL_INVALID;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
```
sidcha: This has some issues. Instead of `strchr` use `strrchr` to get the last occurrence of '/' and and then see if you > 1 chars in your remaining buf space (post striping '#' and '?').

sidcha (changes requested):
URL parsing needs work.

common/utils.c
oliverkurth: I don't follow - this is to skip over the hostname, if any. For example, if we started with "http://vmware.com/pathto/key.asc", `pszPath` would point to "vmware.com/pathto/key.asc". `strchr(pszPath, '/')` skips over the "vmware.com" so we get "/pasthto/key.asc", which is what we want.

OTOH, `strrchr()` would stop at "/key.asc". We want the whole path, not just the filename (agreed that the function name is misleading then. I didn't change it, just moved from `client/gpgcheck.c`).

common/utils.c
oliverkurth: I didn't write this code, just moved it from `client/gpgcheck.c`. It's a few years old. I found out that originally the function `g_filename_from_uri()` from glib was used - which would be ideal because it would handle all sorts of corner cases and also unescape the URL, but we moved away from glib. Recent versions of `libcurl` have the function `curl_url_get()` which we could use, see https://curl.haxx.se/libcurl/c/curl_url_get.html. Unfortunately this is not in Photon 3 yet. But since we link with curl anyway, I think we should use that as soon as it's available.

I don't think we have any use case in tdnf where we would accept a `#` in a URL. It's only useful AFAIK for HTML pages, which we don't use. A `?` may be a valid use case (although odd).

client/rpmtrans.c
oliverkurth: Fixed

client/rpmtrans.c
oliverkurth: Good point. Fixed.

common/utils.c
oliverkurth: Done. Maybe could have been called `pszSchemes` too.

oliverkurth: I also addressed the error message, and added a new code. It's a bit clums though because it needs to be translated by the caller.

common/utils.c
ppadmavilasom: fyi - glib removal is a conscious decision to reduce install size.

common/utils.c
ppadmavilasom: do we have use cases to support http and ftp? @cbsiddharth please check with security on this. i remember discussions around this..

ppadmavilasom (commented):
a question i have is if we should change the current verify path in rpmtrans and create an rpmts to verify and use with `RPMTRANS_FLAG_TEST` instead of creating a new keyring (i dont mean in this change, in existing code). 
also, if you follow yum/dnf behavior of using remote keys, there is a user verification with the key checksums, name, url etc (if not in the `--assumeyes` path). have we thought about it?

oliverkurth:
> a question i have is if we should change the current verify path in rpmtrans and create an rpmts to verify and use with `RPMTRANS_FLAG_TEST` instead of creating a new keyring (i dont mean in this change, in existing code).

I had a temporary keyring in an old version of https://github.com/vmware/tdnf/pull/158 , but it was rejected.

> also, if you follow yum/dnf behavior of using remote keys, there is a user verification with the key checksums, name, url etc (if not in the `--assumeyes` path). have we thought about it?

I have seen that too. I thought about implementing it, but it's more work.

common/utils.c
sidcha: Okay.

common/utils.c
sidcha: Okay, I didn't see that this was a move. Perhaps it should stay within gpgcheck.c if it's not a generic utility function?

oliverkurth: It *is* a generic utility function. I renamed it to a more accurate name.

sidcha (approved):
LGTM.

sidcha (approved):

keerthanakalyan (approved):

common/utils.c
sidcha: @ppadmavilasom, missed this. I don't think we need FTP but we'd need HTTP and HTTPS for sure. Wherever plain HTTP is involved, we must make sure we don't fetch content that needs authentication and this decision must be made on a case-by-case manner. For instance, fetching RPMs from HTTP (or FTP) sources is okay as they are GPG checked later but fetching signature files from such locations is a security gap.

Since this method is a utility method, the callers of this method must go through security audit, IMO.

PR #173, opened 2020-09-14
Add repo signature tests

oliverkurth:
This add tests for checking the signatures of packages. It depends on https://github.com/vmware/tdnf/pull/171 .

Following changes are made:
* all packages will be signed when built
* there is no passphrase for the generated key because piping that to gpg is too complex (requires expect)
* adds 5 test functions (one negative, 4 positive)
* the rpm-sign package had to be added to the Fedora docker image

Merge should squash all individual commits

### DIFF
--- a/pytests/repo/setup-repo.sh
+++ b/pytests/repo/setup-repo.sh
@@ -28,27 +28,16 @@ mkdir -p ${BUILD_PATH}/BUILD \
 	 ${PUBLISH_PATH} \
 	 ${TEST_REPO_DIR}/yum.repos.d
 
-rpmbuild  --define "_topdir ${BUILD_PATH}" \
-	  -r ${BUILD_PATH} -ba ${REPO_SRC_DIR}/*.spec > /dev/null 2>&1
-
-cp -r ${BUILD_PATH}/RPMS ${PUBLISH_PATH}
-
-createrepo ${PUBLISH_PATH} > /dev/null 2>&1
-
-modifyrepo ${REPO_SRC_DIR}/updateinfo-1.xml ${PUBLISH_PATH}/repodata
-modifyrepo ${REPO_SRC_DIR}/updateinfo-2.xml ${PUBLISH_PATH}/repodata
-
 #gpgkey data for unattended key generation
-GPG_PASS=`openssl rand -base64 8`
 cat << EOF > ${TEST_REPO_DIR}/gpgkeydata
      %echo Generating a key for repogpgcheck signatures
+     %no-protection
      Key-Type: default
      Subkey-Type: default
      Name-Real: tdnf test
      Name-Comment: tdnf test key
      Name-Email: tdnftest@tdnf.test
      Expire-Date: 0
-     Passphrase: ${GPG_PASS}
      %commit
      %echo done
 EOF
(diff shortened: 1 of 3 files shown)


pytests/tests/test_signature.py
```diff
@@ -0,0 +1,90 @@
...
+    if url != None:
+        utils.run(['sed', '-i', '/gpgkey/s@.*@gpgkey={}@g'.format(url), tdnf_repo])
+
+# 'wrong' key in repo config, expect fail
+def test_install_nokey(utils):
+    set_gpgcheck(utils, True)
+    set_repo_key(utils, DEFAULT_KEY)
+    pkgname = utils.config["sglversion_pkgname"]
+    ret = utils.run([ 'tdnf', 'install', '-y', pkgname])
+    assert(ret['retval']  == (1514 & 0xff))
```
sidcha: This is no longer needed. You can check for full return codes.

oliverkurth: Nice! Fixed.

sidcha:
CI failure looks legit. Can you fix them?

oliverkurth:
> CI failure looks legit. Can you fix them?

It depends on https://github.com/vmware/tdnf/pull/171 . I can disable that test, but that then would need to be reverted when the PR 171 is merged.

sidcha:
> > CI failure looks legit. Can you fix them?
> 
> It depends on #171 . I can disable that test, but that then would need to be reverted when the PR 171 is merged.

Okay. Let's hold this then.

oliverkurth:
> > > CI failure looks legit. Can you fix them?

> > It depends on #171 . I can disable that test, but that then would need to be reverted when the PR 171 is merged.
> 
> Okay. Let's hold this then.

PR #171 is merged, so this passes now.

pytests/tests/test_signature.py
```diff
@@ -0,0 +1,90 @@
+#
+# Copyright (C) 2019 - 2020 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the GNU General Public License v2 (the "License");
+# you may not use this file except in compliance with the License. The terms
+# of the License are located in the COPYING file of this distribution.
+#
+#   Author: Siddharth Chandrasekaran <csiddharth@vmware.com>
```
sidcha: Please add your name or remove this line entirely.

pytests/tests/test_signature.py
```diff
@@ -0,0 +1,90 @@
...
+import tempfile
+import pytest
+
+DIST = os.environ.get('DIST')
+if DIST == 'photon':
+    DEFAULT_KEY='file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY'
+elif DIST == 'fedora':
+    DEFAULT_KEY='file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-34-primary'
+
+@pytest.fixture(scope='function', autouse=True)
```
sidcha, line 19: Why is done at a function level? Is this intended?

pytests/repo/setup-repo.sh
```diff
@@ -28,34 +28,47 @@ mkdir -p ${BUILD_PATH}/BUILD \
...
-
-createrepo ${PUBLISH_PATH} > /dev/null 2>&1
-
-modifyrepo ${REPO_SRC_DIR}/updateinfo-1.xml ${PUBLISH_PATH}/repodata
-modifyrepo ${REPO_SRC_DIR}/updateinfo-2.xml ${PUBLISH_PATH}/repodata
-
 #gpgkey data for unattended key generation
-GPG_PASS=`openssl rand -base64 8`
+#GPG_PASS=`openssl rand -base64 8`
+GPG_PASS=1234
```
sidcha: Why is this done this way? If you need it in tests later, you should pass this down from config.py to this script and use the same in tests as `utils.gpg_pass`.

sidcha (approved):
LGTM.

pytests/tests/test_signature.py
oliverkurth: Done.

pytests/repo/setup-repo.sh
oliverkurth: This was left over, now removed. GPG_PASS isn't needed, keys are now generated without a passphrase. This has no bearing on the tests. But a passphrase does not work non-interactively with signing the packages.

pytests/tests/test_signature.py
oliverkurth: Yes, this is intentional. Cleanup is needed for each individual test.

oliverkurth:
Check fails but unrelated to this PR, looks like it's caused by PR152.

pytests/tests/test_signature.py
sidcha: Okay.

sidcha:
> Check fails but unrelated to this PR, looks like it's caused by PR152.

Only fedora build is failing as gpgcheck plugin is not loaded. @sshedi can you take a look?

oliverkurth:
Test failure is unrelated.

oliverkurth:
Thanks @cbsiddharth for approving. @sshedi @tapakund @keerthanakalyan can I have one more? I have changes in the pipeline that depend on this.

keerthanakalyan (approved):

Issue #151, opened 2020-06-16
tdnf should support installing from HTTP and HTTP GPG keys


voor:
Currently `tdnf` will fail if it tries to install from a URL provided package, for example:

```
tdnf install -y http://example.com/example.rpm
```

I would expect this to work.

Additionally, GPG files in `.repo` files are required to be files, and cannot be provided as URLs.

Copied from https://github.com/vmware/photon/issues/1017
Somewhat related to https://github.com/vmware/tdnf/issues/77 except URL instead of a local file.

melck:
It will be perfect if multi gpg keys are supported as well. Ggp keys be separated with a space or comma. Ex: [Installation of `kubeadm`](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl).

oliverkurth:
Remote keys are now supported with merged PR https://github.com/vmware/tdnf/pull/171 . Multiple keys are in pending PR https://github.com/vmware/tdnf/pull/179 . Installing RPMs from command line (file or URL) is in pending PR https://github.com/vmware/tdnf/pull/175 .

oliverkurth:
This is done, all related PRs are merged.

PR #175, opened 2020-09-29
install local or remote packages from command line

oliverkurth:
This adds the functionality to install packages using a file or url on the command line, for example:
`tdnf install /path/to/package.rpm`
or
`tdnf install http://foo.bar.com/packages/package.rpm`

```
Example:
./bin/tdnf install https://packages.vmware.com/photon/3.0/photon_updates_3.0_x86_64/x86_64/less-530-2.ph3.x86_64.rpm
less-530-2.ph3.x86_64.rpm               130134   100%

Installing:
less                                      x86_64               530-2.ph3                   @cmdline             236.70k 242379

Total installed size: 236.70k 242379
Is this ok [y/N]: 
```
Internally this works by creating a temporary repository (called `@cmdline`) which contains all packages from the command line that are files or URL, then resolve for them, and install.

This PR also adds test cases to test the new functionality.

This addresses issue #151 .

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -31,7 +31,6 @@ TDNFInit(
     uint32_t dwError = 0;
     int nLocked = 0;
 
-
     pthread_mutex_lock (&gEnv.mutexInitialize);
     nLocked = 1;
     if(!gEnv.nInitialized)
@@ -712,6 +711,12 @@ TDNFOpenHandle(
     BAIL_ON_TDNF_ERROR(dwError);
 
     pTdnf->pSack = pSack;
+    pSack = NULL;
+    dwError = TDNFAddCmdLinePackages(
+                  pTdnf
+                  );
+    BAIL_ON_TDNF_ERROR(dwError);
+
     *ppTdnf = pTdnf;
 
 cleanup:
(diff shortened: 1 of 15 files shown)


oliverkurth:
Test failure is unrelated.

oliverkurth:
Added a check comparing the normalized path when downloading packages to cache.

client/init.c
```diff
@@ -205,8 +205,11 @@ TDNFRefreshSack(
         pTdnf->pArgs->nRefresh = 1;
     }
 
+    dwError = TDNFInitCmdLineRepo(pTdnf, pSack);
+    BAIL_ON_TDNF_ERROR(dwError);
+
     //If there is an empty repo directory, do nothing
-    if(pTdnf->pRepos)
+    if(pTdnf->pRepos->pNext)
```
sidcha, line 213: List with single a node will not enter this case now. Also, an empty list would segfault. Is this intended?

oliverkurth: Yes, this is intentional. The first repo will be the "@cmdline" repo, which always exists and is the first in the list. It does not need to be processed in the loop.

But when I looked over the changes, I noticed that the "@cmdline" repo is still processed in the loop, so I just changed it such that the loop starts with the second (first "regular") repo.

client/api.c
```diff
@@ -711,7 +711,16 @@ TDNFOpenHandle(
                   pTdnf->pArgs->nRefresh);
     BAIL_ON_TDNF_ERROR(dwError);
 
+    queue_init(&pTdnf->queueCmdLinePkgs);
+    queue_empty(&pTdnf->queueCmdLinePkgs);
```
sidcha: TDNFOpenHandle() looks to be a very high level function judging by the rest of the calls made here. Can you move these queue setup stuffs to where they are actually used? That way you can init only if there is a command line repo.

Also back-to-back `queue_init()` and `queue_empty()` calls look odd. Did you check the API? typically, `queue_init()` would also do what `queue_empty()` does.

client/api.c
```diff
@@ -733,6 +742,112 @@ TDNFOpenHandle(
...
+                              "%s/%s",
+                              cwd,
+                              pszPkgName);
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+            else
+            {
+                dwError = TDNFAllocateString(pszPkgName, &pszRPMPath);
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
```
sidcha: Use `realpath()`

client/api.c
```diff
@@ -733,6 +742,112 @@ TDNFOpenHandle(
...
+        }
+        else
+        {
+             dwError = TDNFUriIsRemote(pszPkgName, &nIsRemote);
```
sidcha, line 786: This pattern looks **really** odd. I know we've been over this, I still believe you shouldn't _always_ return `unint32_t`.

Looks much better written as:
```
if (TDNFUriIsRemote()) {
    ....
}
```

client/api.c
```diff
@@ -889,7 +1004,14 @@ TDNFResolve(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    queue_init(&queueGoal);
+    if(nAlterType == ALTER_INSTALL)
```
sidcha, line 994: Also check if `pTdnf->queueCmdLinePkgs` is not NULL.  See comment earlier.

client/remoterepo.c
```diff
@@ -929,3 +935,123 @@ TDNFDownloadPackage(
...
+    dwError = TDNFPathFromUri(pszPackageLocation, &pszRemotePath);
+    if (dwError == ERROR_TDNF_URL_INVALID)
+    {
+        dwError = TDNFAllocateString(pszPackageLocation, &pszRemotePath);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    else
+    {
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
```
sidcha: Suggested:
```    
    dwError = TDNFPathFromUri(pszPackageLocation, &pszRemotePath);
    if (dwError == ERROR_TDNF_URL_INVALID)
    {
        dwError = TDNFAllocateString(pszPackageLocation, &pszRemotePath);   
    }
    BAIL_ON_TDNF_ERROR(dwError);
```

client/resolve.c
```diff
@@ -167,6 +170,27 @@ TDNFPrepareAllPackages(
            }
            else
            {
+               dwError = TDNFIsFileOrSymlink(pszPkgName, &nIsFile);
```
sidcha, line 173: Again, odd pattern. Methods with GET, SET, and IS keywords are standardized and using it this way hampers code reading.

client/utils.c
```diff
@@ -219,6 +219,51 @@ TDNFUtilsMakeDirs(
     goto cleanup;
 }
 
+uint32_t
+TDNFIsFileOrSymlink(
```
sidcha, line 223: Is this function (and it's check) really needed? You can always use `realpath()` to convert symlinks to absolute paths very early and then work only with absolute paths and URLs.

sidcha (changes requested):

client/api.c
oliverkurth: In this case the pattern is actually useful, because we have 3 (interesting) states returning from `TDNFUriIsRremote()`: remote URL, local URL (file://), or not a URL at all; and not just a boolean.

But besides that, I would agree with you if this pattern wasn't so strongly used already throughout all tdnf code. With new code I would agree that this is odd, but I think breaking this pattern in tdnf would make it more confusing.

client/utils.c
oliverkurth: Point of this is to find out if the cmd line argument is a file (or a symlink), in contrast to a simple package or a URL. The real path is not interesting at this moment. Also, if the symlink doesn't resolve we shouldn't fail at this point, because if the link exists the user's intention is most likely to use this as the package path to be installed. They will get the error later, but if we interpret this as a plain package from a repo this will be more confusing.

client/api.c
oliverkurth: `pTdnf->queueCmdLinePkgs` is not a pointer.

client/api.c
oliverkurth: I could swear I saw the pattern of `queue_empty()`  following `queue_init()` elsewhere so I copied it, but I must have imagined things...

sidcha (approved):
You have a local file checked-in; apart from that, LGTM.

oliverkurth:
Thanks! Removed garbage file, squashed changes and pushed. I can merge tomorrow unless someone objects.

PR #176, opened 2020-09-30
Add SSL options support

mrgirgin:
- Parse SSL related fields from .repo file: `sslverify, sslcacert, sslclientcert, sslclientkey`
- Pass these options to curl.
- Define a new error value for errors occurring in this phase.

### DIFF
--- a/client/repoutils.c
+++ b/client/repoutils.c
@@ -680,3 +680,101 @@ cleanup:
 error:
     goto cleanup;
 }
+
+uint32_t
+TDNFRepoApplySSLSettings(
+    PTDNF pTdnf,
+    const char* pszRepo,
+    CURL *pCurl
+    )
+{
+    uint32_t dwError = 0;
+    PTDNF_REPO_DATA_INTERNAL pRepos = NULL;
+
+    if(!pTdnf || IsNullOrEmptyString(pszRepo) || !pCurl)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    if(!pTdnf->pRepos)
+    {
+        dwError = ERROR_TDNF_NO_REPOS;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    pRepos = pTdnf->pRepos;
+
+    while(pRepos)
+    {
+        if(!strcmp(pszRepo, pRepos->pszId))
+        {
+            break;
+        }
+        pRepos = pRepos->pNext;
+    }
+
+    if(!pRepos)
+    {
+        dwError = ERROR_TDNF_REPO_NOT_FOUND;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if(curl_easy_setopt(
+            pCurl,
+            CURLOPT_SSL_VERIFYPEER,
+            ((pRepos->nSSLVerify) ? 1 : 0)) != CURLE_OK)
+    {
+        dwError = ERROR_TDNF_SET_SSL_SETTINGS;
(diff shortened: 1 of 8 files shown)


client/repoutils.c
```diff
@@ -680,3 +680,101 @@ TDNFRepoApplyProxySettings(
...
+    }
+
+    if(curl_easy_setopt(
+            pCurl,
+            CURLOPT_SSL_VERIFYPEER,
+            ((pRepos->nSSLVerify) ? 1 : 0)) != CURLE_OK)
+    {
+        dwError = ERROR_TDNF_SET_SSL_SETTINGS;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
```
oliverkurth, line 728: Can we get a more detailed error here from curl? Same for the other 3 `curl_easy_setopt()` calls.

tapakund: Pls refer this for setting curl error

https://github.com/vmware/tdnf/blob/31a577946b6286daf10c7af5f1b94f36a8b2e59f/client/remoterepo.c#L793

mrgirgin: I tried to replicated the behavior from `TDNFRepoApplyProxySettings` earlier in this file. I frankly feel like pointing out which user input caused the error is more useful than the `curl`'s internal error. In most of the cases I have encountered, `curl` already prints to stderr about what went wrong anyway.

oliverkurth: So if curl prints out to stderr before returning this might be good enough - we just need information what went wrong. Can you give an example?

sidcha: Here is a sample:

```
curl#60: Peer certificate cannot be authenticated with given CA certificates
```

sidcha (approved):
LGTM, thanks for the PR!

tapakund (commented):
LGTM

oliverkurth:
LGTM, it's straight forward. I verified locally that the tests now pass with PR178 merged in. @mrgirgin , before we can merge you need to sign the CLA - can you do that?

Also, it would be great to have some tests for this.

sidcha:
@mrgirgin, CLA bot seems to have some issue we are working with the respective teams to get this sorted out. No action is needed from your side for now.

client/structs.h
```diff
@@ -36,11 +36,16 @@ typedef struct _TDNF_REPO_DATA_INTERNAL_
     int nEnabled;
     int nSkipIfUnavailable;
     int nGPGCheck;
+    int nSSLVerify;
     long lMetadataExpire;
     char* pszId;
     char* pszName;
     char* pszBaseUrl;
     char* pszMetaLink;
+    char* pszUrlGPGKey;
```
oliverkurth: This can go away - it's replaced by ppszUrlGPGKeys below.

include/tdnftypes.h
```diff
@@ -268,11 +268,16 @@ typedef struct _TDNF_REPO_DATA
     int nEnabled;
     int nSkipIfUnavailable;
     int nGPGCheck;
+    int nSSLVerify;
     long lMetadataExpire;
     char* pszId;
     char* pszName;
     char* pszBaseUrl;
     char* pszMetaLink;
+    char* pszUrlGPGKey;
```
oliverkurth: Same here.

oliverkurth (changes requested):
Also, when done, please squash the changes into one. You can force push your changes to the PR.

mrgirgin:
> Also, when done, please squash the changes into one. You can force push your changes to the PR.

You should be able to "Squash & Merge" in the Github UI.

oliverkurth (approved):

oliverkurth:
Thank you @mrgirgin ! It's merged.

PR #178, opened 2020-10-01
Fixed TDNFDownloadFile call in repogpgcheck plugin

sshedi:
repogpgcheck plugin calls TDNFDownloadFile but it's signtaure was wrong.
As a result this was ending up in unacceptable results.


tapakund (approved):
LGTM

oliverkurth:
"prototype was wrong" would be a better description (not "signature"), but looks good. I am surprised this didn't crash before.

oliverkurth (approved):

PR #180, opened 2020-10-06
Minor pytest enhancements

sidcha:
- Add feature to test self signed HTTPS repos at some point in future
- Default to DIST==photon in local builds
- Print the exact command that caused the test failure for convenience


oliverkurth (approved):
LGTM, nice!

keerthanakalyan (approved):

PR #182, opened 2020-10-13
normalize paths for gpg key URLs to prevent directory traversals

oliverkurth:
This adds the function `TDNFNormalizePath()` to prevent directory traversals using the path to the key. The key is downloaded to cache using the same path as in the URL (analogous to the location of cached rpm packages). Potentially, this could result in a location outside of the cache in some arbitrary location in the filesystem, unless the path is checked. The check works by 'normalizing' the constructed cache path by parsing it and evaluating "up paths" like `../`. Double slashes and "same directory" nodes (`./`) are also eliminated. The result is checked to make sure it is within the cache directory.

The function `realpath()` from libc could not be used because the path may not exist yet.

Tests are also added.

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -447,7 +447,8 @@ error:
     return dwError;
 }
 
-uint32_t TDNFUriIsRemote(
+uint32_t
+TDNFUriIsRemote(
     const char* pszKeyUrl,
     int *pnRemote
 )
@@ -495,8 +496,8 @@ uint32_t TDNFPathFromUri(
 
     if(IsNullOrEmptyString(pszKeyUrl) || !ppszPath)
     {
-      dwError = ERROR_TDNF_INVALID_PARAMETER;
-      BAIL_ON_TDNF_ERROR(dwError);
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
     }
 
     for(i = 0; szProtocols[i]; i++) {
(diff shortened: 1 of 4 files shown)


common/utils.c
```diff
@@ -550,3 +551,87 @@ uint32_t TDNFPathFromUri(
...
+
+    dwError = TDNFAllocateMemory(1, strlen(pszPath) + 1, (void **)&pszNormalPath);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    q = pszNormalPath;
+
+    while(*p)
+    {
+        /* double slashes */
+        if (*p == '/' && (p > pszPath && p[-1] == '/'))
```
sidcha: Can we avoid look-behind (the -ve char offsets)? I'm not too sure about this, but from the problem statement (in PR description), I have a feeling this can be done with just look-aheads. This improves reliability a lot as you can get rid of one check before looking back.

common/utils.c
```diff
@@ -550,3 +551,87 @@ uint32_t TDNFPathFromUri(
...
+        {
+            p++;
+        }
+        /* single dots */
+        else if (*p == '.' &&
```
sidcha: When using p[n] form, use p[0] instead of *p for readability/consistency.

common/utils.c
sidcha: Also, are you handling multiple '/' chars? 

This if condition should be `while (p[0] == '/' && p[1] == '/') p++`

sidcha (changes requested):

common/utils.c
oliverkurth: This handles multiple slashes already. Only the first slash is copied (condition is false), every subsequent one is ignored (condition true).

common/utils.c
oliverkurth: Do you insist? In my eyes this looks better because it makes it clearer where the pointer is.

common/utils.c
oliverkurth: The look back is needed because we need to ignore a slash if the previous one is a slash too, and similar for the other conditions. I think all look backs are sufficiently protected.

Thinking more about it, another way would be to set a flag that is `true` initially, then reset if `*p` did not point to a slash. Not sure if that makes it more readable though.

oliverkurth:
After discussing with Muhammad, fixed two issues:
- use `realpath()` for every existing node in the path to expand symlinks
- normalize the top cache dir too before comparing with the normalized key path

common/utils.c
sidcha: No. That was just a stylestic remark.

common/utils.c
sidcha: > The look back is needed because we need to ignore a slash if the previous one is a slash too, and similar for the other conditions. I think all look backs are sufficiently protected.

With the suggested code (`while (p[0] == '/' && p[1] == '/') p++`), the idea is to keep the cursor (p) at the last char of interest (this would be a loop invariant). So in case of '///////' p would point to the last '/'. With that assumption, rest of the code can just look ahead. This works well with '..' handling too (`p[1] == '.' && p[2] == '.'`). This way all movement of the cursor will be at the top; when you are done with a section of text, a `continue` should get you back in the loop meeting all the loop invariant again.

common/utils.c
```diff
@@ -550,3 +551,142 @@ uint32_t TDNFPathFromUri(
...
+            }
+            /* it's okay if path doesn't exist, bail on other errors */
+            else if (errno != ENOENT)
+            {
+                dwError = ERROR_TDNF_SYSTEM_BASE + errno;
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+            /* skip over last slash in path */
+	    if (p[1] == 0)
+            {
```
sidcha, line 648: Please fix white space issue here.

common/utils.c
```diff
@@ -550,3 +551,142 @@ uint32_t TDNFPathFromUri(
...
+
+    q = pszNormalPath;
+
+    while(*p)
+    {
+        /* double slashes */
+        if (*p == '/' && p[1] == '/')
+        {
+            p++;
+	    continue;
```
sidcha: Please fix white space issue here.

common/utils.c
```diff
@@ -550,3 +551,142 @@ uint32_t TDNFPathFromUri(
...
+        *q++ = *p++;
+    }
+    *q = 0;
+
+    /* an empty path should evaluate to a slash
+     * like realpath() does */
+    if (pszNormalPath[0] == 0)
+    {
+        TDNF_SAFE_FREE_MEMORY(pszNormalPath);
+	pszNormalPath = strdup("/");
```
sidcha: Please fix white space issue here.

sidcha (approved):
Nice, LGTM!

common/utils.c
oliverkurth: Thanks Siddharth! I reworked the code, and it indeed looks much cleaner this way. There is no looking back any more.

common/utils.c
oliverkurth: Thx, fixed

keerthanakalyan (approved):
LGTM

PR #185, opened 2020-10-23
Update release version to 3.0.0-beta

keerthanakalyan:


oliverkurth:
Looks good. I wonder if we could somehow add the git sha in the version. This way there would be no doubt from which source the binary was built rom.

oliverkurth (approved):

keerthanakalyan:
> Looks good. I wonder if we could somehow add the git sha in the version. This way there would be no doubt from which source the binary was built rom.

Yes, definitely. We can add the shasum from next release onwards.

PR #189, opened 2020-10-27
fix resource leaks

oliverkurth:
Fix a few resource leaks uncovered by coverity.


client/repolist.c
```diff
@@ -564,6 +564,7 @@ TDNFCloneRepo(
                       pRepoIn->ppszUrlGPGKeys,
                       &pRepo->ppszUrlGPGKeys);
     }
+    BAIL_ON_TDNF_ERROR(dwError);
```
tapakund: better to keep this code within the if

tapakund, line 567: better to keep this within the if condition.

keerthanakalyan (approved):

PR #191, opened 2020-10-28
Fix pytest in fedora

keerthanakalyan:
Fedora latest has python 3.9.0 and latest pytest which
does not have '_pytest.deprecated'. Moving to fedora 32 till we fix this issue.


tapakund (approved):
LGTM

sshedi (approved):

PR #195, opened 2020-11-12
Topic/okurth/exclude option

oliverkurth:
Use exclude option also for other altering commands, not just update. Add tests.


keerthanakalyan (approved):
LGTM
Please validate whether --exclude with --security and --sec-severity works fine with subpackages.
For example: tdnf --exclude=curl update --security
should exclude all subpackages of curl and its dependencies from the install/update pkgs list.

PR #197, opened 2020-11-20
read multiple keys from one file

oliverkurth:
A single key file can have multiple keys, but tdnf read only one. This change makes it read all public keys from the file. This also fixes a potential bug where tdnf wouldn't accept a key file when the first packet in it was not a public key but other pgp data.


keerthanakalyan (approved):
LGTM

PR #199, opened 2020-11-27
print improvements

sshedi:
Added a utilty function to print only if quiet not set
Also print errors irrespective of quiet option

If a non-exisitng package is provided for install/update let's show the
name of the invalid package in error message

### DIFF
--- a/client/client.c
+++ b/client/client.c
@@ -80,17 +80,11 @@ TDNFPkgsToExclude(
 
     if (!pTdnf->pArgs->nDisableExcludes && pTdnf->pConf->ppszExcludes)
     {
-        if (!pTdnf->pArgs->nQuiet)
-        {
-            printf("Warning: The following packages are excluded "
-                   "from tdnf.conf:\n");
-        }
+        pr_info("Warning: The following packages are excluded "
+                "from tdnf.conf:\n");
         while (pTdnf->pConf->ppszExcludes[nIndex])
         {
-            if (!pTdnf->pArgs->nQuiet)
-            {
-                printf("  %s\n", pTdnf->pConf->ppszExcludes[nIndex]);
-            }
+            pr_info("  %s\n", pTdnf->pConf->ppszExcludes[nIndex]);
             dwCount++;
             nIndex++;
         }
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -52,9 +52,9 @@ progress_cb(
 
     dPercent = ((double)dlNow / (double)dlTotal) * 100.0;
     if (!isatty(STDOUT_FILENO)) {
-        printf("%s %3.0f%% %ld\n", (char *)pUserData, dPercent, dlNow);
+        pr_info("%s %3.0f%% %ld\n", (char *)pUserData, dPercent, dlNow);
     } else {
-        printf("%-35s %10ld %5.0f%%\r", (char *)pUserData, dlNow, dPercent);
+        pr_info("%-35s %10ld %5.0f%%\r", (char *)pUserData, dlNow, dPercent);
     }
 
     fflush(stdout);
@@ -143,7 +143,7 @@ TDNFGetDigestForFile(
     fd = open(filename, O_RDONLY);
     if (fd == -1)
     {
-        fprintf(stderr, "Metalink: validating (%s) FAILED\n", filename);
+        pr_err("Metalink: validating (%s) FAILED\n", filename);
         dwError = errno;
         BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
     }
@@ -152,7 +152,7 @@ TDNFGetDigestForFile(
 
     if (!digest_type)
     {
-        fprintf(stderr, "Unknown message digest %s\n", hash->hash_type);
+        pr_err("Unknown message digest %s\n", hash->hash_type);
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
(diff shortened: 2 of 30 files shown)


client/client.c
```diff
@@ -82,14 +82,14 @@ TDNFPkgsToExclude(
     {
         if (!pTdnf->pArgs->nQuiet)
```
oliverkurth: This `if` shouldn't be needed any more.

client/client.c
```diff
@@ -82,14 +82,14 @@ TDNFPkgsToExclude(
...
         {
-            printf("Warning: The following packages are excluded "
+            pr_stdout("Warning: The following packages are excluded "
                    "from tdnf.conf:\n");
         }
         while (pTdnf->pConf->ppszExcludes[nIndex])
         {
             if (!pTdnf->pArgs->nQuiet)
```
oliverkurth: Same for this `if`.

client/repo.c
```diff
@@ -813,7 +811,7 @@ TDNFGetRepoMD(
     {
         if(!pTdnf->pArgs->nQuiet)
```
oliverkurth: Another unneeded `if`.

client/remoterepo.c
```diff
@@ -924,7 +924,7 @@ TDNFDownloadPackage(
 
     if(!nSilent)
```
oliverkurth: Unneeded now. We can also remove getting the value for `nSilent` and eliminate the variable in this function.

client/rpmtrans.c
```diff
@@ -409,7 +409,7 @@ TDNFRunTransaction(
     //TODO do callbacks for output
     if(!nSilent)
```
oliverkurth: Unneeded `if`.

tools/cli/lib/api.c
```diff
@@ -251,19 +251,19 @@ TDNFCliInfoCommand(
...
+        pr_stdout("Arch          : %s\n", pPkg->pszArch);
+        pr_stdout("Epoch         : %d\n", pPkg->dwEpoch);
+        pr_stdout("Version       : %s\n", pPkg->pszVersion);
+        pr_stdout("Release       : %s\n", pPkg->pszRelease);
+        pr_stdout("Install Size  : %s (%u)\n", pPkg->pszFormattedSize, pPkg->dwInstallSizeBytes);
+        pr_stdout("Repo          : %s\n", pPkg->pszRepoName);
+        pr_stdout("Summary       : %s\n", pPkg->pszSummary);
+        pr_stdout("URL           : %s\n", pPkg->pszURL);
+        pr_stdout("License       : %s\n", pPkg->pszLicense);
+        pr_stdout("Description   : %s\n", pPkg->pszDescription);
```
oliverkurth: Does it make sense to suppress the output here? The point of the whole command is to print this out.

tools/cli/lib/api.c
```diff
@@ -273,7 +273,7 @@ TDNFCliInfoCommand(
 
     if(dwCount > 0)
     {
-        printf("\nTotal Size: %s (%lu)\n", pszFormattedSize, dwTotalSize);
+        pr_stdout("\nTotal Size: %s (%lu)\n", pszFormattedSize, dwTotalSize);
```
oliverkurth: Also here - we want this info.

tools/cli/lib/help.c
```diff
@@ -35,60 +35,60 @@ TDNFCliShowHelp(
...
-    printf("           [--refresh]\n");
-    printf("           [--exclude [file1,file2,...]]\n");
-    printf("           [--security]\n");
-    printf("           [--sec-severity CVSS_v3.0_Severity]\n");
-    printf("           [--reboot-required]\n");
-    printf("           [--skipsignature]\n");
-    printf("           [--skipdigest]\n");
-    printf("           [--disableexcludes]\n");
+    pr_stdout("options    [-c [config file]]\n");
+    pr_stdout("           [--debugsolver]\n");
```
oliverkurth: This is also information that should be printed always.

oliverkurth (commented):
Generally, I wonder if instead of defining the output stream in the function names(`pr_stdout` and `pr_stderr`), it would be better to have `log()` functions (`log_info()`, `log_err()`, `log_debug()`, etc...) for different levels of verbosity. That way we could have a "debug" level with more verbosity, and we could later make simple changes if we want to output to another destination (a log file, or syslog, etc...).

tools/cli/lib/api.c
sshedi: The idea here is, if '-q' argument is given, don't print anything except errors. If you all think otherwise, I will just revert the changes in this function.

sshedi:
> Generally, I wonder if instead of defining the output stream in the function names(`pr_stdout` and `pr_stderr`), it would be better to have `log()` functions (`log_info()`, `log_err()`, `log_debug()`, etc...) for different levels of verbosity. That way we could have a "debug" level with more verbosity, and we could later make simple changes if we want to output to another destination (a log file, or syslog, etc...).

I understand your concern here, it makes sense. But in tdnf we are using console logging only and lot of our test cases depend on these prints. So, I tried to keep it simple with minimal changes. I'll try to make it better and push my changes. Thanks for the input.

tools/cli/lib/api.c
oliverkurth: FWIW, dnf does show the help output with `-q`. Similar for the `info` command.

oliverkurth:
> I understand your concern here, it makes sense. But in tdnf we are using console logging only and lot of our test cases depend on these prints. So, I tried to keep it simple with minimal changes. I'll try to make it better and push my changes. Thanks for the input.

This is no doubt an improvement. We can add debugging or more complex logging always later.

oliverkurth:
The tests are failing - any idea what's going on there?

tools/cli/lib/api.c
sshedi: Thank you, I have made the necessary changes. Please let me know if something needs to be printed irrespective of '-q' option.

oliverkurth (approved):
LGTM. I pulled the PR locally and tested a few things, looks good.

tapakund (approved):
LGTM

keerthanakalyan:
@sshedi : Can you please resolve the conflicts?

PR #200, opened 2020-12-01
add --downloadonly and --downloaddir options

oliverkurth:
Add `--downloadonly` and `--downloaddir` options and tests.


client/rpmtrans.c
```diff
@@ -90,7 +94,7 @@ TDNFRpmExecTransaction(
     }
     if(ts.pCachedRpmsArray)
     {
-        if(!nKeepCachedRpms)
+        if(!nKeepCachedRpms && !nDownloadOnly)
```
tapakund, line 97: In makecache code path, we are cleaning up tdnf cached rpms only considering "keepcache" from the tdnf.conf. So, with this feature in place, we would need to consider "pTdnf->pArgs->nDownloadOnly" as well before cleaning up the cache, right?

tapakund (commented):
Have another query as well.

Suppose user does ```tdnf install -y --downloadonly --pkgname``` and does a ctrl+c in between. The rpm might be partially downloaded, so in case we use that to install we will face an error. This might be a general case in case ```keepcache=true``` as well.

We can discuss on this.

oliverkurth:
> Have another query as well.
> 
> Suppose user does `tdnf install -y --downloadonly --pkgname` and does a ctrl+c in between. The rpm might be partially downloaded, so in case we use that to install we will face an error. This might be a general case in case `keepcache=true` as well.

With change 0a513623670b88ba709e4173a25108219ca9d70f that shouldn't happen.

client/rpmtrans.c
oliverkurth: I am confused. How would `makecache` download rpms? Why would you give that `--downloadonly` option to the`makecache` command? Can you elaborate, I may be missing something.

client/remoterepo.c
```diff
@@ -1082,3 +1082,48 @@ TDNFDownloadPackageToCache(
...
+    char* pszFileName;
+
+    dwError = TDNFPathFromUri(pszPackageLocation, &pszRemotePath);
+    if (dwError == ERROR_TDNF_URL_INVALID)
+    {
+        dwError = TDNFAllocateString(pszPackageLocation, &pszRemotePath);
+    }
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    pszFileName = basename(pszRemotePath);
```
sshedi, line 1120: validate return value

client/remoterepo.c
```diff
@@ -1082,3 +1082,48 @@ TDNFDownloadPackageToCache(
...
+
+    dwError = TDNFAllocateStringPrintf(
+                  &pszFilePath,
+                  "%s/%s",
+                  pszDirectory,
+                  pszFileName);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFDownloadPackage(pTdnf, pszPackageLocation, pszPkgName,
+        pszRepoName, pszDirectory);
```
sshedi: nit: put this in same line may be?

client/remoterepo.c
```diff
@@ -1082,3 +1082,48 @@ TDNFDownloadPackageToCache(
...
+    const char* pszRepoName,
+    const char* pszDirectory,
+    char** ppszFilePath
+    )
+{
+    uint32_t dwError = 0;
+    char* pszFilePath = NULL;
+    char* pszRemotePath = NULL;
+    char* pszFileName;
+
```
sshedi, line 1112: validate function arguments

pytests/tests/test_downloadonly.py
```diff
@@ -0,0 +1,55 @@
...
+def test_install_download_only_to_directory(utils):
+    pkgname = utils.config["sglversion_pkgname"]
+    os.makedirs(DOWNLOADDIR)
+    ret = utils.run([ 'tdnf', 'install', '-y', '--downloadonly', '--downloaddir', DOWNLOADDIR, pkgname])
+    print (ret)
+    assert(ret['retval']  == 0)
+    assert(not utils.check_package(pkgname))
+    assert(len(glob.glob('{}/{}*.rpm'.format(DOWNLOADDIR, pkgname))) > 0)
+
+# --downloaddir option without --downloadonly should fail
```
sshedi, line 49: Why --downloaddir needs --downloadonly? Why can't we just use --downloaddir and set --downloadonly by default? Feels like too much to type while providing cmd line argument :)

sshedi (commented):
I feel there is room for improvement here.

If user uses --downloadonly, we need to download to a separate directory in /var/cache/tdnf and not use default download location. reason being, suppose if I download a rpm and do a `tdnf install <som-pkg>`, then downloaded rpm will get removed along with newly installed rpm.

Also, if there is a rpm in download directory already; and later if we do a tdnf install of same package, whether same rpm will be used for installation (without downloading again)?

I'm not sure about this, can we download only latest rpm or older version of rpms as well? In case if we can download older version rpms, then how to handle install scenario of same package?

Just my thoughts on this. If you feel task is being morphed to something else we can take it in a new PR I guess.

oliverkurth:
> I feel there is room for improvement here.
> 
> If user uses --downloadonly, we need to download to a separate directory in /var/cache/tdnf and not use default download location. reason being, suppose if I download a rpm and do a `tdnf install <som-pkg>`, then downloaded rpm will get removed along with newly installed rpm.

No, see below:

> Also, if there is a rpm in download directory already; and later if we do a tdnf install of same package, whether same rpm will be used for installation (without downloading again)?

I tested this - if we first do `-downloadonly`, then install, the package won't be downloaded again:
```
root [ /build/build ]# ./bin/tdnf --downloadonly -y install less

Installing:
less                                        x86_64                530-2.ph3                     photon-updates        236.70k 242379

Total installed size: 236.70k 242379

Downloading:
less                                    130134   100%

Complete!
root [ /build/build ]# ./bin/tdnf -y install less

Installing:
less                                        x86_64                530-2.ph3                     photon-updates        236.70k 242379

Total installed size: 236.70k 242379

Downloading:
Testing transaction
Running transaction
Installing/Updating: less-530-2.ph3.x86_64

Complete!
```
> I'm not sure about this, can we download only latest rpm or older version of rpms as well? In case if we can download older version rpms, then how to handle install scenario of same package?

You can do that, but why would this be any different from a normal install of an older package:
```
root [ /build/build ]# ./bin/tdnf --downloadonly -y install less=530-1.ph3

Downgrading:
less                                        x86_64                530-1.ph3                     photon                229.65k 235163

Total installed size: 229.65k 235163

Downloading:
less                                    121878   100%

Complete!
```

If I want to install the package to the latest version, it will download that if even if there is an older version since the filenames won't match.

> Just my thoughts on this. If you feel task is being morphed to something else we can take it in a new PR I guess.

pytests/tests/test_downloadonly.py
oliverkurth: This is `dnf` behavior. I think we should be trying to make it similar to their behavior to avoid confusion.

client/remoterepo.c
oliverkurth: The man page for `basename` documents no error case, so not sure if needed.

client/remoterepo.c
oliverkurth: Thx, done.

client/remoterepo.c
oliverkurth: I indented the 2nd line more, maybe that looks better.

oliverkurth:
The failed test is unrelated. The container doesn't build because of an unsigned package:
```
Step 4/5 : RUN dnf -q -y install gcc make cmake libcurl-devel rpm-devel rpm-build libsolv-devel                       popt-devel sed createrepo_c glib2-devel libxml2 findutils                       python3-pytest python3-requests python3-urllib3 python3-pyOpenSSL                       python3 python3-devel valgrind gpgme-devel libmetalink-devel                       openssl-devel rpm-sign which
 ---> Running in 85736030c516
Package libuv-1.40.0-1.fc32.x86_64.rpm is not signed
Error: GPG check FAILED
The command '/bin/sh -c dnf -q -y install gcc make cmake libcurl-devel rpm-devel rpm-build libsolv-devel                       popt-devel sed createrepo_c glib2-devel libxml2 findutils                       python3-pytest python3-requests python3-urllib3 python3-pyOpenSSL                       python3 python3-devel valgrind gpgme-devel libmetalink-devel                       openssl-devel rpm-sign which' returned a non-zero code: 1
The command "docker build -t ${DIST}/tdnf-build -f ci/Dockerfile.${DIST} ." exited with 1.
```

oliverkurth:
I have added some output to clarify that `tdnf` will only download, similar to what `dnf` does:
```
root [ /build/build ]# ./bin/tdnf --downloadonly --downloaddir=/tmp install less

Installing:
less                                        x86_64                530-2.ph3                     photon-updates        236.70k 242379

Total installed size: 236.70k 242379
tdnf will only download packages needed for the transaction
Is this ok [y/N]: y

Downloading:
less                                    130134   100%

Complete!
Packages have been downloaded to /tmp.
root [ /build/build ]# ./bin/tdnf --downloadonly install less

Installing:
less                                        x86_64                530-2.ph3                     photon-updates        236.70k 242379

Total installed size: 236.70k 242379
tdnf will only download packages needed for the transaction
Is this ok [y/N]: y

Downloading:

Complete!
Packages have been downloaded to cache.
root [ /build/build ]# 
```

keerthanakalyan (approved):
LGTM

client/remoterepo.c
```diff
@@ -1082,3 +1082,59 @@ TDNFDownloadPackageToCache(
...
+    const char* pszPkgName,
+    const char* pszRepoName,
+    const char* pszDirectory,
+    char** ppszFilePath
+    )
+{
+    uint32_t dwError = 0;
+    char* pszFilePath = NULL;
+    char* pszRemotePath = NULL;
+    char* pszFileName;
```
tapakund: init pszFileName with NULL

tapakund (commented):
LGTM

client/remoterepo.c
oliverkurth: Done.

tapakund (approved):

PR #201, opened 2020-12-02
optimise makecache path

tapakund:
- In case of using baseurl:
  When we do makecache first time, tdnf will
  download the repomd.xml and copy all the
  repodata.
  From second time, if we do "makecache" or
  "repolist --refresh" we will check the
  shasum of repomd.xml with the existing
  repomd.xml and if both are not same, will
  replace the repodata.
  If we switch from metalink to baseurl. In that case
  we will download he repomd.xml and compare with
  the existing one. If both are same, we will not
  replace and use the existing one.

- In case of metalink:
  If we were using baseurl and switched to metalink.
  We will download metalink file and then download
  repomd.xml and compare the shasum with existing
  repomd.xml. If both matches we will not replace
  the repomd.xml. We will just keep the metalink file
  and save the baseurl.
  From next time onwards, we will check the shasum
  for metalink file, if it matches we will not proceed
  to download the repomd.xml.

### DIFF
--- a/client/repo.c
+++ b/client/repo.c
@@ -84,7 +84,7 @@ TDNFInitRepo(
     pSolvRepoInfo->pRepo = pRepo;
     pRepo->appdata = pSolvRepoInfo;
 
-    dwError = SolvCalculateCookieForRepoMD(pRepoMD->pszRepoMD, pSolvRepoInfo->cookie);
+    dwError = SolvCalculateCookieForFile(pRepoMD->pszRepoMD, pSolvRepoInfo->cookie);
     BAIL_ON_TDNF_ERROR(dwError);
 
     pSolvRepoInfo->nCookieSet = 1;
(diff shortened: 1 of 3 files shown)


tapakund:
cc @keerthanakalyan @oliverkurth @sshedi

client/repo.c
```diff
@@ -886,65 +902,59 @@ TDNFGetRepoMD(
...
+        dwError = TDNFRepoSetBaseUrl(pTdnf, pRepoData, pszBaseUrlFile);
+        BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    if (nReplaceRepoMD)
+    {
+        /* Remove the old repodata, solvcache and lastRefreshMarker before replacing the new repomd file and metalink files. */
+        TDNFRepoRemoveCache(pTdnf, pRepoData->pszId);
+        TDNFRemoveSolvCache(pTdnf, pRepoData->pszId);
+        TDNFRemoveLastRefreshMarker(pTdnf, pRepoData->pszId);
```
keerthanakalyan, line 946: Should we remove rpmcache (TDNFRemoveRpmCache()) also based on the flag pTdnf->pConf->nKeepCache?

tapakund: Sure @keerthanakalyan , now we are removing the cached rpms if ```keepcache=true``` is not set in tdnf.conf file. Thanks!

keerthanakalyan (approved):
LGTM

PR #202, opened 2020-12-03
Further fixes based on coverty scan report

sshedi:

### DIFF
--- a/client/repoutils.c
+++ b/client/repoutils.c
@@ -458,7 +458,14 @@ cleanup:
     }
     if (!IsNullOrEmptyString(pszRpmCacheDir))
     {
-        rmdir(pszRpmCacheDir);
+        if (rmdir(pszRpmCacheDir))
+        {
+            /*
+             * Not using BAIL_ON_TDNF_SYSTEM_ERROR here to prevent infinite loop
+             * Also it makes easy to read the code
+             */
+            dwError = ERROR_TDNF_SYSTEM_BASE + errno;
+        }
     }
     TDNF_SAFE_FREE_MEMORY(pszFilePath);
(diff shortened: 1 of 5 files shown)


client/repoutils.c
```diff
@@ -458,7 +458,10 @@ TDNFRemoveRpmCache(
     }
     if (!IsNullOrEmptyString(pszRpmCacheDir))
     {
-        rmdir(pszRpmCacheDir);
+        if (rmdir(pszRpmCacheDir))
+        {
+            dwError = errno;
+        }
```
oliverkurth, line 468: I think we should have a `BAIL_ON_TDNF_ERROR` here. I know it makes no difference, but it would be cleaner.

oliverkurth (approved):
LGTM .

client/repoutils.c
sshedi: The idea here is, not to end up in an infinite loop. If I use BAIL_ON_TDNF_ERROR we go to error: and then again come back to cleanup: which makes it difficult to read the code because of jumps(which is why goto statements are discouraged). I personally prefer code moving in single direction, not back and forth.

oliverkurth: Yeah, it's a corner case. I'll leave t up to you. Maybe add a comment.

But - I just spotted another issue with this: this would make the function return `errno`. `BAIL_ON_TDNF_SYSTEM_ERROR` would make it return `ERROR_TDNF_SYSTEM_BASE + errno`. Functions are expected to return TDNF errors, not system errors.

sshedi: Good catch, fixed it. Thanks.

keerthanakalyan (approved):

Issue #90, opened 2019-11-15
CLI Argument validation should happen before any other actions are performed


sidcha:
For an invalid command construct, we must early-return to speedup tests runtime. In this case, metadata refresh should not happen.
```
root@photon-ea4f47d2e8d0 [ ~/tdnf/pytests ]# tdnf clean
Refreshing metadata for: 'VMware Photon Linux 3.0(x86_64) Updates'
Error(903) : Clean requires an option: packages, metadata, dbcache, plugins, expire-cache, all
```

ppadmavilasom:
please feel free to send PRs

sidcha:
Will do

oliverkurth:
Fixed with https://github.com/vmware/tdnf/commit/d3ecd04a8ef15f0366b482cc82d4b6b344d7b030 .

PR #206, opened 2020-12-10
Delay refreshing repo metadata to after argument validation

oliverkurth:
This addresses issue #https://github.com/vmware/tdnf/issues/90.

Also includes rearranging code in TDNFRefreshSack().

### DIFF
--- a/client/init.c
+++ b/client/init.c
@@ -205,9 +205,6 @@ TDNFRefreshSack(
         pTdnf->pArgs->nRefresh = 1;
     }
 
-    dwError = TDNFInitCmdLineRepo(pTdnf, pSack);
-    BAIL_ON_TDNF_ERROR(dwError);
-
     /* First repo is the "@cmdline" repo, which always exists.
      * skip over it - options do not apply, and it is initialized. */
     if(pTdnf->pRepos->pNext)
@@ -283,3 +280,11 @@ cleanup:
 error:
     goto cleanup;
 }
+
+uint32_t
+TDNFRefresh(
+    PTDNF pTdnf)
+{
+    return TDNFRefreshSack(pTdnf, pTdnf->pSack, pTdnf->pArgs->nRefresh);
+}
+
--- a/client/init.c
+++ b/client/init.c
@@ -194,6 +194,8 @@ TDNFRefreshSack(
     uint32_t dwError = 0;
     char* pszRepoCacheDir = NULL;
     int nMetadataExpired = 0;
+    PTDNF_REPO_DATA_INTERNAL pRepo = NULL;
+
     if(!pTdnf || !pTdnf->pArgs)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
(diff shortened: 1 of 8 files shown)


client/init.c
```diff
@@ -205,75 +207,74 @@ TDNFRefreshSack(
...
+                          &nMetadataExpired);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            TDNF_SAFE_FREE_MEMORY(pszRepoCacheDir);
+            pszRepoCacheDir = NULL;
+        }
+
+        if(nMetadataExpired)
+        {
+            dwError = TDNFRepoRemoveCache(pTdnf, pRepo->pszId);
```
keerthanakalyan, line 243: Should we also remove rpmcache based on pTdnf->pConf->nKeepCache check?

client/init.c
```diff
@@ -205,75 +207,74 @@ TDNFRefreshSack(
...
-        PTDNF_REPO_DATA_INTERNAL pTempRepo = pTdnf->pRepos->pNext;
-        while(pTempRepo)
+        /* skip the @cmdline repo - options do not apply, and it is
+           initialized. */
+        if (strcmp(pRepo->pszName, "@cmdline") == 0)
+        {
+            continue;
+        }
+
+        if(!pRepo->nEnabled)
```
sshedi: nit: the above two if conditions can be combined into one

client/init.c
```diff
@@ -205,75 +207,74 @@ TDNFRefreshSack(
...
+            BAIL_ON_TDNF_ERROR(dwError);
         }
+
+        if(pSack)
+        {
+            dwError = TDNFInitRepo(pTdnf, pRepo, pSack);
+        }
+        if(dwError)
+        {
+            if(pRepo->nSkipIfUnavailable)
```
sshedi: nit: these two ifs can be combined

tools/cli/lib/installcmd.c
```diff
@@ -186,6 +186,9 @@ TDNFCliAlterCommand(
                   &nPackageCount);
     BAIL_ON_CLI_ERROR(dwError);
 
+    dwError = TDNFCliRefresh(pContext);
```
sshedi, line 189: Why can't all these call to TDNFCliRefresh can be done in one place just befoe: https://github.com/vmware/tdnf/blob/99885bd58326bc9defd0d516f51e55da0adebaa9/tools/cli/main.c#L158

sshedi (approved):
LGTM, need to know reason behind calling TDNFCliRefresh at multiple places.

oliverkurth:
> LGTM, need to know reason behind calling TDNFCliRefresh at multiple places.

See https://github.com/vmware/tdnf/issues/90 . The issue is that when attempting to issue an invalid command, the repo metadata are refreshed before `tdnf` bails. This is especially annoying for the `clean` command because its whole point is to **delete** the cache, so data will be downloaded uselessly. But also for other commands we should bail quickly and not have to wait a minute to see the failure.

Each command has its own code path to validate arguments, therefore we need to put the call to refresh the metadata for each command separately.

Maybe this makes it clearer - old `tdnf` is installed, `./bin/tdnf` has the changes from this PR. Refreshing the metadata took about a minute:
```
root [ /build/build ]# tdnf clean all
Cleaning repos: photon-extras photon-updates photon
Cleaning up everything
root [ /build/build ]# tdnf clean    
Refreshing metadata for: 'VMware Photon Linux 3.0 (x86_64)'
Refreshing metadata for: 'VMware Photon Linux 3.0 (x86_64) Updates'
Refreshing metadata for: 'VMware Photon Extras 3.0 (x86_64)'
Error(903) : Clean requires an option: packages, metadata, dbcache, plugins, expire-cache, all

root [ /build/build ]# tdnf clean all
Cleaning repos: photon-extras photon-updates photon
Cleaning up everything
root [ /build/build ]# ./bin/tdnf clean    
Error(903) : Clean requires an option: packages, metadata, dbcache, plugins, expire-cache, all
root [ /build/build ]# 
```

client/init.c
oliverkurth: I think not. `TDNFRepoRemoveCache)` removes the metadata only. I think the `keepcache` option is meant just for packages, which are not removed here.

client/init.c
oliverkurth: This was intended because I thought it's easier to read, but I think it's fine when the two conditions are on separare lines.

client/init.c
oliverkurth: Thanks, fixed.

client/init.c
```diff
@@ -194,6 +194,8 @@ TDNFRefreshSack(
     uint32_t dwError = 0;
     char* pszRepoCacheDir = NULL;
     int nMetadataExpired = 0;
+    PTDNF_REPO_DATA_INTERNAL pRepo;
```
tapakund: lets initialise pRepo with NULL

tapakund (approved):
LGTM.

client/init.c
oliverkurth: Done

keerthanakalyan (approved):

client/init.c
```diff
@@ -283,3 +280,11 @@ TDNFRefreshSack(
 error:
     goto cleanup;
 }
+
+uint32_t
+TDNFRefresh(
+    PTDNF pTdnf)
+{
+    return TDNFRefreshSack(pTdnf, pTdnf->pSack, pTdnf->pArgs->nRefresh);
```
shreyasbe, line 284: Please add NULL Check for "pTdnf->pArgs".

PR #207, opened 2021-01-05
Implement repository priority option

oliverkurth:
Implement the `priority` option for repositories. A lower value for `priority` indicates higher priority. The default is `50`.

Implemented by sorting the repositories by the `priority` value and initializing them in that order using `TDNFInitRepo()`.

Tested by copying `/etc/yum.repos.d/photon-updates.repo` a few times and setting different random priority values. Observed that packages will be installed from repo with lowest priority value.

### DIFF
--- a/client/init.c
+++ b/client/init.c
@@ -194,6 +194,13 @@ error:
     goto cleanup;
 }
 
+static
+int _repo_compare(const void *ppRepo1, const void *ppRepo2)
+{
+    return (*(PTDNF_REPO_DATA_INTERNAL*)(ppRepo1))->nPriority -
+           (*(PTDNF_REPO_DATA_INTERNAL*)(ppRepo2))->nPriority;
+}
+
 uint32_t
 TDNFRefreshSack(
     PTDNF pTdnf,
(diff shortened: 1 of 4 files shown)


client/init.c
```diff
@@ -216,6 +226,32 @@ TDNFRefreshSack(
         {
             continue;
         }
+        nCount++;
+    }
+
+    /* nCount may be 0 if --disablerepo=* is used */
+    if (nCount > 0)
```
sshedi, line 243: no need of this if condition, you can move this block inside the above loop where you do nCount++

client/init.c
```diff
@@ -184,6 +184,13 @@ TDNFCloneSetOpts(
     goto cleanup;
 }
 
+static
+int _repo_compare(const void *ppRepo1, const void *ppRepo2)
+{
```
sshedi, line 199: Do we need to validate args?

sshedi: BTW, what's the least and highest priority value?

client/init.c
oliverkurth: Not sure what I am missing - the block starting in line 220 is a loop for counting the repos. Only after counting all (valid) repos we know how much memory to allocate and whether to sort, or skip that step when there is no valid repo.

sshedi: I thought instead of allocating for full array after taking nCount, we can allocate and reallocate where we do nCount now. Never mind, the existing version is better, memory allocations are costly operations.

sshedi (approved):

client/init.c
oliverkurth: I don't think we need to validate args in `repo_compare`. This is a basic helper function, only used for `qsort()`.

There are no limits to the priority value, except what's in the `int` range.

sshedi (approved):

PR #208, opened 2021-01-07
fix for 'none of the providers can be installed' false positive issue

sshedi:


oliverkurth:
Would be good to describe in PRs what is being fixed, and how it is addressed.

Is it intended to be merged into `vmware:dev-1.2.0` or is that a mistake?

keerthanakalyan (approved):

oliverkurth:
This PR is more than a year old - is it still valid? Should it still be merged, or can we discard it?

PR #210, opened 2021-01-12
Fixes based on coverity scan

tapakund:
Signed-off-by: Tapas Kundu <tkundu@vmware.com>

PR #211, opened 2021-01-12
Address Coverity issues

oliverkurth:
Addressing all coverity issues, except two resource leaks that will be addressed with https://github.com/vmware/tdnf/pull/210 . Most issues are addressed with code changes, except 4 that are addressed with comments to make coverity ignore them. The two 'deadcode' issues are left for possible future code changes. The two 'toctou' issues can be ignored because no real harm would occur in a potential race condition. 

Analysis summary report:
------------------------
Files analyzed                 : 53
Total LoC input to cov-analyze : 54535
Functions analyzed             : 478
Paths analyzed                 : 139338
Time taken by analysis         : 00:00:26
Defect occurrences found       : 6 Total
                                 2 DEADCODE
                                 2 RESOURCE_LEAK
                                 2 TOCTOU

Processing 2 errors.

0%   10   20   30   40   50   60   70   80   90   100%
|----|----|----|----|----|----|----|----|----|----|
***************************************************
Ignored 4 errors due to coverity_ignore comments in the code.
Processed errors.

### DIFF
--- a/include/tdnf-common-defines.h
+++ b/include/tdnf-common-defines.h
@@ -33,6 +33,12 @@
         }                                                   \
     } while(0)
 
+#define BAIL_ON_TDNF_SYSTEM_ERROR_UNCOND(dwError)                  \
+    do {                                                    \
+        dwError = ERROR_TDNF_SYSTEM_BASE + dwError;     \
+        goto error;                                     \
+    } while(0)
+
 #define TDNF_SAFE_FREE_MEMORY(pMemory)          \
     do {                                        \
         if (pMemory) {                          \
--- a/client/defines.h
+++ b/client/defines.h
@@ -203,4 +203,5 @@ typedef enum
     {ERROR_TDNF_CURLE_UNSUPPORTED_PROTOCOL,          "ERROR_TDNF_CURLE_UNSUPPORTED_PROTOCOL",          "Curl doesn't Support this protocol"},\
     {ERROR_TDNF_CURLE_FAILED_INIT,                   "ERROR_TDNF_CURLE_FAILED_INIT",                   "Curl Init Failed"},\
     {ERROR_TDNF_CURLE_URL_MALFORMAT,                 "ERROR_TDNF_CURLE_URL_MALFORMAT",                 "URL seems to be corrupted. Please clean all and makecache"},\
+    {ERROR_TDNF_SYSTEM_BASE,                         "ERROR_TDNF_SYSTEM_BASE",                         "unknown system error"},\
 };
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -145,7 +145,7 @@ TDNFGetDigestForFile(
     {
         pr_err("Metalink: validating (%s) FAILED\n", filename);
         dwError = errno;
-        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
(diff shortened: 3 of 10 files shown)


include/tdnf-common-defines.h
```diff
@@ -33,6 +33,12 @@
         }                                                   \
     } while(0)
 
+#define BAIL_ON_TDNF_SYSTEM_ERROR0(dwError)                  \
```
keerthanakalyan: I'm not able to understand its usage. We set the errno to dwError before calling this macro. Why can't we use the BAIL_ON_SYSTEM_ERROR instead ?

oliverkurth: Sorry, I should have explained that one.

The reason is that in lots of places we set dwError to errno, based on the assumption that `errno != 0`. For example:
```
111  	    fp = fopen(pszFile, "w");
112  	    if (!fp)
113  	    {
114  	        dwError = errno;
115  	        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
116  	    }
```
`BAIL_ON_TDNF_SYSTEM_ERROR` would then `goto error:`, but only if `dwError != 0`. But coverity does not make that assumption, so in the case that `!fp` is `true` and `errno==0` the code would continue without bailing out, with an invalid `fp`. To fix that, I created the `BAIL_ON_TDNF_SYSTEM_ERROR0` macro that would bail in any case, regardless of the value of dwError (that is even if `errno==0`). It makes sense too, because an error has definitely happened although errno is not set. tdnf would show "unknown system error".

This is what coverity showed before the change:
```
111  	    fp = fopen(pszFile, "w");
(6) Event cond_true:	Condition "!fp", taking true branch.
(7) Event var_compare_op:	Comparing "fp" to null implies that "fp" might be null.
Also see events:	[var_deref_model]
112  	    if (!fp)
113  	    {
114  	        dwError = errno;
(8) Event cond_false:	Condition "dwError", taking false branch.
(9) Event if_end:	End of if statement.
115  	        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
116  	    }
(10) Event var_deref_model:	Passing null pointer "fp" to "fputs", which dereferences it.
Also see events:	[var_compare_op]
117  	    fputs(data, fp);
118  	    fclose(fp);
119  
```

sidcha: You could just add ERROR_BASE to errno in-place and force it to always take the exit path when `errno == 0` without introducing a new macro. This makes it very obvious to the readers on what you are trying to do. Besides, it's not that much of a common use case.

At the very least, this macro has to be refactored to something else (BAIL_UNCOND pops up immediately) as:
- this is identical to the conditional bail at first glance.
- doens't convey tha lack of check.

oliverkurth: I like `BAIL_ON_TDNF_SYSTEM_ERROR_UNCOND`, much better than `BAIL_ON_TDNF_SYSTEM_ERROR0`.

keerthanakalyan: Yes, `BAIL_ON_TDNF_SYSTEM_ERROR_UNCOND` sounds good.

sshedi (approved):

include/tdnf-common-defines.h
oliverkurth: Done.

keerthanakalyan (approved):

PR #212, opened 2021-01-13
Fix 'clean all' when already clean

oliverkurth:
When doing `tdnf clean all` after another `tdnf clean all`, `tdnf` would throw an error because it did not find the cache directories (thanks to @tapakund for pointing that out). This issue only became apparent with https://github.com/vmware/tdnf/pull/206 . This PR fixes that issue.

During testing I found also found that `pszVarBaseArch` wouldn't be set if the config file did not include any substitution variables.

Also adds two tests for `tdnf clean all`.

### DIFF
--- a/client/config.c
+++ b/client/config.c
@@ -197,7 +197,6 @@ TDNFConfigExpandVars(
         dwError = TDNFGetKernelArch(&pConf->pszVarBaseArch);
         BAIL_ON_TDNF_ERROR(dwError);
     }
-
 cleanup:
     return dwError;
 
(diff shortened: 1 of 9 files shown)


client/repoutils.c
```diff
@@ -259,46 +259,49 @@ TDNFRepoRemoveCache(
...
     pDir = opendir(pszRepoCacheDir);
-    if(pDir == NULL)
+    if(pDir == NULL && errno != ENOENT)
     {
         dwError = errno;
         BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
     }
 
-    while ((pEnt = readdir (pDir)) != NULL )
+    if (pDir != NULL)
```
sshedi: This works. Please check https://man7.org/linux/man-pages/man3/nftw.3p.html
Please remove the code duplication TDNFRemoveRpmCache & TDNFRepoRemoveCache functions. We can have a single function to delete a directory.

sshedi (changes requested):

client/repoutils.c
oliverkurth: I think I agree, we should have a single function to recursively remove directories. I have just found that RPMs will not be removed at all unless they are in a flat `x86_64` or `noarch` directory, see https://github.com/vmware/tdnf/issues/213. If the structure is different, RPMs will not be cleaned. Fixing this is more complex, and should go into a separate change.

oliverkurth: Directories are now removed recursively.

oliverkurth:
Question: should we remove downloaded keys too?

keerthanakalyan:
> Question: should we remove downloaded keys too?



> Question: should we remove downloaded keys too?

Yes, It would be a good idea to clean downloaded keys as well since the command is `clean all`.

oliverkurth:
> > Question: should we remove downloaded keys too?
> 
> > Question: should we remove downloaded keys too?
> 
> Yes, It would be a good idea to clean downloaded keys as well since the command is `clean all`.

Done.

client/config.c
```diff
@@ -313,37 +312,32 @@ TDNFConfigReplaceVars(
...
+    dwError = TDNFConfigExpandVars(pTdnf);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    /* fill variable values such as release and basearch
+       if required */
     if(strstr(*ppszString, TDNF_VAR_RELEASEVER) ||
        strstr(*ppszString, TDNF_VAR_BASEARCH))
     {
-        dwError = TDNFConfigExpandVars(pTdnf);
+        pConf = pTdnf->pConf;
```
tapakund, line 323: Can you please explain this commit https://github.com/vmware/tdnf/pull/212/commits/a3d805a7e2b1305823cdc6b8afe50ce54bba25c2

client/repoutils.c
```diff
@@ -603,55 +392,50 @@ TDNFRemoveSolvCache(
                   TDNF_SOLVCACHE_DIR_NAME);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    pDir = opendir(pszSolvCacheDir);
-    if(pDir == NULL)
+    TDNFRecursivelyRemoveDir(pszSolvCacheDir);
```
tapakund: we need to get the return from TDNFRecursivelyRemoveDir  to dwError

client/repoutils.c
```diff
@@ -603,55 +392,50 @@ TDNFRemoveSolvCache(
...
-    TDNF_SAFE_FREE_MEMORY(pszSolvCacheDir);
-    if(pDir)
+    dwError = TDNFAllocateStringPrintf(
+                  &pszKeysDir,
+                  "%s/%s/keys",
+                  pTdnf->pConf->pszCacheDir,
+                  pszRepoId);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    TDNFRecursivelyRemoveDir(pszKeysDir);
```
tapakund: not setting dwError here as well

common/utils.c
```diff
@@ -7,6 +7,7 @@
  */
 
 #include "includes.h"
+#include <ftw.h>
```
tapakund, line 9: we should have #include <ftw.h> before to #include "includes.h"

tapakund (changes requested):

client/config.c
oliverkurth: Before, `TDNFConfigExpandVars()` would only be executed if `TDNF_VAR_RELEASEVER` or `TDNF_VAR_BASEARCH` was present in the config file, so if there are no `$releasever` or `$basearch`. But in that case, `pConf->pszVarBaseArch` wouldn't be set.

This issue is not directly related, but I discovered it by working on this PR.

client/repoutils.c
oliverkurth: Done

client/repoutils.c
oliverkurth: Done

common/utils.c
oliverkurth: Done

PR #217, opened 2021-01-20
use %{_unitdir} for SYSTEM_DIR in rpm build script

oliverkurth:
Newer systemd uses `/usr/lib`. tdnf needs to install to the proper directory. This fixes it automatically in the rpm build script.


sshedi (approved):
LGTM

PR #222, opened 2021-02-05
ignore unset pTdnf->pArgs in TDNFRefresh()

oliverkurth:
As requested by @shreyasbe , this is to fix pmd.

### DIFF
--- a/client/init.c
+++ b/client/init.c
@@ -206,13 +206,13 @@ TDNFRefreshSack(
     int nMetadataExpired = 0;
     PTDNF_REPO_DATA_INTERNAL pRepo = NULL;
 
-    if(!pTdnf || !pTdnf->pArgs)
+    if(!pTdnf)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    if (nCleanMetadata == 1)
+    if (nCleanMetadata == 1 && pTdnf->pArgs)
     {
         pTdnf->pArgs->nRefresh = 1;
     }
@@ -291,6 +291,11 @@ uint32_t
 TDNFRefresh(
     PTDNF pTdnf)
 {
+    uint32_t dwError = 0;
+    if(!pTdnf || !pTdnf->pSack || !pTdnf->pArgs)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        return dwError;
+    }
     return TDNFRefreshSack(pTdnf, pTdnf->pSack, pTdnf->pArgs->nRefresh);
 }
-
--- a/client/api.c
+++ b/client/api.c
@@ -165,6 +165,9 @@ TDNFAlterCommand(
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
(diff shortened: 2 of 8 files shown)


client/init.c
```diff
@@ -291,6 +291,13 @@ uint32_t
 TDNFRefresh(
     PTDNF pTdnf)
 {
-    return TDNFRefreshSack(pTdnf, pTdnf->pSack, pTdnf->pArgs->nRefresh);
+    if (pTdnf->pArgs)
```
sshedi: Please validate pTdnf & pTdnf->pSack.

client/init.c
```diff
@@ -291,6 +291,13 @@ uint32_t
...
+    {
+        return TDNFRefreshSack(pTdnf, pTdnf->pSack, pTdnf->pArgs->nRefresh);
+    }
+    else
```
sshedi: else is redundant here.

sshedi (changes requested):

oliverkurth:
So looks like I broke the API with PR206 - it caused the TDNFRefresh() call to be made in the client (!), and this of course failed. I now moved the calls to `TDNFRefresh()` to client/api.c where they belong. This makes the calls to be made by the server. Similarly, the call to `TDNFAddCmdLinePackages()` had to be moved.

I tested this various invocations of `pmd-cli` when built with these changes and they looked okay. The `pmd-cli pkg check` command segfaults, but I believe this was broken before. @tapakund , please rebuild pmd with these changes and test thoroughly.

jaankit:
> So looks like I broke the API with PR206 - it caused the TDNFRefresh() call to be made in the client (!), and this of course failed. I now moved the calls to `TDNFRefresh()` to client/api.c where they belong. This makes the calls to be made by the server. Similarly, the call to `TDNFAddCmdLinePackages()` had to be moved.
> 
> I tested this various invocations of `pmd-cli` when built with these changes and they looked okay. The `pmd-cli pkg check` command segfaults, but I believe this was broken before. @tapakund , please rebuild pmd with these changes and test thoroughly.

"pmd-cli pkg check" is failing because corresponding API is not implemented(pkg_invoke_check()) in pmd when it gets called from tdnf but "check" is added in cmdlist at here: https://github.com/vmware/pmd/blob/1ffd119b8831829b30714e3efd6ad48b86c93c9f/tools/cli/pkgmgmt/pkg_main.c#L29 . Thus, it is crashing. This need to be fixed in pmd. This issue was present in old version of pmd also.

Tested some of the command with this change. Works fine. Thorough testing is in progress.

client/init.c
```diff
@@ -291,6 +291,10 @@ uint32_t
 TDNFRefresh(
     PTDNF pTdnf)
 {
+    if (!pTdnf->pArgs)
```
tapakund: Can we modify this function to:
uint32_t
TDNFRefresh(
    PTDNF pTdnf)
{
    uint32_t dwError = 0;
    if(!pTdnf || !pTdnf->pSack || !pTdnf->pArgs)
    {
        return dwError;
    }
    return TDNFRefreshSack(pTdnf, pTdnf->pSack, pTdnf->pArgs->nRefresh);
}

tapakund (changes requested):
All other changes looks fine.

client/init.c
oliverkurth: Done.

PR #223, opened 2021-02-18
print updateinfo and other warnings even if in quiet mode

prashant1221:
print output of updateinfo and some other errors and warning
even if quiet mode is set.


tapakund (approved):
LGTM

oliverkurth (approved):

PR #225, opened 2021-03-02
'rpmbuild --sign' option is deprecated, use 'rpmsign' instead

oliverkurth:
With Photon 4.0 GA, the `rpmbuild` command no longer supports the `--sign` option. This broke the Travis CI tests. With this change RPMs are signed with the `rpmsign` command.

### DIFF
--- a/pytests/repo/setup-repo.sh
+++ b/pytests/repo/setup-repo.sh
@@ -53,9 +53,8 @@ EOF
 
 echo building packages
 rpmbuild  --define "_topdir ${BUILD_PATH}" \
-	  --define "__gpg_sign_cmd %%{__gpg} gpg --batch" \
-	  -r ${BUILD_PATH} -ba ${REPO_SRC_DIR}/*.spec --sign
-
+	  -r ${BUILD_PATH} -ba ${REPO_SRC_DIR}/*.spec
+rpmsign --addsign ${BUILD_PATH}/RPMS/*/*.rpm
 cp -r ${BUILD_PATH}/RPMS ${PUBLISH_PATH}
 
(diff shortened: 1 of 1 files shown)


sshedi (approved):

pytests/repo/setup-repo.sh
```diff
@@ -53,9 +53,8 @@ EOF
 
 echo building packages
 rpmbuild  --define "_topdir ${BUILD_PATH}" \
-	  --define "__gpg_sign_cmd %%{__gpg} gpg --batch" \
```
sshedi, line 56: I think we need to do this based on version of Photon/rpmbuild used.

sshedi (changes requested):

pytests/repo/setup-repo.sh
sshedi: Fedora-33 also moved to rpm-4.16.x, I think we are using fedora-32 in our ci, can you check with fedora:latest please?

oliverkurth: > I think we need to do this based on version of Photon/rpmbuild used.

I tested with Photon 3.0, works fine. Photon 2.0 has rpmsign too.

sshedi (approved):
LGTM

pytests/repo/setup-repo.sh
oliverkurth: > Fedora-33 also moved to rpm-4.16.x, I think we are using fedora-32 in our ci, can you check with fedora:latest please?

Just tested with `fedora:latest` too, it works.

PR #228, opened 2021-03-25
setopt: Support rpm transactions like noscripts etc

ssahani:
sudo tdnf install --setopt=tsflags=noscripts

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -31,6 +31,7 @@ TDNFRpmExecTransaction(
     int nKeepCachedRpms = 0;
     int nDownloadOnly = 0;
     TDNFRPMTS ts = {0};
+    PTDNF_CMD_OPT pSetOpt = NULL;
 
     if(!pTdnf || !pTdnf->pArgs || !pTdnf->pConf || !pSolvedInfo)
     {
@@ -65,6 +66,18 @@ TDNFRpmExecTransaction(
     }
 
     ts.nTransFlags = rpmtsSetFlags (ts.pTS, RPMTRANS_FLAG_NONE);
+    for (pSetOpt = pTdnf->pArgs->pSetOpt; pSetOpt; pSetOpt = pSetOpt->pNext)
+      {
+        if (pSetOpt->nType != CMDOPT_KEYVALUE && !strcasecmp(pSetOpt->pszOptName, "tsflags"))
+          {
+            continue;
+          }
+
+        if (!strcasecmp(pSetOpt->pszOptValue, "noscripts"))
+          {
+            ts.nTransFlags = rpmtsSetFlags (ts.pTS, (RPMTRANS_FLAG_NOSCRIPTS | _noTransScripts | _noTransTriggers));
+          }
+      }
 
     if(rpmtsSetRootDir (ts.pTS, pTdnf->pArgs->pszInstallRoot))
     {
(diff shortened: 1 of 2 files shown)


client/rpmtrans.c
```diff
@@ -65,6 +66,18 @@ TDNFRpmExecTransaction(
...
 
     ts.nTransFlags = rpmtsSetFlags (ts.pTS, RPMTRANS_FLAG_NONE);
+    for (pSetOpt = pTdnf->pArgs->pSetOpt; pSetOpt; pSetOpt = pSetOpt->pNext)
+      {
+        if (pSetOpt->nType != CMDOPT_KEYVALUE)
+          {
+            continue;
+          }
+
+        if (!strcasecmp(pSetOpt->pszOptName, "tsflags") && !strcasecmp(pSetOpt->pszOptValue, "noscripts"))
```
oliverkurth: Please break this line into two.

common/setopt.c
```diff
@@ -52,6 +52,9 @@ AddSetOpt(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    dwError = AddSetOptWithValues(pCmdArgs, CMDOPT_KEYVALUE, pCmdOpt->pszOptName, pCmdOpt->pszOptValue);
```
oliverkurth, line 55: Please break this line.

oliverkurth (changes requested):
Looks good. Can you also add a test? Also good would be the output of a test that you have done.

client/rpmtrans.c
ssahani: This look up in O(n).  Why not use a hashtable to store all the options doing in O(1) ? Now I tweaked a bit so that this condition should not occur. More meant for optimization.

common/setopt.c
ssahani: why ?

oliverkurth: Just to make it more readable.

client/rpmtrans.c
oliverkurth: My request is just for formatting - the line is pretty long, so it should be broken to make it easier to read.

O(n) is not a problem here - this code is not time critical, and there aren't that many options to parse in a typical command line.

common/setopt.c
ssahani: not required

client/rpmtrans.c
ssahani: It's not abt one option or tons. We need to do that

common/setopt.c
ssahani: It's a bug here . It's getting parsed but not stored.

client/rpmtrans.c
```diff
@@ -65,6 +66,18 @@ TDNFRpmExecTransaction(
     }
 
     ts.nTransFlags = rpmtsSetFlags (ts.pTS, RPMTRANS_FLAG_NONE);
+    for (pSetOpt = pTdnf->pArgs->pSetOpt; pSetOpt; pSetOpt = pSetOpt->pNext)
+      {
+        if (pSetOpt->nType != CMDOPT_KEYVALUE && !strcasecmp(pSetOpt->pszOptName, "tsflags"))
```
oliverkurth, line 71: I think this should be `... || strcasecmp(pSetOpt->pszOptName, "tsflags")`.

common/setopt.c
oliverkurth: > not required

The line is too long (> 80 chars). Please try to keep lines shorter than that.

oliverkurth:
Thank you for the PR - I worked on a new version, see https://github.com/vmware/tdnf/pull/231 . Closing this one.

PR #232, opened 2021-04-19
move special metalink handling out of TDNFDownloadFile() to caller

oliverkurth:
Cleanup: there are 7 callers to `TDNFDownloadFile()`, only one of them cares about metalinks. The special code handling this is at the very end of the function, using the supplied arguments and can therefore be placed into the single caller. This way the other 6 callers do not need to supply the NULL args.

### DIFF
--- a/client/prototypes.h
+++ b/client/prototypes.h
@@ -245,9 +245,7 @@ TDNFDownloadFile(
     const char *pszRepo,
     const char *pszFileUrl,
     const char *pszFile,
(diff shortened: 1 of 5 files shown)


tapakund (approved):
LGTM

oliverkurth:
This change was also part of PR #233 (change a9c9e8e0c9df1de1511ba847772959aa18bd60ce), closing this PR.

PR #233, opened 2021-05-07
implement the reposync feature

oliverkurth:
This adds the "reposync" feature. This synchronizes a remote repository with a local one. By default, all packages will be downloaded to a local directory, unless they already exist. Optionally, metadata will be downloaded as well.

These options are implemented:
* `--delete`: remove old package that are not part of the repository any more
* `--download-metadata`: download metadata as well. The directory can be used as a repository after download.
* `--gpgcheck`: check the gpg signature. If invalid, the package will be deleted.
* `--norepopath`: no subdirectory with the repo name will be created. This option is only valid if there is more than one repository configured. It is also incompatible with the `--delete` option to prevent accidentally deleting unrelated packages.
* `--urls`: instead of downloading, the URLs of all files will be printed to `stdout`.
* `--download-path`: by default, files will be downloaded relative to the current directory. With this option another directory can be specified.
* `--metadata-path`: download metadata to another directory
* `--arch`: download specific architectures (can be given repeatedly)
* `--source`: download only source packages, same as `--arch src`. Incompatible with the `--arch` option. 
*  `--newest-only`: download latest versions only

The dnf reposync plugin has three more options that are not yet implemented in this PR: `--newest`, `--arch` and `--source`. These will be implemented in another PR.

Simple example usage, syncing a single repo:

`tdnf --disablerepo=* --enablerepo=photon-release reposync`

Explanation of changes (to help reviewers):

The main driver of the feature is the `TDNFRepoSync()` function in `client/api.c`. The function first generates a list of packages, this is very similar to `TDNFInfo()` and `TDNFList()`.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -1139,6 +1139,11 @@ TDNFRepoSync(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    if (pReposyncArgs->nNewestOnly)
+    {
+        TDNFPkgInfoFilterNewest(pTdnf->pSack, pPkgInfos);
+    }
+
     /* iterate through all packages */
     for (pPkgInfo = pPkgInfos; pPkgInfo; pPkgInfo = pPkgInfo->pNext)
     {
--- a/client/api.c
+++ b/client/api.c
@@ -1125,7 +1125,7 @@ TDNFRepoSync(
     }
     else
     {
-        dwError = TDNFAllocateString(pReposyncArgs->pszDownloadPath, &pszRootPath);
+        dwError = TDNFNormalizePath(pReposyncArgs->pszDownloadPath, &pszRootPath);
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
@@ -1154,7 +1154,7 @@ TDNFRepoSync(
             dwError = TDNFUtilsMakeDir(pszDir);
             BAIL_ON_TDNF_ERROR(dwError);
 
-            dwError = TDNFDownloadPackageToDirectory(pTdnf,
+            dwError = TDNFDownloadPackageToTree(pTdnf,
                             pPkgInfo->pszLocation, pPkgInfo->pszName,
                             pPkgInfo->pszRepoName, pszDir,
                             &pszFilePath);
--- a/client/api.c
+++ b/client/api.c
@@ -1056,7 +1056,7 @@ TDNFRepoSync(
     uint32_t dwRepoCount = 0;
     TDNFRPMTS ts = {0};
 
-    if(!pTdnf || !pTdnf->pSack)
+    if(!pTdnf || !pTdnf->pSack || !pReposyncArgs)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
@@ -1194,9 +1194,9 @@ TDNFRepoSync(
         else
         {
             dwError = TDNFCreatePackageUrl(pTdnf,
-                                            pPkgInfo->pszRepoName,
-                                            pPkgInfo->pszLocation,
-                                            &pszUrl);
+                                           pPkgInfo->pszRepoName,
+                                           pPkgInfo->pszLocation,
+                                           &pszUrl);
             BAIL_ON_TDNF_ERROR(dwError);
 
             pr_info("%s\n", pszUrl);
--- a/client/api.c
+++ b/client/api.c
@@ -980,6 +980,144 @@ error:
     goto cleanup;
 }
 
+uint32_t
+TDNFRepoSync(
+    PTDNF pTdnf,
+    PTDNF_REPOSYNC_ARGS pReposyncArgs
+    )
+{
+    uint32_t dwError = 0;
+    PTDNF_PKG_INFO pPkgInfos = NULL;
+    PTDNF_PKG_INFO pPkgInfo = NULL;
+    PTDNF_REPO_DATA_INTERNAL pRepo = NULL;
+    char *pszRepoDir = NULL;
+    PSolvQuery pQuery = NULL;
+    PSolvPackageList pPkgList = NULL;
+    char *pszRootPath = NULL;
+    char *pszDir = NULL;
+    char *pszFilePath;
+    uint32_t dwCount = 0;
+
+    if(!pTdnf || !pTdnf->pSack)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFRefresh(pTdnf);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = SolvCreateQuery(pTdnf->pSack, &pQuery);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFApplyScopeFilter(pQuery, SCOPE_ALL);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = SolvApplyListQuery(pQuery);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = SolvGetQueryResult(pQuery, &pPkgList);
+    BAIL_ON_TDNF_ERROR(dwError);
+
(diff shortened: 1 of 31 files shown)


client/api.c
```diff
@@ -980,6 +981,319 @@ TDNFRepoList(
...
+{
+    uint32_t dwError = 0;
+    char *pszKeepFile = NULL;
+    struct stat statKeep;
+
+    UNUSED(sbuf);
+    UNUSED(type);
+    UNUSED(ftwb);
+
+    if (strcmp(&pszFilePath[strlen(pszFilePath)-4], ".rpm") == 0)
```
sshedi, line 1000: This can be improved. Instead of comparing file extension, we can programmatically determine file type.
Explore a bit on <magic.h> (available in file-devel package)
Sample code:
```
#include <stdio.h>
#include <string.h>
#include <magic.h>

int main(void)
{
  char *fname = "file.rpm";
  struct magic_set *magic = magic_open(MAGIC_MIME | MAGIC_CHECK);

  magic_load(magic, NULL);

  const char *d = magic_file(magic, fname);

  printf("%s\n", d);

  printf("File is RPM? : %s\n", strstr(d, "x-rpm") ? "yes" : "no");

  magic_close(magic);

  return 0;
}
```
Missing few validations, can we use it?

sshedi (changes requested):

client/api.c
oliverkurth: This is not the right place for this check. The point of this function is to remove files from the repository directories that are no longer part of the repository. Whether they are valid RPMs is not even relevant - if they were not part of the repo, but live in the directory they need to be removed, no matter if they are valid.

Maybe this check would make sense just after downloading, but even then this would be caught either with the signature check, or when trying to read the header.

tools/cli/main.c
```diff
@@ -55,6 +55,8 @@ int main(int argc, char* argv[])
         {"reinstall",          TDNFCliReinstallCommand},
         {"remove",             TDNFCliEraseCommand},
         {"repolist",           TDNFCliRepoListCommand},
+        {"reposync",           TDNFCliRepoSyncCommand},
+        {"provides",           TDNFCliProvidesCommand},
```
keerthanakalyan: Is this (provides) added by mistake? I see that this is also getting removed in some other commit.

client/api.c
```diff
@@ -980,6 +981,352 @@ TDNFRepoList(
...
+                            pPkgInfo->pszLocation, pPkgInfo->pszName,
+                            pPkgInfo->pszRepoName, pszDir,
+                            &pszFilePath);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            /* if gpgcheck option is given, check for a valid signature. If that fails,
+               delete the package */
+            if (pReposyncArgs->nGPGCheck)
+            {
+                dwError = TDNFGPGCheckPackage(&ts, pTdnf, pPkgInfo->pszRepoName, pszFilePath, NULL);
```
keerthanakalyan, line 1200: Shouldn't we BAIL from this error?

client/remoterepo.c
```diff
@@ -986,6 +1015,49 @@ TDNFDownloadPackageToCache(
...
+cleanup:
+    TDNF_SAFE_FREE_MEMORY(pszNormalRpmCacheDir);
+    TDNF_SAFE_FREE_MEMORY(pszRpmCacheDir);
+    return dwError;
+error:
+    goto cleanup;
+}
+
+uint32_t
+TDNFDownloadPackageToTree(
```
keerthanakalyan, line 1043: What is the main difference between TDNFDownloadPackageToTree() and the existing TDNFDownloadPackageToDirectory() function? Both have same args and calls TDNFDownloadPackage().
Names are confusing, if you could add some comments about these functions it would be easy to understand.

common/utils.c
```diff
@@ -731,3 +731,21 @@ TDNFRecursivelyRemoveDir(const char *pszPath)
 error:
     goto cleanup;
 }
+
+/* search pszSearch in the string ppszList, result will be in pRet */
+uint32_t
+TDNFStringMatchesOneOf(const char *pszSearch, char **ppszList, int *pRet)
```
keerthanakalyan, line 738: Argument validation is missing

tools/cli/lib/parseargs.c
```diff
@@ -80,6 +83,17 @@ static struct option pstOptions[] =
     {"disableexcludes", no_argument, &_opt.nDisableExcludes, 1}, //--disableexcludes
     {"downloadonly",  no_argument, &_opt.nDownloadOnly, 1}, //--downloadonly
     {"downloaddir",   required_argument, 0, 0},            //--downloaddir
+    // reposync options
+    {"arch",          required_argument, 0, 0},
```
keerthanakalyan, line 87: These new commands and options needs to be added to help.c

client/api.c
```diff
@@ -980,6 +981,352 @@ TDNFRepoList(
     goto cleanup;
 }
 
+static int
```
keerthanakalyan, line 984: Return type is uint32_t

tools/cli/main.c
oliverkurth: Yes, that was a mistake, and I reverted this in another commit.

client/api.c
oliverkurth: No, that's not how the dnf plugin behaves.

client/api.c
oliverkurth: No, it needs to be `int`. This callback is used for `ntfw()`, and needs to match the prototype. See `man 3 nftw`:
```
       int nftw(const char *dirpath,
               int (*fn) (const char *fpath, const struct stat *sb,
                          int typeflag, struct FTW *ftwbuf),
               int nopenfd, int flags);
```

common/utils.c
oliverkurth: Thanks! Fixed.

client/remoterepo.c
oliverkurth: Good point. I have added comments.

tools/cli/lib/parseargs.c
oliverkurth: Good point. Fixed.

client/api.c
keerthanakalyan: What if there is an error which is not of rpm GPG error something like NULL pointer? In that case we will be proceeding?

client/api.c
keerthanakalyan: In that case, should we change the dwError to int or handle the error differently? 
As per man of ntfw():
`If fn() returns nonzero, then the tree walk is terminated and the value returned by fn() is returned as the result of ftw() or nftw().`
In case of error from this function, the return value will not be < 0 which is used in the code here:
` if (nftw(pszRepoDir, _rm_rpms, 10, FTW_DEPTH|FTW_PHYS) < 0)`

client/api.c
oliverkurth: Oh, you are right. Agreed, we should not continue for non gpg errors.

client/api.c
oliverkurth: Yes, you are right that we should bail in line 1260 on any non-zero return. A -1 return is presumably a system error in which case we need to check `errno` , but a value greater zero is a TDNF error generated by the callback. Good point!

client/api.c
oliverkurth: Fixed.

client/api.c
oliverkurth: Fixed.

keerthanakalyan (approved):

client/api.c
```diff
@@ -980,6 +981,368 @@ TDNFRepoList(
...
+_rm_rpms(
+    const char *pszFilePath,
+    const struct stat *sbuf,
+    int type,
+    struct FTW *ftwb
+    )
+{
+    uint32_t dwError = 0;
+    char *pszKeepFile = NULL;
+    struct stat statKeep;
```
tapakund: better to initialise statKeep with {0}

tapakund (approved):
overall LGTM.

client/api.c
oliverkurth: Thank you, fixed!

PR #236, opened 2021-06-02
fix segfault when no key is set in repo config

oliverkurth:
When gpgcheck is enabled but no key in the repo is configured tdnf crashes with a segfault. Fixing this, and adding a test to catch this error.

Example:
```
[docker-ce-stable]
name=Docker CE Stable - $basearch
baseurl=https://download.docker.com/linux/centos/8/$basearch/stable
enabled=1
gpgcheck=1
#gpgkey=https://download.docker.com/linux/centos/gpg
```


tapakund (approved):
LGTM

PR #239, opened 2021-06-10
use directories relative to installroot and add setopt=reposdir=<repodir>

### LINKED ISSUE #238: when using --installroot, tdnf should search for the config files relative to it
When using the `--installroot` option, tdnf searches for the repo configurations in the host. It would be more useful to search for these relative to the install root. dnf first searches under the install root, then under the host. This should be implemented in tdnf as well.

In addition, it would be good to implement `--setopt=reposdir=<reposdir>`.

oliverkurth:
This addresses issue #238 .

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -685,6 +685,9 @@ TDNFOpenHandle(
     uint32_t dwError = 0;
     PTDNF pTdnf = NULL;
     PSolvSack pSack = NULL;
+    char *pszCacheDir = NULL;
+    char *pszRepoDir = NULL;
+    int nHasOptReposdir = 0;
 
     if(!pArgs || !ppTdnf)
     {
@@ -703,6 +706,53 @@ TDNFOpenHandle(
                   pTdnf->pArgs->pszConfFile,
                   TDNF_CONF_GROUP);
     BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFHasOpt(pTdnf->pArgs, TDNF_SETOPT_KEY_REPOSDIR, &nHasOptReposdir);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (!IsNullOrEmptyString(pTdnf->pArgs->pszInstallRoot) &&
+        strcmp(pTdnf->pArgs->pszInstallRoot, "/"))
+    {
+        int nIsDir = 0;
+
+        dwError = TDNFAllocateStringPrintf(&pszCacheDir, "%s/%s",
+                                           pTdnf->pArgs->pszInstallRoot,
+                                           pTdnf->pConf->pszCacheDir);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        TDNF_SAFE_FREE_MEMORY(pTdnf->pConf->pszCacheDir);
+        pTdnf->pConf->pszCacheDir = pszCacheDir;
+        pszCacheDir = NULL;
+
+        if (!nHasOptReposdir)
+        {
+            dwError = TDNFAllocateStringPrintf(&pszRepoDir, "%s/%s",
+                                               pTdnf->pArgs->pszInstallRoot,
+                                               pTdnf->pConf->pszRepoDir);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            dwError = TDNFIsDir(pszRepoDir, &nIsDir);
+            if (dwError == ERROR_TDNF_FILE_NOT_FOUND)
+            {
+                nIsDir = 0;
+                dwError = 0;
+            }
+            BAIL_ON_TDNF_ERROR(dwError);
+            if (nIsDir)
+            {
+                TDNF_SAFE_FREE_MEMORY(pTdnf->pConf->pszRepoDir);
+                pTdnf->pConf->pszRepoDir = pszRepoDir;
+                pszRepoDir = NULL;
+            }
+        }
+    }
+
+    if (nHasOptReposdir)
+    {
+        TDNF_SAFE_FREE_MEMORY(pTdnf->pConf->pszRepoDir);
+        dwError = TDNFGetCmdOptValue(pTdnf->pArgs, TDNF_SETOPT_KEY_REPOSDIR, &pTdnf->pConf->pszRepoDir);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
 
     dwError = TDNFLoadPlugins(pTdnf);
     BAIL_ON_TDNF_ERROR(dwError);
@@ -729,6 +779,8 @@ TDNFOpenHandle(
     *ppTdnf = pTdnf;
 
 cleanup:
+    TDNF_SAFE_FREE_MEMORY(pszCacheDir);
+    TDNF_SAFE_FREE_MEMORY(pszRepoDir);
     return dwError;
 
 error:
--- a/client/init.c
+++ b/client/init.c
@@ -29,6 +29,7 @@ TDNFCloneCmdArgs(
     uint32_t dwError = 0;
     int nIndex = 0;
     PTDNF_CMD_ARGS pCmdArgs = NULL;
+    char *pszConfFileInstallRoot = NULL;
 
     dwError = TDNFAllocateMemory(
                   1,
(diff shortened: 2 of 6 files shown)


client/api.c
```diff
@@ -704,6 +707,49 @@ TDNFOpenHandle(
...
+        {
+            nIsDir = 0;
+            dwError = 0;
+        }
+        BAIL_ON_TDNF_ERROR(dwError);
+        if (nIsDir)
+        {
+            TDNF_SAFE_FREE_MEMORY(pTdnf->pConf->pszRepoDir);
+            pTdnf->pConf->pszRepoDir = pszRepoDir;
+            pszRepoDir = NULL;
```
keerthanakalyan: Why is it assigned to NULL here and in cleanup there is call to TDNF_SAFE_FREE_MEMORY(pszRepoDir)?  Is it done on purpose?

client/api.c
```diff
@@ -704,6 +707,49 @@ TDNFOpenHandle(
...
+        int nIsDir = 0;
+
+        dwError = TDNFAllocateStringPrintf(&pszCacheDir, "%s/%s",
+                                           pTdnf->pArgs->pszInstallRoot,
+                                           pTdnf->pConf->pszCacheDir);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        TDNF_SAFE_FREE_MEMORY(pTdnf->pConf->pszCacheDir);
+        pTdnf->pConf->pszCacheDir = pszCacheDir;
+        pszCacheDir = NULL;
```
keerthanakalyan, line 725: Why is it assigned to NULL here and in cleanup there is call to TDNF_SAFE_FREE_MEMORY(pszCacheDir)?  Is it done on purpose?

client/api.c
```diff
@@ -704,6 +707,49 @@ TDNFOpenHandle(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+        if (nIsDir)
+        {
+            TDNF_SAFE_FREE_MEMORY(pTdnf->pConf->pszRepoDir);
+            pTdnf->pConf->pszRepoDir = pszRepoDir;
+            pszRepoDir = NULL;
+        }
+    }
+
+    dwError = TDNFHasOpt(pTdnf->pArgs, "reposdir", &nHasOptReposdir);
```
keerthanakalyan: Instead of hardcoding "reposdir" it is good to have a macro.

client/api.c
```diff
@@ -704,6 +707,49 @@ TDNFOpenHandle(
                   TDNF_CONF_GROUP);
     BAIL_ON_TDNF_ERROR(dwError);
 
+    if (!IsNullOrEmptyString(pTdnf->pArgs->pszInstallRoot) &&
+        strcmp(pTdnf->pArgs->pszInstallRoot, "/"))
+    {
+        int nIsDir = 0;
```
sshedi, line 716: use bool

client/api.c
```diff
@@ -704,6 +707,49 @@ TDNFOpenHandle(
...
+            pTdnf->pConf->pszRepoDir = pszRepoDir;
+            pszRepoDir = NULL;
+        }
+    }
+
+    dwError = TDNFHasOpt(pTdnf->pArgs, "reposdir", &nHasOptReposdir);
+    BAIL_ON_TDNF_ERROR(dwError);
+    if (nHasOptReposdir)
+    {
+        TDNF_SAFE_FREE_MEMORY(pTdnf->pConf->pszRepoDir);
```
sshedi, line 752: Why can't we check for `TDNFHasOpt(pTdnf->pArgs, "reposdir", &nHasOptReposdir);` before `if (nIsDir)` ?

We allocate memory for pTdnf->pConf->pszRepoDir and do all that computation and later if we have "reposdir" option, we are freeing it. There is room for improvement here

client/init.c
```diff
@@ -62,10 +63,21 @@ TDNFCloneCmdArgs(
                   &pCmdArgs->pszInstallRoot);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    if(IsNullOrEmptyString(pCmdArgsIn->pszConfFile))
+    if(IsNullOrEmptyString(pCmdArgsIn->pszConfFile) &&
+       !IsNullOrEmptyString(pCmdArgsIn->pszInstallRoot))
     {
+        int nExists = 0;
```
sshedi, line 75: use bool, fix the same in TDNFIsFileOrSymlink function args

client/init.c
```diff
@@ -62,10 +63,21 @@ TDNFCloneCmdArgs(
                   &pCmdArgs->pszInstallRoot);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    if(IsNullOrEmptyString(pCmdArgsIn->pszConfFile))
+    if(IsNullOrEmptyString(pCmdArgsIn->pszConfFile) &&
+       !IsNullOrEmptyString(pCmdArgsIn->pszInstallRoot))
```
sshedi: We set pszInstallRoot to '/' in `TDNFCliParseArgs`, please re-check once

client/api.c
sshedi: pszCacheDir is re-used in this function and the pointer is stored in  pTdnf->pConf->pszCacheDir and later this is  freed in TDNFCloseHandle?

client/api.c
```diff
@@ -704,6 +707,49 @@ TDNFOpenHandle(
...
+        pTdnf->pConf->pszCacheDir = pszCacheDir;
+        pszCacheDir = NULL;
+
+        dwError = TDNFAllocateStringPrintf(&pszRepoDir, "%s/%s",
+                                           pTdnf->pArgs->pszInstallRoot,
+                                           pTdnf->pConf->pszRepoDir);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFIsDir(pszRepoDir, &nIsDir);
+        if (dwError == ERROR_TDNF_SYSTEM_BASE + ENOENT)
```
sshedi: Use ERROR_TDNF_FILE_NOT_FOUND from tdnferror.h.

Do the same in repoutils.c as well(in a different PR or piggyback it here, no issues). There are multiple entries of the same in repoutils.c

client/api.c
sshedi: Agreed. In case if we want to change name, we can do it in one place

sshedi (changes requested):
Few changes needed

client/api.c
keerthanakalyan: My point was there is no use of calling TDNF_SAFE_FREE_MEMORY(pszCacheDir) if pszCacheDir is assigned to NULL since TDNF_SAFE_FREE_MEMORY doesnt do anything is pointer is NULL. Also, TDNFCloseHandle calls TDNFFreeConfig which frees TDNF_SAFE_FREE_MEMORY(pConf->pszCacheDir). Wouldn't this cause a leak to pszCacheDir pointer?

keerthanakalyan: Nevermind, there is no leak here as the pointer is freed in TDNFFreeConfig, but there is no use of calling TDNF_SAFE_FREE_MEMORY(pszCacheDir) since pszCacheDir is assigned to NULL.

oliverkurth: There may be a bail to `error:` when `pszCacheDir` is set, in that case it needs to be freed. Actually, in this case the only is in line 718 caused by `TDNFAllocateStringPrintf()`, which would not set the pointer in case of an error, so it's not necessary - but the code may change, and free'ing the pointer may be forgotten. I think it's better practice to clean up just in case.

client/api.c
oliverkurth: Same reasoning as above for `pszCacheDir`, but in this case more valid since there is another `BAIL_ON_TDNF_ERROR` in between.

client/api.c
oliverkurth: The name won't change, this is the same option that dnf/yum have. But I can make it a macro anyway.

client/api.c
oliverkurth: Good point, improved.

client/init.c
oliverkurth: Good catch - got this in `client/api.c`, but missed it here.

client/init.c
oliverkurth: I think using boolean not consistent with the original coding style. Although booleans are used these seem to be the odd cases, and in most places it's int.

keerthanakalyan (approved):

PR #244, opened 2021-06-23
add photon 3.0 to ci tests

oliverkurth:
Might be a good idea to test in Photon 3 since latest tdnf is in Photon 3, and Photon 4 uses another libsolv version now.

### DIFF
--- a/ci/Dockerfile.photon-3.0
+++ b/ci/Dockerfile.photon-3.0
@@ -1,0 +1,17 @@
+FROM photon:3.0
+
+MAINTAINER okurth@vmware.com
+
+RUN tdnf update  -q -y
+RUN tdnf remove  -q -y toybox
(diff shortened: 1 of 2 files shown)


ci/Dockerfile.photon-3.0
```diff
@@ -0,0 +1,17 @@
+FROM photon:3.0
+
+MAINTAINER csiddharth@vmware.com
```
sshedi: Please change it to your name or remove it. Siddarth is not with VMware now

sshedi (approved):
One minor suggestion, LGTM

ci/Dockerfile.photon-3.0
oliverkurth: Totally missed that. Fixed.

PR #246, opened 2021-06-28
add --repoid/--repo and --repofrompath options

### LINKED ISSUE #234: Implement the --repoid option
Using `--disablerepo=* --enablerepo=some-rep` is tedious. The `--repoid some-repo` would make it easier to select a particular repo.

### LINKED ISSUE #235: Implement the repofrompath option
The `--repofrompath` option would make it easy to use an ad-hoc repository, for example one just created with the `reposync` command.

oliverkurth:
This addresses issues #234 and #235 . Usages:
```
tdnf --repo=<repoid> ...
tdnd --repofrompath=<repoid>,<url> --repo=<repoid> ...
```
Tests are added as well.

### DIFF
--- a/client/defines.h
+++ b/client/defines.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (C) 2015-2021 VMware, Inc. All Rights Reserved.
  *
  * Licensed under the GNU Lesser General Public License v2.1 (the "License");
  * you may not use this file except in compliance with the License. The terms
@@ -48,6 +48,8 @@ typedef enum
             goto error;                                            \
         }                                                          \
     } while(0)
+
+#define STR_IS_TRUE(s) ((s) && (!strcmp((s), "1") || !strcasecmp((s), "true")))
 
 //Misc
 #define TDNF_RPM_EXT                      ".rpm"
@@ -136,7 +138,8 @@ typedef enum
 #define TDNF_REPO_DEFAULT_SSLVERIFY       1
 #define TDNF_REPO_DEFAULT_RETRIES         10
 #define TDNF_REPO_DEFAULT_PRIORITY        50
-#define TDNF_REPO_DEFAULT_METADATA_EXPIRE "172800" // 48 hours in seconds
+#define TDNF_REPO_DEFAULT_METADATA_EXPIRE 172800 // 48 hours in seconds
+#define TDNF_REPO_DEFAULT_METADATA_EXPIRE_STR STRINGIFYX(TDNF_REPO_DEFAULT_METADATA_EXPIRE)
 
 // var names
 #define TDNF_VAR_RELEASEVER               "$releasever"
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -34,12 +34,14 @@ TDNFLoadRepoData(
     PTDNF_REPO_DATA_INTERNAL pReposTemp = NULL;
     PTDNF_REPO_DATA_INTERNAL pRepos = NULL;
     PTDNF_CONF pConf = NULL;
+    PTDNF_CMD_OPT pSetOpt = NULL;
     DIR *pDir = NULL;
     struct dirent *pEnt = NULL;
     int nLen = 0;
     int nLenRepoExt = 0;
+    char **ppszUrlIdTuple = NULL;
 
-    if(!pTdnf || !pTdnf->pConf || !ppReposAll)
+    if(!pTdnf || !pTdnf->pConf || !pTdnf->pArgs || !ppReposAll)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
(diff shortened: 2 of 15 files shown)


client/defines.h
```diff
@@ -46,6 +49,8 @@ typedef enum
         }                                                          \
     } while(0)
 
+#define STR_IS_TRUE(s) (s && (!strcmp(s, "1") || !strcasecmp(s, "true")))
```
sshedi: ignore case while checking for true.
use braces for `s`, we might pass an address with some pointer math

client/defines.h
```diff
@@ -28,6 +28,9 @@ typedef enum
     DETAIL_INFO
 }TDNF_PKG_DETAIL;
 
+#define STRINGIFYX(N) STRINGIFY(N)
```
sshedi: Not sure if feasible, can this be moved to some common place? So that other modules can also use it if needed.

client/defines.h
```diff
@@ -131,7 +136,8 @@ typedef enum
 #define TDNF_REPO_DEFAULT_SSLVERIFY       1
 #define TDNF_REPO_DEFAULT_RETRIES         10
 #define TDNF_REPO_DEFAULT_PRIORITY        50
-#define TDNF_REPO_DEFAULT_METADATA_EXPIRE "172800" // 48 hours in seconds
+#define TDNF_REPO_DEFAULT_METADATA_EXPIRE 172800 // 48 hours in seconds
+#define TDNF_REPO_DEFAULT_METADATA_EXPIRE_STR STRINGIFYX(TDNF_REPO_DEFAULT_METADATA_EXPIRE)
```
sshedi, line 142: good idea :)

client/repolist.c
```diff
@@ -49,6 +51,27 @@ TDNFLoadRepoData(
...
     BAIL_ON_TDNF_ERROR(dwError);
 
+    for(pSetOpt = pTdnf->pArgs->pSetOpt;
+        pSetOpt;
+        pSetOpt = pSetOpt->pNext)
+    {
+        if(strcmp(pSetOpt->pszOptName, "repofrompath") == 0)
+        {
+            TDNFSplitStringToArray(pSetOpt->pszOptValue, ",", &ppszUrlIdTuple);
+            if ((ppszUrlIdTuple[0] == NULL) || ppszUrlIdTuple[1] == NULL)
```
sshedi, line 61: Can you give an example of how inputs comes here? Will there be minimum 2 items in the ppszUrlIdTuple array always?

Can we give multiple repo paths at the same time?

client/repolist.c
```diff
@@ -49,6 +51,27 @@ TDNFLoadRepoData(
...
+            {
+                dwError = ERROR_TDNF_INVALID_PARAMETER;
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+            dwError = TDNFCreateRepoFromPath(&pReposAll,
+                ppszUrlIdTuple[0], ppszUrlIdTuple[1]);
```
sshedi: nit: indent it properly

client/repolist.c
```diff
@@ -111,6 +134,7 @@ TDNFLoadRepoData(
         closedir(pDir);
     }
     TDNF_SAFE_FREE_MEMORY(pszRepoFilePath);
+    TDNF_SAFE_FREE_STRINGARRAY(ppszUrlIdTuple);
```
sshedi, line 138: why is this needed? if ppszUrlIdTuple[1] is null and ppszUrlIdTuple[0] has garbage address, it might crash.

client/repolist.c
```diff
@@ -139,17 +163,124 @@ TDNFCreateCmdLineRepo(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    dwError = TDNFCreateRepo(&pRepo, "@cmdline");
```
sshedi: Please move `@cmdline` to a macro, it's usage is spread across files now.

client/repolist.c
```diff
@@ -139,17 +163,124 @@ TDNFCreateCmdLineRepo(
...
+    {
+        TDNFFreeReposInternal(pRepo);
+    }
+    goto cleanup;
+}
+
+uint32_t
+TDNFCreateRepoFromPath(
+    PTDNF_REPO_DATA_INTERNAL* ppRepo,
+    const char *pszId,
```
sshedi, line 187: validate function args

client/repolist.c
```diff
@@ -139,17 +163,124 @@ TDNFCreateCmdLineRepo(
...
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFSafeAllocateString("@cmdline", &pRepo->pszName);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    *ppRepo = pRepo;
+cleanup:
+    return dwError;
+error:
+    if(ppRepo)
```
sshedi: No need to set ppRepo to NULL

pytests/tests/test_repofrompath.py
```diff
@@ -0,0 +1,69 @@
...
+                     'makecache'],
+                    cwd=workdir)
+    assert(ret['retval'] == 0)
+
+    pkgname = utils.config["mulversion_pkgname"]
+    utils.erase_package(pkgname)
+    ret = utils.run(['tdnf',
+                     '-y', '--nogpgcheck',
+                     '--repofrompath=synced-repo,{}'.format(synced_dir),
+                     '--repo=synced-repo',
```
sshedi, line 64: Why do we need to specify `--repo` again? it's already given with `--repofrompath`
it will be good if we can specify multiple repos in one shot.
Like: `tdnf -y --repofrompath=repo-a,<repo-a-path>,repo-b,<repo-b-path>`

If what you have done is in alignment with dnf, let's use the same. This is just an idea

sshedi (changes requested):
Thanks for the feature. Minor changes needed.

client/defines.h
oliverkurth: The comparsion is case insensitive already - `strcasecmp()`.

Braces added, thanks!

client/defines.h
oliverkurth: Moved to `common/defines.h`.

client/repolist.c
oliverkurth: If the syntax is correctly given then `ppszUrlIdTuple` will have exactly two items: the repoid and the URL.

You can specify multiple paths by giving the option multiple times.

client/repolist.c
oliverkurth: How would `ppszUrlIdTuple[0]` have garbage? It's allocated and initialized in `TDNFSplitStringToArray()`. This is a pretty common pattern throughout the code.

pytests/tests/test_repofrompath.py
oliverkurth: Setting `--repo` disables all repos and enable only this one, which is what we want. It's a shortcut for `--disablerepo="*" --enablerepo=<repoid>`.

As explained in another comment, you can specify multiple repos with multiple `--repo` options.

client/repolist.c
oliverkurth: > Thanks for the feature. Minor changes needed.

Thanks for the review - I totally missed your comments for almost 3 weeks 😞 .

client/repolist.c
```diff
@@ -139,17 +164,120 @@ TDNFCreateCmdLineRepo(
...
+}
+
+uint32_t
+TDNFCreateRepoFromPath(
+    PTDNF_REPO_DATA_INTERNAL* ppRepo,
+    const char *pszId,
+    const char *pszPath
+    )
+{
+    uint32_t dwError;
```
tapakund: better to initialise dwError = 0 in all the function added.

tapakund (approved):
Looks good.

PR #248, opened 2021-07-07
implement the repoquery command

### LINKED ISSUE #74: Support for querying package dependencies
rpm -qR is the typical way of listing a package's dependencies. dnf has repoquery. tdnf doesn't seem to have a cli option for listing package dependencies. Would it be possible to enable dnf's repoquery option for tdnf?

oliverkurth:
This implements he `repoquery` command. This includes tests and help text.

Options implemented (see help.c):
```
    pr_crit("repoquery select options:\n");
    pr_crit("           [--available\n");
    pr_crit("           [--duplicates\n");
    pr_crit("           [--extras\n");
    pr_crit("           [--file <file>\n");
    pr_crit("           [--installed\n");
    pr_crit("           [--whatdepends <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatrequires <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatenhances <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatobsoletes <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatprovides <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatrecommends <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatrequires <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatsuggests <capability1>[,<capability2>[..]]]\n");
    pr_crit("           [--whatsupplements <capability1>[,<capability2>[..]]]\n");
    pr_crit("repoquery query options:\n");
    pr_crit("           [--depends\n");
    pr_crit("           [--enhances\n");
    pr_crit("           [--list\n");
    pr_crit("           [--obsoletes\n");
    pr_crit("           [--provides\n");
    pr_crit("           [--recommends\n");
    pr_crit("           [--requires\n");
    pr_crit("           [--requires-pre\n");
    pr_crit("           [--suggests\n");
    pr_crit("           [--supplements\n");
```

This includes the majority of select option of the original dnf, the missing ones are mostly not applicable. Still missing are query options (options that display details about the packages).

Example usage:
```
./bin/tdnf repoquery --disablerepo=* --enablerepo=photon-updates --whatrequires libsolv
tdnf-3.1.0-1.ph4.x86_64
libsolv-devel-0.6.35-8.ph4.x86_64
tdnf-3.0.0-5.ph4.x86_64
libsolv-devel-0.6.35-7.ph4.x86_64
```

```
root [ /build/bld-ph4 ]# ./bin/tdnf repoquery --whatrequires libsolv tdnf
tdnf-3.0.0-5.ph4.x86_64
tdnf-3.1.0-1.ph4.x86_64
tdnf-3.0.0-5.ph4.x86_64
```


tools/cli/main.c
```diff
@@ -455,6 +457,17 @@ TDNFCliInvokeRepoSync(
...
 }
 
+uint32_t
+TDNFCliInvokeRepoQuery(
+    PTDNF_CLI_CONTEXT pContext,
+    PTDNF_REPOQUERY_ARGS pRepoQueryArgs,
+    PTDNF_PKG_INFO *ppPkgInfos,
+    uint32_t *pdwCount
+    )
+{
```
sshedi, line 467: validate function args

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+    uint32_t dwError = 0;
+    PTDNF_PKG_INFO pPkgInfo = NULL;
+    PSolvQuery pQuery = NULL;
+    PSolvPackageList pPkgList = NULL;
+    int i;
+    uint32_t dwCount;
+    Id idDepends, idRequiresPre;
+
+    if(!pTdnf || !pTdnf->pSack || !pRepoqueryArgs ||
+       !ppPkgInfo)
```
sshedi: validate pdwCount

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+    uint32_t *pdwCount
+    )
+{
+    uint32_t dwError = 0;
+    PTDNF_PKG_INFO pPkgInfo = NULL;
+    PSolvQuery pQuery = NULL;
+    PSolvPackageList pPkgList = NULL;
+    int i;
+    uint32_t dwCount;
+    Id idDepends, idRequiresPre;
```
sshedi: nit: declare individually

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+
+        if (pRepoqueryArgs->nInstalled || pRepoqueryArgs->nDuplicates)
+        {
+            dwError = SolvAddSystemRepoFilter(pQuery);
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+    }
+
+    dwError = TDNFApplyScopeFilter(pQuery,
+        pRepoqueryArgs->nUpgrades ? SCOPE_UPGRADES : SCOPE_ALL);
```
sshedi: nit: indentation

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+    idRequiresPre = pool_str2id(pTdnf->pSack->pPool, TDNF_ID_REQUIRES_PRE, 1);
+
+    dwError = TDNFRefresh(pTdnf);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = SolvCreateQuery(pTdnf->pSack, &pQuery);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (!pRepoqueryArgs->nExtras)
+    {
```
sshedi, line 1397: we can use `TDNFApplyScopeFilter` inside this if as well, check

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+        dwError = SolvApplySinglePackageFilter(pQuery, pRepoqueryArgs->pszSpec);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = SolvApplyListQuery(pQuery);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (pRepoqueryArgs->nExtras)
+    {
+        dwError = SolvApplyExtrasFilter(pQuery);
```
sshedi, line 1428: Can we expand `TDNFApplyScopeFilter` with Extras and Duplicate filter?

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+    if (pRepoqueryArgs->pszFile)
+    {
+        dwError = SolvApplyFileProvidesFilter(pQuery, pRepoqueryArgs->pszFile);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = SolvGetQueryResult(pQuery, &pPkgList);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (pRepoqueryArgs->nChangeLogs)
```
sshedi: nit: instead of calling TDNFPopulatePkgInfoArray inside if, assign DETAIL_CHANGELOG to a variable and call function after if.

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+    }
+    else
+    {
+        dwError = TDNFPopulatePkgInfoArray(pTdnf->pSack, pPkgList, DETAIL_LIST, &pPkgInfo, &dwCount);
+    }
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (pRepoqueryArgs->nDepends)
+    {
+        dwError = TDNFPopulatePkgInfoArrayDependencies(
```
sshedi: Same here. first if & else if are almost identical. I think it can be simplified

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+    }
+    if(pPkgList)
+    {
+        SolvFreePackageList(pPkgList);
+    }
+    return dwError;
+error:
+    if(ppPkgInfo)
+    {
+      *ppPkgInfo = NULL;
```
sshedi: not needed. why assign to null at end just before leaving

client/api.c
```diff
@@ -1332,7 +1332,204 @@ TDNFRepoSync(
...
+    }
+    TDNFFreePackageInfo(pPkgInfo);
```
sshedi, line 1508: nit: null check before call, even though TDNFFreePackageInfo does it, checking before calling will save a function call.

client/defines.h
```diff
@@ -24,8 +24,11 @@
 
 typedef enum
 {
+    /* this should be a bitmask */
```
sshedi, line 27: Agree. PTDNF_REPOQUERY_ARGS structure can be simplified and all the subsequent function calls will become simpler and can be called based on set bits. I think using bitmask is the right way here

client/packageutils.c
```diff
@@ -184,6 +187,34 @@ TDNFPopulatePkgInfoArray(
...
+        {
+            dwError = SolvGetChangeLogFromId(
+                          pSack,
+                          dwPkgId,
+                          &pPkgInfo->pChangeLogEntries);
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+        else if (nDetail == DETAIL_SOURCEPKG)
+        {
+            dwError = SolvGetSourceFromId(
```
sshedi, line 200: declare all the newly introduced variables here. no need of full function scope for them

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+    )
+{
+    uint32_t dwError = 0;
+    const char* pszTmp = NULL;
+    Solvable *pSolv = NULL;
+    char *pszName = NULL;
+    char *pszArch = NULL;
+    char *pszEVR = NULL;
+
+    if(!pSack ||
```
sshedi: nit: All will fit in a single line

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+    }
+
+    pSolv = pool_id2solvable(pSack->pPool, dwPkgId);
+    if(!pSolv)
+    {
+        dwError = ERROR_TDNF_NO_DATA;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pszTmp = solvable_lookup_str(pSolv, SOLVABLE_SOURCENAME);
```
sshedi, line 2098: put SOLVABLE_SOURCENAME, SOLVABLE_SOURCEARCH & SOLVABLE_SOURCEEVR in a local structure with corresponding SOLVABLE_NAME & SOLVABLE_EVR like key value pair and loop.

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    *ppszName = pszName;
+    *ppszArch = pszArch;
+    *ppszEVR = pszEVR;
+cleanup:
+    return dwError;
+
+error:
+    if(ppszName)
```
sshedi: no need of nullyfying these pointers

client/packageutils.c
```diff
@@ -184,6 +187,34 @@ TDNFPopulatePkgInfoArray(
...
+                          &pszSrcEVR);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            dwError = TDNFAllocateStringPrintf(
+                          &pPkgInfo->pszSourcePkg,
+                          "%s-%s.%s",
+                          pszSrcName, pszSrcEVR, pszSrcArch);
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+        if ((uint32_t)dwPkgIndex < dwCount - 1)
```
sshedi: may be this was there earlier, make dwPkgIndex uint32_t and remove casting

client/packageutils.c
```diff
@@ -1005,3 +1036,97 @@ TDNFPopulatePkgInfos(
...
+TDNFPopulatePkgInfoArrayDependencies(
+    PSolvSack pSack,
+    PSolvPackageList pPkgList,
+    Id idKey,
+    PTDNF_PKG_INFO pPkgInfos
+    )
+{
+    uint32_t dwError = 0;
+    uint32_t dwCount = 0;
+    int dwPkgIndex = 0;
```
sshedi: use uint32_t

client/packageutils.c
```diff
@@ -1005,3 +1036,97 @@ TDNFPopulatePkgInfos(
...
+    dwError = SolvGetPackageListSize(pPkgList, &dwCount);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if(dwCount == 0)
+    {
+        dwError = ERROR_TDNF_NO_MATCH;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    for (dwPkgIndex = 0; (uint32_t)dwPkgIndex < dwCount; dwPkgIndex++)
```
sshedi: no need to cast

client/packageutils.c
```diff
@@ -184,6 +187,34 @@ TDNFPopulatePkgInfoArray(
...
+                          &pszSrcName,
+                          &pszSrcArch,
+                          &pszSrcEVR);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            dwError = TDNFAllocateStringPrintf(
+                          &pPkgInfo->pszSourcePkg,
+                          "%s-%s.%s",
+                          pszSrcName, pszSrcEVR, pszSrcArch);
+            BAIL_ON_TDNF_ERROR(dwError);
```
sshedi, line 212: bail check outisde of if else ladder

client/packageutils.c
```diff
@@ -1005,3 +1036,97 @@ TDNFPopulatePkgInfos(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = SolvGetDependenciesFromId(pSack, dwPkgId, idKey, &pPkgInfo->ppszDependencies);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+cleanup:
+    return dwError;
+
+error:
```
sshedi, line 1165: I think we need to stop doing this. Here cleanup and error are not doing anything special.
Just use `goto cleanup;` instead of `BAIL_ON_TDNF_ERROR`.

Reduces code size and personally I don't like backward flow of goto. gotos should be one directional

client/packageutils.c
```diff
@@ -1005,3 +1036,97 @@ TDNFPopulatePkgInfos(
...
+uint32_t
+TDNFPopulatePkgInfoArrayFileList(
+    PSolvSack pSack,
+    PSolvPackageList pPkgList,
+    PTDNF_PKG_INFO pPkgInfos
+    )
+{
+    uint32_t dwError = 0;
+    uint32_t dwCount = 0;
+    int dwPkgIndex = 0;
```
sshedi: uint32 please

client/packageutils.c
```diff
@@ -1005,3 +1036,97 @@ TDNFPopulatePkgInfos(
...
+
+cleanup:
+    return dwError;
+
+error:
+    goto cleanup;
+}
+
+uint32_t
+TDNFPopulatePkgInfoArrayFileList(
```
sshedi, line 1170: TDNFPopulatePkgInfoArrayFileList & TDNFPopulatePkgInfoArrayDependencies are almost identical.
Write one function and use a wrapper or something better. This is too much duplication.

client/packageutils.c
```diff
@@ -1005,3 +1036,97 @@ TDNFPopulatePkgInfos(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = SolvGetFileListFromId(pSack, dwPkgId, &pPkgInfo->ppszFileList);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+cleanup:
+    return dwError;
+
+error:
```
sshedi, line 1211: same here. We should start using linux kernel coding guidelines in tdnf with some compromising in tab spaces.
This doesn't look nice. Not blaming anyone here, it's time to move on

client/repo.c
```diff
@@ -1134,6 +1135,23 @@ TDNFEnsureRepoMDParts(
                       pRepoMD->pszUpdateInfo);
         BAIL_ON_TDNF_ERROR(dwError);
     }
+
+    if(!IsNullOrEmptyString(pRepoMDRel->pszOther))
+    {
+        dwError = TDNFAppendPath(
+                      pRepoMDRel->pszRepoCacheDir,
+                      pRepoMDRel->pszOther,
+                      &pRepoMD->pszOther);
```
sshedi, line 1148: normalize resultant path

client/repo.c
```diff
@@ -1268,6 +1286,16 @@ TDNFParseRepoMD(
         dwError = 0;
     }
     BAIL_ON_TDNF_ERROR(dwError);
+    dwError = TDNFFindRepoMDPart(
+                  pRepo,
+                  TDNF_REPOMD_TYPE_OTHER,
+                  &pRepoMD->pszOther);
+    BAIL_ON_TDNF_ERROR(dwError);
```
sshedi: check here, you are comparing error value after bailing out

client/repo.c
```diff
@@ -1134,6 +1135,23 @@ TDNFEnsureRepoMDParts(
                       pRepoMD->pszUpdateInfo);
         BAIL_ON_TDNF_ERROR(dwError);
     }
+
+    if(!IsNullOrEmptyString(pRepoMDRel->pszOther))
```
sshedi, line 1143: Can you please rewrite `TDNFDownloadRepoMDPart` function? once again, it's bloated with almost identical function calls

client/structs.h
```diff
@@ -104,6 +104,7 @@ typedef struct _TDNF_REPO_METADATA
     char *pszPrimary;
     char *pszFileLists;
     char *pszUpdateInfo;
+    char *pszOther;
```
sshedi, line 107: what is pszOther here? can we use something more meaingfull instead?

common/strings.c
```diff
@@ -499,3 +499,34 @@ TDNFStringEndsWith(
 error:
     goto cleanup;
 }
+
+static int
+_cmpstringp(const void *p1, const void *p2)
+{
+    return strcmp(* (char * const *) p1, * (char * const *) p2);
```
sshedi, line 530: i think this is looking a bit complex.
We can use `return strcmp(*(const char*) p1, *(const char*) p2);
strcmp accepts a constant character pointer, it's enough

common/strings.c
```diff
@@ -499,3 +499,34 @@ TDNFStringEndsWith(
 error:
     goto cleanup;
 }
+
+static int
+_cmpstringp(const void *p1, const void *p2)
```
sshedi, line 528: use str1, str2 instead of p1, p2

common/strings.c
```diff
@@ -499,3 +499,34 @@ TDNFStringEndsWith(
...
+    uint32_t dwError = 0;
+    int nCount;
+
+    if (IsNullOrEmptyString(ppszArray))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    for(nCount = 0; ppszArray[nCount]; nCount++);
```
sshedi, line 547: pass array size as function argument instead.

common/strings.c
```diff
@@ -499,3 +499,34 @@ TDNFStringEndsWith(
...
+
+    qsort(ppszArray, nCount, sizeof(char *), _cmpstringp);
+
+cleanup:
+    return dwError;
+error:
```
sshedi, line 553: just use one of cleanup or error

common/strings.c
```diff
@@ -499,3 +499,34 @@ TDNFStringEndsWith(
...
+
+    if (IsNullOrEmptyString(ppszArray))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    for(nCount = 0; ppszArray[nCount]; nCount++);
+
+    qsort(ppszArray, nCount, sizeof(char *), _cmpstringp);
```
sshedi, line 549: 1. nit: use qsort_r
2. nCount should be size_t not int

common/utils.c
```diff
@@ -228,6 +243,15 @@ TDNFFreePackageInfoContents(
         TDNF_SAFE_FREE_MEMORY(pPkgInfo->pszFormattedSize);
         TDNF_SAFE_FREE_MEMORY(pPkgInfo->pszRelease);
         TDNF_SAFE_FREE_MEMORY(pPkgInfo->pszLocation);
+        TDNFFreeStringArray(pPkgInfo->ppszDependencies);
```
sshedi: nit: null check before calling

common/utils.c
```diff
@@ -228,6 +243,15 @@ TDNFFreePackageInfoContents(
...
+        TDNFFreeStringArray(pPkgInfo->ppszFileList);
+        for (pEntry = pPkgInfo->pChangeLogEntries;
```
sshedi, line 248: nicely done. I think a while loop makes it more readable here

include/tdnftypes.h
```diff
@@ -168,6 +176,10 @@ typedef struct _TDNF_PKG_INFO
     char* pszFormattedSize;
     char* pszRelease;
     char* pszLocation;
+    char **ppszDependencies;
+    char **ppszFileList;
+    PTDNF_PKG_CHANGELOG_ENTRY pChangeLogEntries;
```
sshedi: it's a good practice to declare struct members in ascending or descending order. Helps in padding
For local variables in a function, gcc will take care of it

solv/prototypes.h
```diff
@@ -266,6 +272,35 @@ SolvGetNevraFromId(
     char **ppszEVR
     );
 
+uint32_t
+SolvGetDependenciesFromId(
+    PSolvSack pSack,
+    uint32_t dwPkgId,
+    Id idKey,
+    char ***pppszDependencies);
```
sshedi, line 280: Can we avoid using triple pointers? It should be feasible, please check

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+    if(!pSolv)
+    {
+        dwError = ERROR_TDNF_NO_DATA;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (idKey == pool_str2id(pSack->pPool, TDNF_ID_DEPENDS, 0))
+    {
+        queue_init(&queueDeps);
+        for (i = 0; i < (int)(sizeof(allDepKeys)/sizeof(Id)); i++)
```
sshedi: Use ARRAY_SIZE from include/tdnf-common-defines.h

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+    uint32_t dwPkgId,
+    Id idKey,
+    char ***pppszDependencies)
+{
+    uint32_t dwError = 0;
+    Solvable *pSolv = NULL;
+    Queue queueDeps = {0};
+    char **ppszDependencies = NULL;
+    const char *pszDep = NULL;
+    int nNumDeps, i, j;
```
sshedi, line 1940: use uint32_t, should be okay except for nNumDeps

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+    }
+
+    if (idKey == pool_str2id(pSack->pPool, TDNF_ID_DEPENDS, 0))
+    {
+        queue_init(&queueDeps);
+        for (i = 0; i < (int)(sizeof(allDepKeys)/sizeof(Id)); i++)
+        {
+            Queue queueTmp = {0};
+            solvable_lookup_deparray(pSolv, allDepKeys[i], &queueTmp, -1);
+
```
sshedi, line 1969: declare and use j in here

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
     TDNF_SAFE_FREE_MEMORY(pszEVR);
     goto cleanup;
 }
+
+uint32_t
+SolvGetDependenciesFromId(
+    PSolvSack pSack,
+    uint32_t dwPkgId,
+    Id idKey,
+    char ***pppszDependencies)
```
sshedi, line 1933: This triple pointer should be avoided to maximum extent, there is no need of triple pointer here I believe
Sadly, there are few in tdnf

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+{
+    uint32_t dwError = 0;
+    Solvable *pSolv = NULL;
+    Dataiterator di;
+    PTDNF_PKG_CHANGELOG_ENTRY pEntry = NULL;
+    PTDNF_PKG_CHANGELOG_ENTRY pEntryNext = NULL;
+    PTDNF_PKG_CHANGELOG_ENTRY pEntries = NULL;
+
+    if(!pSack ||
+       !ppEntries)
```
sshedi: in one line please

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+        if (pEntries == NULL)
+        {
+            pEntries = pEntry;
+        }
+        pEntry->pNext = pEntryNext;
+
+        pEntry->timeTime = (time_t)pool_lookup_num(
+            pSack->pPool, SOLVID_POS, SOLVABLE_CHANGELOG_TIME, 0);
+        dwError = TDNFAllocateString(
+            pool_lookup_str(pSack->pPool, SOLVID_POS, SOLVABLE_CHANGELOG_AUTHOR),
```
sshedi, line 2200: nit: fix indentation

solv/tdnfpackage.c
```diff
@@ -1924,3 +1924,302 @@ SolvGetNevraFromId(
...
+    dataiterator_free(&di);
+
+    *ppEntries = pEntry;
+
+cleanup:
+    return dwError;
+
+error:
+    for (pEntry = pEntries;
+         pEntry;
```
sshedi: any particular reason you split for loop into 3 lines? I find it hard to read
a while loop reads better here. Also null check before function call inside loop

solv/tdnfquery.c
```diff
@@ -1114,3 +1126,286 @@ SolvFindAllUpDownCandidates(
...
+                    queue_push(&queueFiltered, idPkg);
+                    break;
+                }
+            }
+            else
+            {
+                size_t k;
+                int nBreak = 0;
+
+                for (k = 0; k < sizeof(allDepKeys)/sizeof(Id); k++)
```
sshedi: use ARRAY_SIZE

solv/tdnfquery.c
```diff
@@ -1114,3 +1126,286 @@ SolvFindAllUpDownCandidates(
...
+                {
+                    if (solvable_matchesdep(pSolvable, allDepKeys[k], idDep, 0))
+                    {
+                        queue_push(&queueFiltered, idPkg);
+                        nBreak = 1;
+                        break;
+                    }
+                }
+                if (nBreak)
```
sshedi: no need of nBreak, check for value of k outside loop

solv/tdnfquery.c
```diff
@@ -1114,3 +1126,286 @@ SolvFindAllUpDownCandidates(
...
+
+        if (pSolvable->repo == pPool->installed)
+        {
+            int nFound = 0;
+            for (j = 0; j < pQuery->queueResult.count; j++)
+            {
+                if (i != j)
+                {
+                    Id idPkg2 = pQuery->queueResult.elements[j];
+                    Solvable *pSolvable2 = pool_id2solvable(pPool, idPkg2);
```
sshedi, line 1263: validate pSolvable2

solv/tdnfquery.c
```diff
@@ -1114,3 +1126,286 @@ SolvFindAllUpDownCandidates(
...
+    if(!pQuery)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_LIBSOLV_ERROR(dwError);
+    }
+
+    pPool = pQuery->pSack->pPool;
+
+    queue_init(&queueExtras);
+    for (i = 0; i < pQuery->queueResult.count; i++)
```
sshedi, line 1244: The inner and outer loops are almost identical, this can be further simplied

solv/tdnfquery.c
```diff
@@ -1114,3 +1126,286 @@ SolvFindAllUpDownCandidates(
...
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_LIBSOLV_ERROR(dwError);
+    }
+
+    pPool = pQuery->pSack->pPool;
+
+    queue_init(&queueDuplicates);
+
+    for (i = 0; i < pQuery->queueResult.count; i++)
```
sshedi, line 1318: same here, code duplication

solv/tdnfrepo.c
```diff
@@ -150,14 +150,52 @@ SolvLoadRepomdUpdateinfo(
...
+cleanup:
+    if(fp != NULL)
+    {
+        fclose(fp);
+    }
+    return dwError;
+
+error:
+    goto cleanup;
+
```
sshedi, line 187: empty line here and at line 189

solv/tdnfrepo.c
```diff
@@ -150,14 +150,52 @@ SolvLoadRepomdUpdateinfo(
...
+        BAIL_ON_TDNF_LIBSOLV_ERROR(dwError);
+    }
+cleanup:
+    if(fp != NULL)
+    {
+        fclose(fp);
+    }
+    return dwError;
+
+error:
```
sshedi, line 185: get rid of this error, it's not doing anything

We can improvise BAIL_ON_TDNF_ERROR macro a bit.
```
#define BAIL_ON_TDNF_ERROR(err, label) do { if (err) goto label; while (0)
```

solv/tdnfrepo.c
```diff
@@ -150,14 +150,52 @@ SolvLoadRepomdUpdateinfo(
...
 uint32_t
 SolvReadYumRepo(
     Repo *pRepo,
     const char *pszRepoName,
     const char *pszRepomd,
     const char *pszPrimary,
     const char *pszFilelists,
-    const char *pszUpdateinfo
+    const char *pszUpdateinfo,
+    const char *pszOther
```
sshedi, line 198: validate if applicable

tools/cli/lib/api.c
```diff
@@ -485,6 +485,125 @@ TDNFCliRepoSyncCommand(
...
+        {
+            for (j = 0; pPkgInfo->ppszFileList[j]; j++);
+            nCount +=j;
+        }
+        else if (pPkgInfo->pChangeLogEntries)
+        {
+            PTDNF_PKG_CHANGELOG_ENTRY pEntry;
+            for (pEntry = pPkgInfo->pChangeLogEntries; pEntry; pEntry = pEntry->pNext)
+            {
+                char szTime[20];
```
sshedi: initialize array to 0 and validate ret val of strftime

tools/cli/lib/api.c
```diff
@@ -485,6 +485,125 @@ TDNFCliRepoSyncCommand(
...
+        dwError = ERROR_TDNF_CLI_INVALID_ARGUMENT;
+        BAIL_ON_CLI_ERROR(dwError);
+    }
+
+    dwError = TDNFCliParseRepoQueryArgs(pCmdArgs, &pRepoqueryArgs);
+    BAIL_ON_CLI_ERROR(dwError);
+
+    dwError = pContext->pFnRepoQuery(pContext, pRepoqueryArgs, &pPkgInfos, &dwCount);
+    BAIL_ON_CLI_ERROR(dwError);
+    for (i = 0; i < (int)dwCount; i++)
```
sshedi, line 513: can't we declare i,j as uint32_t?

tools/cli/lib/api.c
```diff
@@ -485,6 +485,125 @@ TDNFCliRepoSyncCommand(
...
+                pPkgInfo->pszName,
+                pPkgInfo->pszVersion,
+                pPkgInfo->pszRelease,
+                pPkgInfo->pszArch);
+        }
+    }
+
+    if (nCount > 0)
+    {
+        k = 0;
```
sshedi: declare k here

tools/cli/lib/api.c
```diff
@@ -485,6 +485,125 @@ TDNFCliRepoSyncCommand(
...
+
+        dwError = TDNFAllocateMemory(nCount + 1, sizeof(char *), (void**)&ppszLines);
+        BAIL_ON_CLI_ERROR(dwError);
+        for (i = 0; i < (int)dwCount; i++)
```
sshedi: you can do `for(k =0, i = 0; ...)`

tools/cli/lib/parserepoqueryargs.c
```diff
@@ -0,0 +1,205 @@
...
+                            dwError = TDNFSplitStringToArray(pSetOpt->pszOptValue,
+                                ",",
+                                &pRepoqueryArgs->pppszWhatKeys[i]);
+                            BAIL_ON_CLI_ERROR(dwError);
+                            nFound = 1;
+                        }
+                    }
+                }
+            }
+        }
```
sshedi: reduce the indentation levels here, ideally it should not exceed 3 levels

tools/cli/lib/parserepoqueryargs.c
```diff
@@ -0,0 +1,205 @@
...
+        TDNF_CLI_SAFE_FREE_STRINGARRAY(pRepoqueryArgs->ppszWhatDepends);
+        if (pRepoqueryArgs->pppszWhatKeys)
+        {
+            int i;
+            for (i = 0; i < REPOQUERY_KEY_COUNT; i++)
+            {
+                TDNF_CLI_SAFE_FREE_STRINGARRAY(pRepoqueryArgs->pppszWhatKeys[i]);
+            }
+        }
+        TDNFFreeMemory(pRepoqueryArgs->pppszWhatKeys);
```
sshedi: Use SAFE_FREE macro

sshedi (changes requested):
Thanks for the feature. I can imagine the effort you have put into this.

Done with round 1 of review. One common comment is, code redundancy has to be taken care of else things will get out of control.

Also, run pytests with valgrind enabled.

tools/cli/main.c
oliverkurth: That's redundant. If you look at the other functions in that file you will see that they are also just one liners passing args without change to another function. Arg checking should take place there.

Normally I would agree because it's a common pattern in tdnf to check arguments over and over again. But here in this file, the pattern is not to, if we just pass the args on.

client/api.c
oliverkurth: Done

client/api.c
oliverkurth: Done

client/api.c
oliverkurth: This is error handling, I don't think we need to optimize here.

solv/tdnfpackage.c
oliverkurth: Unfortunately `SOLVABLE_SOURCENAME`/`SOLVABLE_SOURCEEVR` and `SOLVABLE_SOURCEARCH` are handled differently so we can only loop over the first two. I don't think it would make nicer.

client/packageutils.c
oliverkurth: I'd prefer those right after the function call, unless those function calls are all the same pattern. The `TDNFAllocateStringPrintf()` breaks the pattern.

client/packageutils.c
oliverkurth: I strongly disagree here. This is a common pattern throughout tdnf code, like a "poor man's exception handling", and this uses gotos. I agree that jumping backwards make code harder to understand in most cases, but not here where it is used in a consistent way.

client/repo.c
oliverkurth: I don't understand your meaning here.

client/structs.h
oliverkurth: I didn't come up with that name. It's part of the xml spec for repo metadata, see for example https://packages.vmware.com/photon/4.0/photon_4.0_x86_64/repodata/repomd.xml:
```
<repomd xmlns="http://linux.duke.edu/metadata/repo" xmlns:rpm="http://linux.duke.edu/metadata/rpm">
<revision>1628135114</revision>
<data type="primary">
...
</data>
<data type="filelists">
...
</data>
<data type="other">
```

common/strings.c
oliverkurth: That does not work. This was copied straight from the man page, and it has a comment explaining it. And it's exactly the same use case as in the example - we are sorting pointers to chars, but qsort passes pointers to the object to be sorted to the comparison function. Therefore we need to dereference the pointers, before we pass them to `strcmp()`.
```
           /* The actual arguments to this function are "pointers to
              pointers to char", but strcmp(3) arguments are "pointers
              to char", hence the following cast plus dereference */
```

common/strings.c
oliverkurth: I don't quite remember why I wrote this function, I could have called qsort directly, but one point maybe was so the caller doesn't need to count.

common/strings.c
oliverkurth: Why use `qsort_r()`? We don't need a 3rd argument.

common/utils.c
oliverkurth: That's done in the function. Same argument as before. Actually, we can use `TDNF_SAFE_FREE_STRINGARRAY` which also checks the pointer, but at least it doesn't clutter the code.

common/utils.c
oliverkurth: I disagree. I don't like using `while` when `for` fits perfectly.

include/tdnftypes.h
oliverkurth: I have no idea what you mean. Can you clarify?

solv/prototypes.h
oliverkurth: It's for returning something in the argument list. It's not that confusing when you think abut it like a reference.

It would be different if the tdnf coding style would allow return a value with `return` instead of only returning errors that way.

solv/tdnfpackage.c
oliverkurth: Oh, I missed that exists.

solv/tdnfpackage.c
oliverkurth: No. See https://github.com/openSUSE/libsolv/blob/master/src/queue.h for the type of `queueTmp.count`, it's `int`.

solv/tdnfpackage.c
oliverkurth: Joined the lines, it does not really matter IMHO.
Disagree about the `while`. `for` is good here, just as above.
`pEntry` cannot be `NULL` inside the loop. It's impossible.

solv/tdnfquery.c
oliverkurth: Good catch!

client/repo.c
sshedi:
```
dwError = fun();
BAIL_ON_TDNF_ERROR(dwError); ---> this exits if dwError is not zero, so the next if condition won't get executed ever
if (dwError == ...) { ... }
```

common/strings.c
sshedi: just a defensive approach

include/tdnftypes.h
sshedi: structure members in ascending or descending order of their size.
For example:
```
struct A {
   int a;
   char b[10];
   char c;
};
```
should be 
```
struct A {
   char c;
   int a;
   char b[10];
};
```

solv/prototypes.h
sshedi: I think it's okay. Just for the sake of returning an integer error value, we should not compromise readability and have easily maintainable code

sshedi: I think we should improve our coding guidelines. Because of guidelines we are ending up with triple pointers which is unnecessary, we need to fix the guidelines first. It should be flexible and open for new approaches.

`If a function has the possibility to fail, it must return a 32 bit error code` - this doesn't make any sense. If we want error code for printing error messages, then we can display the error in failed function itself return pointer, integer, bool or whatever we want from the function.

sshedi: `Functions must have a single exit point` - usually such guidelines are followed in MISRA C standards. TDNF is not an embedded system application. We definitely need more flexibility. We can discus and have a better coding guidelines for tdnf offline.

client/repo.c
oliverkurth: Oh got it, thanks! I'll remove the bailout before the check.

tools/cli/lib/parserepoqueryargs.c
oliverkurth: I don't think putting the inner levels into a function makes it any more readable. I will add comments to the closing brackets to make it easier to understand.

client/repo.c
oliverkurth: In another change. The whole thing is a mess.

common/strings.c
oliverkurth: It doesn't gain anything.

include/tdnftypes.h
oliverkurth: Okay. I moved `pChangeLogEntries` to the bottom, but before `pNext` because that should stand out.

solv/prototypes.h
oliverkurth: Let's discuss this internally. I have a few ideas along these lines.

solv/tdnfrepo.c
oliverkurth: That's not making it more readable.

client/api.c
oliverkurth: No, unfortunately not. The two sets of filters (scope and extras/duplicates) work differently. The former use a query filter that will be executed by `SolvApplyListQuery()`. The latter filters through the results from that function. So the first set needs to be before `SolvApplyListQuery()` and the second after.

It's all a little messy and should be rewritten, but it's complex. It took me some time to figure this out, and libsolv is badly documented.

Also, the "extras" and "duplicates" options should also be available to the `list` and `info` commands. This should be in another change though.

client/defines.h
oliverkurth: I actually meant this for `TDNFPopulatePkgInfoArray()` to fill in details for bits that are set in a bitmask. Not for PTDNF_REPOQUERY_ARGS.

This should be in another change since it affects the `list` and `info` commands as well.

client/packageutils.c
oliverkurth: I think we can make it simpler if we use the bitmask idea for `TDNF_PKG_DETAIL` as discussed above, and rewrite this.

client/repo.c
oliverkurth: As discussed elsewhere, we should use a new function that constructs paths.

solv/tdnfpackage.c
oliverkurth: Again, think of it as reference to an array of string pointers.

client/api.c
oliverkurth: Done, using ternary operators.

client/api.c
oliverkurth: I found no easy way, but while pondering about it I found that the logic can be made much simpler because there is only one dependency key allowed. It's much cleaner now.

client/api.c
oliverkurth: Yes, right. Done.

solv/tdnfquery.c
oliverkurth: I don't know how to make this simpler.

solv/tdnfrepo.c
oliverkurth: Done.

tools/cli/lib/api.c
oliverkurth: Not really needed. Lets keep it together witth `i` and `j`.

tools/cli/lib/api.c
oliverkurth: We could, but I don't like it.

client/api.c
```diff
@@ -1345,6 +1345,161 @@ TDNFRepoSync(
...
+    PTDNF_PKG_INFO* ppPkgInfo,
+    uint32_t *pdwCount
+    )
+{
+    uint32_t dwError = 0;
+    PTDNF_PKG_INFO pPkgInfo = NULL;
+    PSolvQuery pQuery = NULL;
+    PSolvPackageList pPkgList = NULL;
+    int nDetail;
+    uint32_t dwCount;
```
tapakund: Better to initialise dwCount to 0.

client/api.c
```diff
@@ -1345,6 +1345,161 @@ TDNFRepoSync(
...
+    TDNF_SCOPE nScope = SCOPE_ALL;
+
+    if(!pTdnf || !pTdnf->pSack || !pRepoqueryArgs ||
+       !ppPkgInfo || !pdwCount)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    /* check if args make sense */
```
tapakund: pls add some more comments, would help someone reading this easy.

client/packageutils.c
```diff
@@ -184,6 +184,38 @@ TDNFPopulatePkgInfoArray(
...
+            dwError = TDNFAllocateStringPrintf(
+                          &pPkgInfo->pszSourcePkg,
+                          "%s-%s.%s",
+                          pszSrcName, pszSrcEVR, pszSrcArch);
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+        if (dwPkgIndex < dwCount - 1)
+        {
+            pPkgInfo->pNext = &pPkgInfos[dwPkgIndex+1];
+        }
```
tapakund, line 217: we need to free pszSrcName, pszSrcArch and pszSrcEVR

oliverkurth: Good catch. Fixed.

client/api.c
oliverkurth: Done

client/api.c
oliverkurth: Done

tapakund (approved):
LGTM

PR #249, opened 2021-07-20
tdnf-python: fix regression in tdnf.repolist()

sshedi:
repo->pszMetaLink can be null and PyBytes_FromString doesn't validate
null pointers which was resulting in a seg fault.

Other changes:

1. Added dist flag support to spec (which helps in reinstalling tdnf if
   needed)
2. Fixed a typo in python/README.md
3. Improvements to pytests/conftest.py
4. Added config file parameter support to tdnf-python

### DIFF
--- a/ci/docker-entrypoint.sh
+++ b/ci/docker-entrypoint.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 rm -rf build
-mkdir build && cd build
-cmake .. && make && make python
-make check
+mkdir -p build
+cd build || exit 1
+cmake .. && make -j32 && make python -j32
+make check -j32
--- a/pytests/conftest.py
+++ b/pytests/conftest.py
@@ -1,91 +1,80 @@
 #
-# Copyright (C) 2019 VMware, Inc. All Rights Reserved.
+# Copyright (C) 2019-2021 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the GNU General Public License v2 (the "License");
 # you may not use this file except in compliance with the License. The terms
 # of the License are located in the COPYING file of this distribution.
 #
-#   Author: Siddharth Chandrasekaran <csiddharth@vmware.com>
 
-import pytest
-import json
-import subprocess
 import os
 import re
-import errno
-import shutil
 import ssl
+import sys
+import time
+import json
+import errno
+import socket
+import shutil
+import pytest
+import atexit
 import requests
+import subprocess
 import configparser
 import distutils.spawn
-from pprint import pprint
+from OpenSSL import crypto
 from urllib.parse import urlparse
-from OpenSSL import crypto, SSL
+from multiprocessing import Process
 from http.server import SimpleHTTPRequestHandler, HTTPServer
-import socketserver
-import threading
-import ssl
 
-class TestRepoServer(threading.Thread):
 
-    def __init__(self, root, port=8080, interface="", enable_https=False):
-        super().__init__()
-        self.daemon = True
-        self.port = port
-        self.root = root
-        self.addr = (interface, port)
-        self.enable_https = enable_https
+def StopTestRepoServer(server):
+    server.terminate()
+    server.join()
 
-    def make_cert(self):
-        if os.path.exists("cert.pem") and os.path.exists("key.pem"):
-            self.keyfile = os.path.join(os.getcwd(), 'key.pem')
-            self.certfile = os.path.join(os.getcwd(), 'cert.pem')
-            return
+
+def TestRepoServer(root, port=8080, interface='', enable_https=False):
+    addr = (interface, port)
+    enable_https = enable_https
+
+    def dump_cert(fn, cert):
+        with open(fn, 'wt') as f:
+            tmp = crypto.dump_certificate(crypto.FILETYPE_PEM, cert)
+            f.write(tmp.decode('utf-8'))
+
+    os.chdir(root)
+
+    if enable_https:
         k = crypto.PKey()
         k.generate_key(crypto.TYPE_RSA, 2048)
         cert = crypto.X509()
-        cert.get_subject().C = "IN"
-        cert.get_subject().ST = "Karnataka"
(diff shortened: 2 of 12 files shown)


sshedi:
@tapakund @oliverkurth I will add tests later. I'm still having a hard time to add tests for this.
For the time being, let's get these changes merged first.

pytests/conftest.py
```diff
@@ -53,39 +57,31 @@ def make_cert(self):
...
-                                                    server_side = True)
-            self.httpd.serve_forever()
-        finally:
-            self.httpd.server_close()
-
-    def join(self):
-        self.httpd.shutdown()
-        super().join()
+        keyfile = os.path.join(os.getcwd(), 'key.pem')
+        certfile = os.path.join(os.getcwd(), 'cert.pem')
```
oliverkurth: This is odd - we set `keyfile` and `certfile` in line 45/46, then open the files *without* using the variables, then assign the same values again. I know this isn't new, but should be rewritten. Also, assigning those values to local variables in the if clause (instead of attributes to `self`) in lines 45/46 doesn't seem to make sense if we are returning.

pytests/conftest.py
```diff
@@ -1,45 +1,49 @@
...
+def StopTestRepoServer(server):
+    server.terminate()
+    server.join()
 
-    def make_cert(self):
+def TestRepoServer(root, port=8080, interface="", enable_https=False):
+    port = port
+    root = root
+    addr = (interface, port)
+    enable_https = enable_https
```
oliverkurth, line 37: Remove the lines that are not doing anything (`foo=foo`).

pytests/conftest.py
```diff
@@ -280,4 +274,27 @@ def _run(self, cmd, retvalonly=False, cwd=None):
...
+    server = Process(target=TestRepoServer,
+                     args=(test_utils.config['repo_path'], ))
+    server.start()
+
+    atexit.register(StopTestRepoServer, server)
+
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    retry = 0
+    while retry < 10:
+        retry += 1
```
oliverkurth: Nit: Consider using a `for` loop: `for retry in range(1, 10):`. Just be careful testing the value in line 296.

pytests/tests/test_tdnf_python.py
```diff
@@ -0,0 +1,203 @@
...
+
+    global sglpkgname
+    sglpkgname = utils.config['sglversion_pkgname']
+
+    global config
+    config = utils.config['repo_path'] + '/tdnf.conf'
+
+    path = utils.config['build_dir']
+    path += '/python/build/lib.linux-' + os.uname().machine + '-'
+    path += str(sys.version_info.major)+'.'+str(sys.version_info.minor)
```
oliverkurth: Nit: you can use `format()` here.

pytests/tests/test_tdnf_python.py
```diff
@@ -0,0 +1,203 @@
...
+
+def teardown_test(utils):
+    utils.run(['tdnf', 'erase', '-y', sglpkgname])
+    utils.run(['tdnf', 'erase', '-y', mulpkgname])
+    assert (utils.check_package(sglpkgname) == False)
+    assert (utils.check_package(mulpkgname) == False)
+
+
+def tdnf_py_erase(pkgs, quiet=False, refresh=False):
+    pkgs = [x for x in pkgs if x]
```
oliverkurth, line 51: Why check for unset entries? Can this happen?

sshedi: Yes, for some weird reason the pkg list has entries like [b' ', 'tdnf-test-one']

pytests/conftest.py
```diff
@@ -280,4 +265,25 @@ def _run(self, cmd, retvalonly=False, cwd=None):
...
+        result = sock.connect_ex(('localhost', 8080))
+        if not result:
+            print("Server is up and running")
+            sock.close()
+            break
+        else:
+            print("Server is not running, retrying ...")
+            time.sleep(1)
+
+    if (retry == 10):
```
oliverkurth: This won't tell you if it succeeded or not. It may have succeeded with the last attempt, or not. Better test `result`.

oliverkurth (commented):
I am not familiar with the tdnfmodule stuff yet - I need to catch up on that.

pytests/conftest.py
sshedi: Nice, ty. Done.

ci/docker-entrypoint.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 rm -rf build
-mkdir build && cd build
-cmake .. && make && make python
-make check
+mkdir -p build
+cd build || exit 1
+cmake .. && make -j32 && make python -j32
+make check -j32
```
oliverkurth, line 7: I like that the build is much faster with `-j32`, but I don't think it has an effect for `make check`.

oliverkurth (approved):
Looks good to me.

I would reword the commit message because it does more that what it says, or split it into multiple commits per issue addressed - this would make it easier to revert in case something breaks.

I am not happy with the python API - we shouldn't need to give options to every method call, instead this should be set once in a context class. But that's another topic.

ci/docker-entrypoint.sh
sshedi: I know. I just kept it anyway, assuming it might do something under the hood :)
It's not causing any issues, I think it's okay to keep it.

sshedi:
> Looks good to me.
> 
> I would reword the commit message because it does more that what it says, or split it into multiple commits per issue addressed - this would make it easier to revert in case something breaks.
> 
> I am not happy with the python API - we shouldn't need to give options to every method call, instead this should be set once in a context class. But that's another topic.

I changed my commit message a bit. Thanks.
And yes, python api framework is not looking good, I just made it work with existing code.
If any real requirement comes up we can refactor it.

PR #250, opened 2021-07-26
Fix configreader key reading logic

sshedi:

### DIFF
--- a/pytests/tests/test_cfgreader.py
+++ b/pytests/tests/test_cfgreader.py
@@ -1,0 +1,71 @@
+#
+# Copyright (C) 2019-2021 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the GNU General Public License v2 (the "License");
+# you may not use this file except in compliance with the License. The terms
+# of the License are located in the COPYING file of this distribution.
+#
+
+import os
+import pytest
+
+NEW_REPO_FN = 'test.repo'
+ORIG_REPO_FN = 'photon-test.repo'
+
+
+def create_repoconf(fn):
+    content = '''
+[cfg-test]
+name=    Test Repo
+baseurl = http://localhost:8080/photon-test
+enabled      =1
+gpgcheck=0
+'''
+    with open(fn, 'w') as f:
+        f.write(content)
+
+
+@pytest.fixture(scope='module', autouse=True)
+def setup_test(utils):
+    global NEW_REPO_FN
+    global ORIG_REPO_FN
+
+    NEW_REPO_FN = os.path.join(utils.config['repo_path'], 'yum.repos.d',
+                               NEW_REPO_FN)
+
+    ORIG_REPO_FN = os.path.join(utils.config['repo_path'], 'yum.repos.d',
+                                ORIG_REPO_FN)
+
+    create_repoconf(NEW_REPO_FN)
+
+    # take backup of original repo file
+    os.rename(ORIG_REPO_FN, ORIG_REPO_FN+'.bak')
+
+    yield
+
+    teardown_test(utils)
+
+
+def teardown_test(utils):
+    os.remove(NEW_REPO_FN)
+    os.rename(ORIG_REPO_FN+'.bak', ORIG_REPO_FN)
+    ret = utils.run(['tdnf', 'makecache'])
+    assert(ret['retval'] == 0)
+
+
+def test_makecache(utils):
+    ret = utils.run(['tdnf', 'makecache'])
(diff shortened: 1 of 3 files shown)


pytests/repo/setup-repo.sh
```diff
@@ -75,11 +75,11 @@ gpg --batch --passphrase-fd 0 \
 cat << EOF > ${TEST_REPO_DIR}/yum.repos.d/photon-test.repo
 [photon-test]
 name=basic
-baseurl=http://localhost:8080/photon-test
+baseurl = http://localhost:8080/photon-test
```
sshedi: Looks ugly but needed for testing

sshedi:
@oliverkurth Please review and merge the changes

pytests/repo/setup-repo.sh
oliverkurth: Maybe add a test case with an ugly repo conf? This would isolate the issue. If this goes wrong lots (all?) other tests will fail, and that will be confusing.

oliverkurth (commented):
I wonder if we should use a library to parse the config flles, like https://github.com/benhoyt/inih .

pytests/repo/setup-repo.sh
sshedi: Good suggestion. It has happened already in this test I guess.

sshedi:
> I wonder if we should use a library to parse the config flles, like https://github.com/benhoyt/inih .

We can do that. I'm not sure if we need get approval for it. If approval is needed, I say we should stick to existing method :)

pytests/tests/test_cfgreader.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def teardown_test(utils):
+    os.remove(NEW_REPO_FN)
+    os.rename(ORIG_REPO_FN+'.bak', ORIG_REPO_FN)
+    ret = utils.run(['tdnf', 'makecache'])
+    assert(ret['retval'] == 0)
+
+
+def test_makecache(utils):
```
sshedi, line 56: I think these two tests are enough to give confidence that config reader is functioning normal.

pytests/tests/test_cfgreader.py
```diff
@@ -0,0 +1,71 @@
...
+gpgcheck=0
+'''
+    with open(fn, 'w') as f:
+        f.write(content)
+
+
+@pytest.fixture(scope='module', autouse=True)
+def setup_test(utils):
+    global NEW_REPO_FN
+    global ORIG_REPO_FN
```
oliverkurth, line 31: No need to make them global explicitly, they already are.

pytests/tests/test_cfgreader.py
```diff
@@ -0,0 +1,71 @@
...
+    NEW_REPO_FN = os.path.join(utils.config['repo_path'], 'yum.repos.d',
+                               NEW_REPO_FN)
+
+    ORIG_REPO_FN = os.path.join(utils.config['repo_path'], 'yum.repos.d',
+                                ORIG_REPO_FN)
+
+    create_repoconf(NEW_REPO_FN)
+
+    # take backup of original repo file
+    os.rename(ORIG_REPO_FN, ORIG_REPO_FN+'.bak')
```
oliverkurth, line 42: Why not just use the `-c`/`--config` option instead of moving the config away?

oliverkurth (changes requested):

pytests/tests/test_cfgreader.py
sshedi: This is not config, this is repo file. If I need to use a different config, I need to create both config and repo file. This is easier.

pytests/tests/test_cfgreader.py
sshedi: It doesn't work, we will hit the below error:
`UnboundLocalError: local variable 'NEW_REPO_FN' referenced before assignment`

oliverkurth: Hm, weird.

pytests/tests/test_cfgreader.py
oliverkurth: Oh sorry, I was confused. But then - we could do the test with the config file too.

oliverkurth (approved):

pytests/tests/test_cfgreader.py
sshedi: We can do it but it will be redundant right? The goal here is giving some tricky but valid inputs to config reader. This is doing exactly that.

PR #252, opened 2021-07-30
Removing tmp directory in case of shasum mismatch for metalink.

ntsbtz:


tapakund (changes requested):
I think the fix should be in this error section: https://github.com/vmware/tdnf/blob/cf548d4ba233f0c969ce19358aadea9b83a08782/client/repo.c#L1005

If TDNFGetRepoMD failed, we should clean up the tmp and exit.

Also, instead of using 'rm -rf' can you try to use already existing functionality 'TDNFRemoveTmpRepodata' ?

tapakund (commented):
can you try to remove this https://github.com/vmware/tdnf/blob/d02897d4817eb3551b475dac08d2149ad2ef69b3/client/repo.c#L965 , as this is doing the tmp directory clean up which is already handled in the cleanup now.

Also, as we are using TDNFRemoveTmpRepodata, I see pszFilePath is not used, can you cleanup that as well.  https://github.com/vmware/tdnf/blob/96ea582a9bb0e0097fbcc676045f549344ae0397/client/repoutils.c#L314

ntsbtz:
> can you try to remove this
> 
> https://github.com/vmware/tdnf/blob/d02897d4817eb3551b475dac08d2149ad2ef69b3/client/repo.c#L965
> 
> , as this is doing the tmp directory clean up which is already handled in the cleanup now.
> Also, as we are using TDNFRemoveTmpRepodata, I see pszFilePath is not used, can you cleanup that as well.
> 
> https://github.com/vmware/tdnf/blob/96ea582a9bb0e0097fbcc676045f549344ae0397/client/repoutils.c#L314

Resolved now.

client/remoterepo.c
```diff
@@ -274,7 +274,7 @@ TDNFCheckRepoMDFileHashFromMetalink(
     TDNF_METALINK_FILE *ml_file
     )
 {
-
+
```
tapakund: remove extra spaces

client/repo.c
```diff
@@ -983,6 +978,10 @@ TDNFGetRepoMD(
     *ppRepoMD = pRepoMD;
 
 cleanup:
+    if (!IsNullOrEmptyString(pszTmpRepoDataDir))
+    {
+        TDNFRemoveTmpRepodata(pszTmpRepoDataDir);
```
tapakund: I see TDNFRemoveTmpRepodata is only used here and we are not using the error code returned.Shall we modify TDNFRemoveTmpRepodata function?

tapakund (commented):
Overall Looks good, some minor comment and suggestions. 

Also, would you mind adding some tests on this scenario.

client/remoterepo.c
ntsbtz: done

client/repo.c
keerthanakalyan: We can add a return error check for TDNFRemoveTmpRepodata along with dwError == ERROR_TDNF_CHECKSUM_VALIDATION_FAILED and add a pr_crit print saying "Downloaded repoMD file shasum mismatch, please remove the tmp directory path". This will alert the user to remove malicious repomd files manually, if tmp directory removal is failed from tdnf.

ntsbtz: Now errno is getting set while nftw is getting failed so should we check dwError is not 0 and print saying above or instead if dwError = errno replaced with ERROR_TDNF_CHECKSUM_VALIDATION_FAILED ?

keerthanakalyan: It should be something like below:
```
if (!IsNullOrEmptyString(pszTmpRepoDataDir))
{
        if (TDNFRemoveTmpRepodata(pszTmpRepoDataDir) && dwError == ERROR_TDNF_CHECKSUM_VALIDATION_FAILED )
        {
                pr_crit("Downloaded repomd shasum mismatch, failed to delete %s file.  please delete.", pszTmpRepoDataDir)
         }
}
```

ntsbtz: done

client/repo.c
```diff
@@ -983,6 +978,14 @@ TDNFGetRepoMD(
     *ppRepoMD = pRepoMD;
 
 cleanup:
+    if (!IsNullOrEmptyString(pszTmpRepoDataDir))
+    {
+        if((TDNFRemoveTmpRepodata(pszTmpRepoDataDir)) &&
+           (dwError == ERROR_TDNF_CHECKSUM_VALIDATION_FAILED))
+	{
+	    pr_crit("Downloaded repomd shasum mismatch, failed to delete %s file.  please delete.", pszTmpRepoDataDir);
```
tapakund: can we replace delete with remove?

tapakund (commented):
LGTM, minor comment.

client/repo.c
ntsbtz: done

tapakund (approved):
LGTM

tapakund:
cc @oliverkurth @keerthanakalyan @sshedi

Issue #254, opened 2021-08-05
reposync does not work in the root directory unless a path is specified


oliverkurth:
```
root [ /build ]# cd /
root [ / ]# tdnf reposync --disablerepo=* --enablerepo=photon-updates --newest-only --delete
Error(1524) : URL is invalid.

root [ / ]# cd

root [ ~ ]# tdnf reposync --disablerepo=* --enablerepo=photon-updates --newest-only --delete
ImageMagick                              96389   100%
ImageMagick-c++                         212609   100%
ImageMagick-c++-devel                    92138   100%
^C

root [ ~ ]# cd /

root [ / ]# tdnf reposync --disablerepo=* --enablerepo=photon-updates --newest-only --delete --download-path=/root/photon-updates/
ImageMagick                              96389   100%
ImageMagick-c++                         212609   100%
ImageMagick-c++-devel                    92138   100%
ImageMagick-devel                        88735   100%
^C
```

PR #255, opened 2021-08-05
Fix reposync with '/' as root dir (fixes issue #254)

oliverkurth:
Also adding tests.


oliverkurth:
The version bump is already committed and pushed, idk how to remove it from the PR. Also, I don't know why the bot didn't recognize me as an employee - I just signed it again, not sure if it works now.

sshedi (approved):
Changes look good to me. But the worrisome factor here is, we can't keep doing this for every path string.

We have `TDNFNormalizePath` function and should make use of it.

We need to have a wrapper function for path allocations instead of using `TDNFAllocateString` directly for paths and call `TDNFNormalizePath` after we get the path.

I have also observed that if baseurl has a `/` at the end, tdnf won't work as expected.

Oliver, can you please make the necessary changes in this PR itself? I mean the wrapper function for paths. Slowly we can use the same everywhere.

I'm approving this change for now but it will be nice to solve it once and for all.

oliverkurth:
> Changes look good to me. But the worrisome factor here is, we can't keep doing this for every path string.
> 
> We have `TDNFNormalizePath` function and should make use of it.

I agree that we need refactor creating paths, and I was thinking about exactly that when I made this change. But `TDNFNormalizePath()`  does something different - it creates the "real" path (it uses `realpath()`), following symlinks. What we need is a function that creates a path disregarding symlinks, similar to what `os.path.join()` in python does.

> I have also observed that if baseurl has a `/` at the end, tdnf won't work as expected.

Can you give an example?

> Oliver, can you please make the necessary changes in this PR itself? I mean the wrapper function for paths. Slowly we can use the same everywhere.

I prefer to do that in a separate change, trying to address all cases. I want this change in the next stable release as a bug fix, and bigger changes in the dev branch.

> I'm approving this change for now but it will be nice to solve it once and for all.

sshedi:
Try using below setting in photon-updates.repo
`baseurl=https://packages.vmware.com/photon/$releasever/photon_updates_$releasever_$basearch/`

oliverkurth:
> Try using below setting in photon-updates.repo
> `baseurl=https://packages.vmware.com/photon/$releasever/photon_updates_$releasever_$basearch/`

I tried that and see no issues. I did `tdnf makecache`, `update`, `install` `reposync`. Can you file a separate issue with complete repo steps?

Issue #256, opened 2021-08-16
pytest fails on arm


oliverkurth:
Arch is assumed to be `x86_64` for tests, so they fail on arm. This is just the first failure:

```
==================================================================== FAILURES ====================================================================
________________________________________________________ test_check_local_with_local_rpm _________________________________________________________

utils = <pytests.conftest.TestUtils object at 0xffffb76ffee0>

    def test_check_local_with_local_rpm(utils):
        with tempfile.TemporaryDirectory() as tmpdir:
            dest = os.path.join(tmpdir, 'test.rpm')
            src = os.path.join(utils.config['repo_path'], 'build/RPMS/x86_64/tdnf-test-two-1.0.1-1.x86_64.rpm')
>           shutil.copyfile(src, dest)

tests/test_check_local.py:45: 
...
>           with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
E           FileNotFoundError: [Errno 2] No such file or directory: '/build/build/pytests/repo/build/RPMS/x86_64/tdnf-test-two-1.0.1-1.x86_64.rpm'

/usr/lib/python3.9/shutil.py:264: FileNotFoundError
```

oliverkurth:
Fixed.

PR #257, opened 2021-08-16
make pytest arch independent

oliverkurth:
Fixes issue #256

* determine the architecture with `platform.machine()` (python) or `uname -m` (shell) for package suffixes and directory names
* check valgrind version if suitable for architecture: arm needs >= 3.17.0

### DIFF
--- a/pytests/conftest.py
+++ b/pytests/conftest.py
@@ -25,7 +25,9 @@ from OpenSSL import crypto
 from urllib.parse import urlparse
 from multiprocessing import Process
 from http.server import SimpleHTTPRequestHandler, HTTPServer
+import platform
 
+ARCH=platform.machine()
 
 def StopTestRepoServer(server):
     server.terminate()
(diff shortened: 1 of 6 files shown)


pytests/conftest.py
```diff
@@ -131,6 +133,13 @@ def check_valgrind(self):
             self.config['valgrind_disabled_reason'] = 'valgrind minimum version constraint not met'
             return
 
+        if 'valgrind_minimum_version_{}'.format(ARCH) in self.config:
+            minimum_version = self.version_str_to_int(self.config['valgrind_minimum_version_{}'.format(ARCH)])
```
sshedi: is there a indentation error here?

sshedi (changes requested):

pytests/conftest.py
oliverkurth: No. Why? Outer if block (line 136) checks if `valgrind_minimum_version_ARCH` is set, otherwise we would get a val error in the next line.

Or are you confused because line 137 is too long? I should probably break it, like in 139/140.

sshedi: Yeah, my bad. Sorry. LGTM

sshedi (approved):

PR #258, opened 2021-08-19
Acquire rpm db lock before running transactions.

sshedi:
If not done so, tdnf and rpm will be in race while writing to rpm db.

rpmtxnBegin(pTS->pTS, RPMTXN_WRITE);

This call will try to acquire lock, if it can not then it will wait for
the lock to get opened and then proceed.

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -368,11 +368,12 @@ TDNFRunTransaction(
     PTDNF pTdnf
     )
 {
+    int rc;
     uint32_t dwError = 0;
     int rpmVfyLevelMask = 0;
     uint32_t dwSkipSignature = 0;
     uint32_t dwSkipDigest = 0;
-    int rc;
+    rpmtxn txn_lock = NULL;
 
     if(!pTS || !pTdnf || !pTdnf->pArgs)
     {
@@ -417,8 +418,12 @@ TDNFRunTransaction(
          }
          rpmtsSetVfyLevel(pTS->pTS, ~rpmVfyLevelMask);
     }
+
     rpmtsSetFlags(pTS->pTS, RPMTRANS_FLAG_TEST);
+
+    txn_lock = rpmtxnBegin(pTS->pTS, RPMTXN_WRITE);
     rc = rpmtsRun(pTS->pTS, NULL, pTS->nProbFilterFlags);
+    rpmtxnEnd(txn_lock);
     if (rc != 0)
     {
         dwError = ERROR_TDNF_TRANSACTION_FAILED;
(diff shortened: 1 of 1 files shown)


sshedi:
To see the issue in action:

Open two ssh sessions to your vm.
On one session 1 do `while true; do tdnf erase -y tar; tdnf install -y tar; done`
On the other session 2 run `while true; do  echo "start"; rpm --rebuilddb; echo "done" ; done`

You will see some transaction errors in Session 1 after a short time.

oliverkurth:
`rpmtsRun()` already calls `rpmtsBegin()`, see https://github.com/rpm-software-management/rpm/blob/4afe2d14d33db82ccb41c0a8d5eb1a4db90762fc/lib/transaction.c#L1780

But why do you get an error anyway?

Also, as the comment says, it's not needed for the test run.

oliverkurth:
Please consider using this instead (minus the printfs). It creates a lock as soon as possible, and releases it as late as possible. We still get an inconsistent db, but packages won't be uninstalled arbitrarily at least. Note that this lock will be held during downloads - I think that is acceptable.

Issue filed against rpm: https://github.com/rpm-software-management/rpm/issues/1761

```
diff --git a/client/rpmtrans.c b/client/rpmtrans.c
index 8fa055a..c01cafe 100644
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -32,6 +32,7 @@ TDNFRpmExecTransaction(
     int nDownloadOnly = 0;
     TDNFRPMTS ts = {0};
     PTDNF_CMD_OPT pSetOpt = NULL;
+    rpmtxn txn_lock = NULL;
 
     if(!pTdnf || !pTdnf->pArgs || !pTdnf->pConf || !pSolvedInfo)
     {
@@ -64,7 +65,14 @@ TDNFRpmExecTransaction(
         dwError = ERROR_TDNF_RPMTS_CREATE_FAILED;
         BAIL_ON_TDNF_ERROR(dwError);
     }
-
+    printf("acquiring lock...\n");
+    txn_lock = rpmtxnBegin(ts.pTS, RPMTXN_WRITE);
+    if (!txn_lock)
+    {
+        dwError = ERROR_TDNF_RPMTS_CREATE_FAILED;
+    }
+    printf("locked.\n");
     /* parse transaction flags - so far only tsflags=noscripts is
        supported. */
     ts.nTransFlags = RPMTRANS_FLAG_NONE;
@@ -106,6 +114,9 @@ cleanup:
     if(ts.pTS)
     {
         rpmtsCloseDB(ts.pTS);
+       printf("releasing...\n");
+        rpmtxnEnd(txn_lock);
+       printf("done\n");
         rpmtsFree(ts.pTS);
     }
     if(ts.pCachedRpmsArray)
```

PR #259, opened 2021-08-24
add minversions config option

oliverkurth:
This adds a config option to prevent downgrades below a specified version. This can be configured in the main config file (`/etc/tdnf/tdnf.conf`) like:
```
minversions=tdnf=3.1.4-1
```
Multiple packages can be specified separated by spaces.

Alternatively, packages can be set in files with names match `*.conf` in the directory `/etc/tdnf/minversions.d`, with one package per line:
```
tdnf=3.1.4-1
```
These can be easily added (and owned) by packages.

### DIFF
--- a/client/config.c
+++ b/client/config.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2020 VMware, Inc. All Rights Reserved.
+ * Copyright (C) 2015-2021 VMware, Inc. All Rights Reserved.
  *
  * Licensed under the GNU Lesser General Public License v2.1 (the "License");
  * you may not use this file except in compliance with the License. The terms
@@ -51,6 +51,8 @@ TDNFReadConfig(
     PTDNF_CONF pConf = NULL;
     PCONF_DATA pData = NULL;
     PCONF_SECTION pSection = NULL;
+    char *pszConfFileCopy = NULL;
+    char *pszMinVersionsDir = NULL;
 
     if(!pTdnf ||
        IsNullOrEmptyString(pszConfFile) ||
@@ -143,6 +145,16 @@ TDNFReadConfig(
                   pTdnf);
     BAIL_ON_TDNF_ERROR(dwError);
 
+    /* We need a copy of pszConfFile because dirname() modifies its argument */
+    dwError = TDNFAllocateString(pszConfFile, &pszConfFileCopy);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFAllocateStringPrintf(&pszMinVersionsDir, "%s/minversions.d", dirname(pszConfFileCopy));
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFReadMinVersionsFiles(pszMinVersionsDir, &pConf->ppszMinVersions);
+    BAIL_ON_TDNF_ERROR(dwError);
+
     pTdnf->pConf = pConf;
 
 cleanup:
@@ -150,6 +162,8 @@ cleanup:
     {
         TDNFFreeConfigData(pData);
     }
+    TDNF_SAFE_FREE_MEMORY(pszConfFileCopy);
+    TDNF_SAFE_FREE_MEMORY(pszMinVersionsDir);
     return dwError;
 
 error:
@@ -297,6 +311,9 @@ TDNFFreeConfig(
         TDNF_SAFE_FREE_MEMORY(pConf->pszDistroVerPkg);
         TDNF_SAFE_FREE_MEMORY(pConf->pszVarReleaseVer);
         TDNF_SAFE_FREE_MEMORY(pConf->pszVarBaseArch);
+        TDNF_SAFE_FREE_MEMORY(pConf->pszBaseArch);
+        TDNF_SAFE_FREE_STRINGARRAY(pConf->ppszExcludes);
+        TDNF_SAFE_FREE_STRINGARRAY(pConf->ppszMinVersions);
         TDNFFreeMemory(pConf);
     }
 }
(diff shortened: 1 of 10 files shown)


sshedi:
Changes LGTM. Please add tests for the same.

oliverkurth:
Updated to include files in `/etc/tdnf/minversions.d`. Tests still to do.

client/defines.h
```diff
@@ -205,6 +206,9 @@ typedef enum
     {ERROR_TDNF_RPM_CHECK,           "ERROR_TDNF_RPM_CHECK",           "rpm check reported errors"}, \
     {ERROR_TDNF_METADATA_EXPIRE_PARSE, "ERROR_TDNF_METADATA_EXPIRE_PARSE", "metadata_expire value could not be parsed. Check your repo files."},\
     {ERROR_TDNF_SELF_ERASE, "ERROR_TDNF_SELF_ERASE", "The operation would result in removing the protected package : tdnf"},\
+    {ERROR_TDNF_DOWNGRADE_NOT_ALLOWED,\
+	    "ERROR_TDNF_DOWNGRADE_NOT_ALLOWED",\
+	    "a downgrade is not allowed below the minimal version (check 'minversions' in the configuration)"},\
```
tapakund: how about 'Downgrade is not allowed below the minimal version. Check minversions in the configurations.'

client/packageutils.c
```diff
@@ -900,6 +900,84 @@ TDNFCheckProtectedPkgs(
...
 }
 
+uint32_t
+TDNFHasMinVersionInList(
+    PTDNF pTdnf,
+    PTDNF_PKG_INFO pPkgInfos
+    )
+{
+    uint32_t dwError = 0;
+    PTDNF_PKG_INFO pPkgInfo;
```
tapakund, line 910: Initialise PTDNF_PKG_INFO pPkgInfo = NULL; also int i = 0;

client/packageutils.c
```diff
@@ -900,6 +900,84 @@ TDNFCheckProtectedPkgs(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFSplitStringToArray(pszTemp, "=", &ppszTokens);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        if (ppszTokens[0] && ppszTokens[1])
+        {
+            for(pPkgInfo = pPkgInfos; pPkgInfo; pPkgInfo = pPkgInfo->pNext)
+            {
+                if (strcmp(ppszTokens[0], pPkgInfo->pszName) == 0)
```
tapakund, line 936: should we do strncasecmp , instead of strcmp. It would be more helpful to user. Just a suggestion, you can think and let me know if we can do else ignore.

client/config.c
```diff
@@ -143,13 +145,24 @@ TDNFReadConfig(
                   pTdnf);
     BAIL_ON_TDNF_ERROR(dwError);
 
+    dwError = TDNFAllocateString(pszConfFile, &pszConfFileCopy);
```
tapakund, line 149: do we need pszConfFileCopy? we arent adding anything to pszConfFile, right?

client/config.c
```diff
@@ -440,3 +456,119 @@ _TDNFConfigReadPluginSettings(
...
+    uint32_t dwError = 0;
+    DIR *pDir = NULL;
+    struct dirent *pEnt = NULL;
+    char *pszFile = NULL;
+    char **ppszNewMinVersions = NULL;
+    char ***pppszArrayList = NULL;
+    int nFileCount = 0;
+    int i, j, k;
+    int nLineCount = 0;
+    int nTmp;
```
tapakund: initialise nTmp with 0.

client/config.c
```diff
@@ -440,3 +456,119 @@ _TDNFConfigReadPluginSettings(
...
+
```
tapakund, line 488: pls check for the arguments if they are valid.

client/config.c
```diff
@@ -440,3 +456,119 @@ _TDNFConfigReadPluginSettings(
...
+        if (fnmatch("*.conf", pEnt->d_name, 0) != 0)
+        {
+            continue;
+        }
+        dwError = TDNFAllocateStringPrintf(&pszFile, "%s/%s", pszDir, pEnt->d_name);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFReadFileToStringArray(pszFile, &pppszArrayList[i]);
+        BAIL_ON_TDNF_ERROR(dwError);
+        i++;
```
tapakund, line 530: shouldn't we free pszFile here, as its allocated in loop.

tapakund (commented):
Some nits and comments, would like to take one more look once these are addressed.

client/packageutils.c
oliverkurth: I am not sure about this - it looks like when I do `sudo tdnf install LSoF` (real name is `lsof`) gives a message: `[ignoring case for 'LSoF']`. But this isn't really targeted for users who do not know what they are doing, so I'd prefer to have it case sensitive.

client/config.c
oliverkurth: We use `dirname` on it in line 151, which modifies the argument. We don't want to mess up `pszConfFile` which was received in this function as an argument. I can add a comment.

client/config.c
oliverkurth: Done.

client/config.c
oliverkurth: Done.

client/config.c
oliverkurth: Good catch! Fixed.

client/defines.h
oliverkurth: Fixed.

client/packageutils.c
oliverkurth: Not really needed, both are initialized in the `for()` statements.

client/packageutils.c
tapakund: Agree.

client/config.c
```diff
@@ -434,3 +457,128 @@ _TDNFConfigReadPluginSettings(
 error:
     goto cleanup;
 }
+
+uint32_t
+TDNFReadMinVersionsFiles(
```
tapakund, line 467: Can you help explain this function in more detail?

oliverkurth: I added a few comments.

client/config.c
```diff
@@ -434,3 +457,142 @@ _TDNFConfigReadPluginSettings(
...
+        {
+            ppszNewMinVersions[k++] = pppszArrayList[i][j];
+        }
+        TDNF_SAFE_FREE_MEMORY(pppszArrayList[i]);
+    }
+
+    *pppszMinVersions = ppszNewMinVersions;
+
+cleanup:
+    TDNF_SAFE_FREE_MEMORY(pppszArrayList);
```
tapakund, line 568: I think we should check if pDir is not NULL and call closedir in cleanup. As if there is an error in between the loops reading files then we will miss this.
if(pDir)
    {
        closedir(pDir);
    }

Just check if closedir sets pDir to NULL, else we should set it to NULL after doing closedir

client/config.c
```diff
@@ -434,3 +457,142 @@ _TDNFConfigReadPluginSettings(
...
+
+        dwError = TDNFReadFileToStringArray(pszFile, &pppszArrayList[i]);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        TDNF_SAFE_FREE_MEMORY(pszFile);
+
+        i++;
+    }
+    closedir(pDir);
+    pppszArrayList[i] = *pppszMinVersions;
```
tapakund, line 536: as discussed, it would be good to add some comment for this.

client/config.c
```diff
@@ -434,3 +457,142 @@ _TDNFConfigReadPluginSettings(
...
+    for (i = 0, k = 0; pppszArrayList[i]; i++)
+    {
+        for (j = 0; pppszArrayList[i][j]; j++)
+        {
+            ppszNewMinVersions[k++] = pppszArrayList[i][j];
+        }
+        TDNF_SAFE_FREE_MEMORY(pppszArrayList[i]);
+    }
+
+    *pppszMinVersions = ppszNewMinVersions;
```
tapakund, line 561: So, here we override pppszMinVersions with the whole list of ppszNewMinVersions?

tapakund (approved):
Looks fine to me, I have some comments and queries. Pls help answer/address them. Thanks!

client/config.c
oliverkurth: Good find, fixed.

client/config.c
oliverkurth: Yes, but we saved the content above (line 531).

client/config.c
oliverkurth: Done.

PR #260, opened 2021-09-08
 ITS issue fixed for tdnf.

ntsbtz:


ntsbtz:
@gpiyush-dev please review this .

gpiyush-dev (commented):
Not sure if we should modify COPYING file. Rest looks good to me.

ntsbtz:
@tapakund  Is it fine modifying COPYING file? Rest Piyush has reviewed it can you please merge if nothing  required to change.

client/api.c
```diff
@@ -469,7 +469,7 @@ TDNFClean(
 }
 
 //Show count of installed
-//not a dnf command just sanity check
+//not a dnf command just confidence check
```
sshedi: I think we can use `// not a tdnf command, simple validation` instead

include/tdnf.h
```diff
@@ -136,7 +136,7 @@ TDNFUpdateInfoSummary(
     PTDNF_UPDATEINFO_SUMMARY* ppSummary
     );
 
-//sanity check. displays current installed count.
+//confidence check. displays current installed count.
```
sshedi, line 139: same here. Use a better comment

sshedi (changes requested):

client/api.c
ntsbtz: resolved

keerthanakalyan (approved):

PR #261, opened 2021-09-15
lock early and long

oliverkurth:
This is another way than pr #258 to lock the db database. This locks it as early as possible and releases it as late as possible. This still runs into the same issue as https://github.com/rpm-software-management/rpm/issues/1761 , but this most likely cannot be addressed in tdnf.


oliverkurth:
This is no longer needed, instead we are going to use this fix: https://github.com/rpm-software-management/rpm/issues/1761#issuecomment-921969050

PR #264, opened 2021-10-25
use TDNFJoin() to construct paths

oliverkurth:
This adds the `TDNFJoin()` function to create directory paths.

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -784,3 +784,88 @@ cleanup:
 error:
     goto cleanup;
 }
+
+uint32_t
+TDNFJoinPath(char **ppszPath, ...)
+{
+    uint32_t dwError = 0;
+    va_list ap;
+    char *pszNode = NULL;
+    int i, nCount = 0;
+    char *pszTmp, *pszNodeTmp;
+    char *pszNodeCopy = NULL;
+    char *pszResult = NULL;
+    int nLength = 0;
+    int nLengthTmp = 0;
+
+    if (!ppszPath)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    va_start(ap, ppszPath);
+    for(pszNode = va_arg(ap, char *); pszNode; pszNode = va_arg(ap, char *))
+    {
+        nCount++;
+    }
+    va_end(ap);
+
+    va_start(ap, ppszPath);
+    for(pszNode = va_arg(ap, char *), i = 0; pszNode; pszNode = va_arg(ap, char *), i++)
+    {
+        dwError = TDNFAllocateString(pszNode, &pszNodeCopy);
+        BAIL_ON_TDNF_ERROR(dwError);
+        pszNodeTmp = pszNodeCopy;
+	/* if the first node is an absolute path, the result should be absolute -
+	 * safe this by initializing with a '/' if absolute, otherwise with an empty string
+	 * before stripping all leading slashes */
+        if (i == 0)
+        {
+            if (*pszNodeTmp == '/')
+            {
+                dwError = TDNFAllocateString("/", &pszResult);
+                nLength++;
+            }
+            else
+            {
+                dwError = TDNFAllocateString("", &pszResult);
+            }
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+	/* now strip leading slashes */
+        while(*pszNodeTmp == '/') pszNodeTmp++;
+
+	/* strip trailing slashes */
+        nLengthTmp = strlen(pszNodeTmp);
+        pszTmp = pszNodeTmp + nLengthTmp - 1;
+        while(pszTmp >= pszNodeTmp && *pszTmp == '/')
+        {
+            *pszTmp = 0;
+            pszTmp--;
+        }
+        nLength += nLengthTmp + 2;
+
+        dwError = TDNFReAllocateMemory(nLength, (void **)&pszResult);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        strcat(pszResult, pszNodeTmp);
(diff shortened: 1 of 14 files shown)


common/utils.c
```diff
@@ -760,3 +760,89 @@ TDNFStringMatchesOneOf(const char *pszSearch, char **ppszList, int *pRet)
...
+    int nAbsolute = 0;
+
+    if (!ppszPath)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    va_start(ap, ppszPath);
+    for(pszNode = va_arg(ap, char *); pszNode; pszNode = va_arg(ap, char *))
```
sshedi, line 808: if this loop is for counting number of va args, then it doesn't work.
get rid of next for loop (at line 795) and process args one by one withing this loop.

common/utils.c
```diff
@@ -760,3 +760,89 @@ TDNFStringMatchesOneOf(const char *pszSearch, char **ppszList, int *pRet)
...
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    for(pszNode = va_arg(ap, char *), i = 0; pszNode; pszNode = va_arg(ap, char *), i++)
+    {
+        dwError = TDNFAllocateString(pszNode, &pszNodeCopy);
+        BAIL_ON_TDNF_ERROR(dwError);
+        pszNodeTmp = pszNodeCopy;
+        if (i == 0 && *pszNodeTmp == '/')
+        {
+            nAbsolute = 1;
```
sshedi: nit: do pszNodeTmp++ here, saves one iteration :)

common/utils.c
```diff
@@ -760,3 +760,89 @@ TDNFStringMatchesOneOf(const char *pszSearch, char **ppszList, int *pRet)
...
+        }
+        while(*pszNodeTmp == '/') pszNodeTmp++;
```
sshedi, line 837: User strspn instead.

common/utils.c
```diff
@@ -760,3 +760,89 @@ TDNFStringMatchesOneOf(const char *pszSearch, char **ppszList, int *pRet)
...
+        pszTmp = pszNodeTmp + strlen(pszNodeTmp) - 1;
```
sshedi: store `strlen(pszNodeTmp)` in a variable. it's again used in line 813

common/utils.c
```diff
@@ -760,3 +760,89 @@ TDNFStringMatchesOneOf(const char *pszSearch, char **ppszList, int *pRet)
...
+
+    pszTmp = pszResult;
+    if (nAbsolute)
+    {
+        *pszTmp = '/';
+        pszTmp++;
+    }
+    for (i = 0; i < nCount; i++)
+    {
+        strcpy(pszTmp, ppszNodes[i]);
```
sshedi: Use strcat here?

common/utils.c
```diff
@@ -760,3 +760,89 @@ TDNFStringMatchesOneOf(const char *pszSearch, char **ppszList, int *pRet)
...
+        while(*pszNodeTmp == '/') pszNodeTmp++;
+        pszTmp = pszNodeTmp + strlen(pszNodeTmp) - 1;
+        while(pszTmp >= pszNodeTmp && *pszTmp == '/')
+        {
+            *pszTmp = 0;
+            pszTmp--;
+        }
+        dwError = TDNFAllocateString(pszNodeTmp, &ppszNodes[i]);
+        BAIL_ON_TDNF_ERROR(dwError);
+        nLength += strlen(pszNodeTmp) + 1;
```
sshedi: you can just keep appending path to a resultant buffer upon every iteration and reallocate memory as needed.

sshedi (changes requested):

common/utils.c
oliverkurth: Thx - even better, I need the `while` loop only if the condition is true.

common/utils.c
oliverkurth: That's not making the code any more readable or efficient.

common/utils.c
oliverkurth: Thx! Done.

common/utils.c
oliverkurth: Last arg must be `NULL`. As discussed this internally.

common/utils.c
oliverkurth: Done.

common/utils.c
oliverkurth: Done.

sshedi (approved):
LGTM. One improvement I can think of is, if given an absolute path call NormalizePath function and use the return value as is.

oliverkurth:
Build failure is unrelated: Photon 4.0 updated to openssl 3.0.0. Possible fix:
```
diff --git a/client/remoterepo.c b/client/remoterepo.c
index 22c870e..e53d0d6 100644
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -170,9 +170,13 @@ TDNFGetDigestForFile(
     {
         pr_err("Digest Init Failed\n");
         dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
-        /*MD5 is not approved in FIPS mode. So, overrriding
-          the dwError to show the right error to the user */
+        /* MD5 is not approved in FIPS mode. So, overrriding
+           the dwError to show the right error to the user */
+#if defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
+        if (EVP_default_properties_is_fips_enabled(NULL) && !strcasecmp(hash->hash_type, "md5"))
+#else
         if (FIPS_mode() && !strcasecmp(hash->hash_type, "md5"))
+#endif
         {
             dwError = ERROR_TDNF_FIPS_MODE_FORBIDDEN;
         }
```

PR #267, opened 2021-11-09
wait for lock until timeout

oliverkurth:
Generating the cache can take about a minute if the repository is large (Photon 3.0). On Photon there is a systemd unit that runs `update-info` on startup. This can break testing scripts that run right after booting. This change will make tdnf wait to acquire the lock for up to 5 minutes before timing out and exiting.

### DIFF
--- a/tools/cli/defines.h
+++ b/tools/cli/defines.h
@@ -27,6 +27,9 @@
  */
 #define TDNF_INSTANCE_LOCK_FILE     "/var/run/.tdnf-instance-lockfile"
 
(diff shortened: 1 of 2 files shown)


oliverkurth:
Test failure with Fedora seems unrelated. I cannot reproduce it locally.

oliverkurth:
> Test failure with Fedora seems unrelated. I cannot reproduce it locally.

I can reproduce now with the latest Fedora container. It is unrelated to this change, I see it in the `dev` branch too.

sshedi (approved):

PR #269, opened 2021-11-29
fix building with openssl 3.0

oliverkurth:
Latest Photon 4.0 uses openssl 3.0 which has `FIPS_mode()` deprecated.

### DIFF
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -170,9 +170,13 @@ TDNFGetDigestForFile(
     {
         pr_err("Digest Init Failed\n");
         dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
-        /*MD5 is not approved in FIPS mode. So, overrriding
-          the dwError to show the right error to the user */
+        /* MD5 is not approved in FIPS mode. So, overrriding
+           the dwError to show the right error to the user */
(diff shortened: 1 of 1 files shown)


tapakund (approved):

satya-rajesh (approved):

satya-rajesh:
> Latest Photon 4.0 uses openssl 3.0 which has `FIPS_mode()` deprecated.

@oliverkurth 
A small query, Is tdnf harness run with this change ?

oliverkurth:
> A small query, Is tdnf harness run with this change ?

Not sure if I understand the question. The usual workflow for tdnf is to merge this, then create a release, which can then be consumed in Photon. Or this change is used via spec file as a patch. Then it will be used by the harness. But we have a similar patch already, the difference is that this uses NULL, not ctx. The latter gives a compiler warning because the type doesn't match.

satya-rajesh:
> > A small query, Is tdnf harness run with this change ?
> 
> Not sure if I understand the question. The usual workflow for tdnf is to merge this, then create a release, which can then be consumed in Photon. Or this change is used via spec file as a patch. Then it will be used by the harness. But we have a similar patch already, the difference is that this uses NULL, not ctx. The latter gives a compiler warning because the type doesn't match.

Agree, thanks for the details.

PR #271, opened 2021-12-06
check file from command line for rpm extension

### LINKED ISSUE #270: TDNFAddCmdLinePackages fails when a file or symlink exists with same name as package name
When a file or link is present in the working directory, with the same name as the package name in the cli arguments, tdnf aborts. Here is an example that works normaly:

```
[sgugshell root@octane ~]# tdnf search irix  
libdicl : Dans Irix Compatibility Library
microdnf : Minimal C implementation of DNF tweaked for IRIX
chrpath : Modify rpath of compiled programs
libdicl : Dans Irix Compatibility Library
microdnf : Minimal C implementation of DNF tweaked for IRIX
glew : The OpenGL Extension Wrangler Library
libGL : Irix OpenGL library
libdicl : Dans Irix Compatibility Library
microdnf : Minimal C implementation of DNF tweaked for IRIX
```

The following is less normal:

```
[sgugshell root@octane ~]# touch irix
[sgugshell root@octane ~]# tdnf search irix
```

This is due to a check for an existing file or symlink with the same name as the package name in `client/api.c`, where `TDNFAddCmdLinePackages` checks each package name. There are no comments in that region of the code, so I'm not sure what this test is supposed to protect.

oliverkurth:
Check file from command line for "*.rpm" extension to avoid confusing a file with a package name. Add a test and comments.

Note to reviewers: there are only two changed lines in functional code, the rest is comments and white space fixes, plus the test code. The only change is changing
`if (nIsFile)`
to
`if (nIsFile && (fnmatch("*.rpm", pszPkgName, 0) == 0))`
in two places.

Closes #270

### DIFF
--- a/client/resolve.c
+++ b/client/resolve.c
@@ -173,7 +173,7 @@ TDNFPrepareAllPackages(
                dwError = TDNFIsFileOrSymlink(pszPkgName, &nIsFile);
                BAIL_ON_TDNF_ERROR(dwError);
 
-               if (nIsFile)
+               if (nIsFile && (fnmatch("*.rpm", pszPkgName, 0) == 0))
                {
                    continue;
                }
--- a/pytests/tests/test_rpm_cmdline.py
+++ b/pytests/tests/test_rpm_cmdline.py
@@ -125,7 +125,12 @@ def test_install_as_mixed(utils):
     assert(utils.check_package(pkgname) == True)
     assert(utils.check_package(pkgname2) == True)
 
-
-
-
-
+# test installing a package that has the same name as a file
+# example: touch foo; tdnf install foo
+# (file needs to have "*.rpm" extension to qualify)
+def test_install_same_as_filname(utils):
+    pkgname = utils.config["sglversion_pkgname"]
+    utils.run(['touch', pkgname])
+    ret = utils.run([ 'tdnf', 'install', '-y', '--nogpgcheck', pkgname])
+    assert(ret['retval']  == 0)
+    assert(utils.check_package(pkgname) == True)
--- a/client/api.c
+++ b/client/api.c
@@ -775,12 +775,16 @@ TDNFAddCmdLinePackages(
 
     for(nCmdIndex = 1; nCmdIndex < pCmdArgs->nCmdCount; ++nCmdIndex)
     {
+        /* Add packages with URLs or filenames on the command line
+         * to our virtual @cmdline repo */
         pszPkgName = pCmdArgs->ppszCmds[nCmdIndex];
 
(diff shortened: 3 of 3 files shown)


client/resolve.c
```diff
@@ -173,7 +173,7 @@ TDNFPrepareAllPackages(
                dwError = TDNFIsFileOrSymlink(pszPkgName, &nIsFile);
                BAIL_ON_TDNF_ERROR(dwError);
 
-               if (nIsFile)
+               if (nIsFile && (fnmatch("*.rpm", pszPkgName, 0) == 0))
```
sshedi, line 176: should we print a warning or error if file is not rpm?

pytests/tests/test_rpm_cmdline.py
```diff
@@ -125,7 +125,12 @@ def test_install_as_mixed(utils):
...
     assert(utils.check_package(pkgname2) == True)
 
-
-
-
-
+# test installing a package that has the same name as a file
+# example: touch foo; tdnf install foo
+# (file needs to have "*.rpm" extension to qualify)
+def test_install_same_as_filname(utils):
```
sshedi, line 131: add a negative case as well

sshedi (changes requested):

tapakund (approved):

client/resolve.c
oliverkurth: I don't think so. If there is a file `glibc` and you type`tdnf install glibc`, tdnf will check but installing the package `glibc` (not the file) is almost certainly what you wanted.

pytests/tests/test_rpm_cmdline.py
oliverkurth: What would a negative test do in this case?

client/resolve.c
sshedi: Understood. But in that case, `nIsFile` will be false right? So, you can move the fnmatch part inside the if and print error if not rpm.

pytests/tests/test_rpm_cmdline.py
sshedi: A file with some other extension other than rpm

client/resolve.c
oliverkurth: No, this is to cover the special case where `glibc`, in this example *is* a file, so `nIsFile` is true. And it's not an error - we just ignore the file and install `glibc` from the repositories.

pytests/tests/test_rpm_cmdline.py
oliverkurth: That is what this test does exactly. `pkgname` is the name of a package, not a filename ending in `.rpm`.

sshedi (approved):

PR #272, opened 2021-12-10
add package lock feature

oliverkurth:
This adds the package lock feature. An installed package can be "locked", which means that it cannot be removed, updated or downgraded with tdnf.

To use it, create a directory `locks.d` in the same directory as the tdnf config file (usually `/etc/tdnf/` , unless overwritten with the `-c` option), and create a file with the extension `.conf` with one package name per line. For example:
```
tdnf install lsof
mkdir /etc/tdnf/locks.d
echo lsof > /etc/tdnf/locks.d/lsof.conf
```

### DIFF
--- a/client/goal.c
+++ b/client/goal.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015 VMware, Inc. All Rights Reserved.
+ * Copyright (C) 2015-2022 VMware, Inc. All Rights Reserved.
  *
  * Licensed under the GNU Lesser General Public License v2.1 (the "License");
  * you may not use this file except in compliance with the License. The terms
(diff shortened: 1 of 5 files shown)


client/goal.c
```diff
@@ -619,3 +622,55 @@ TDNFGoalGetAllResultsIgnoreNoData(
...
+            {
+                if (idPkg == s->name)
+                {
+                    nFound = 1;
+                    break;
+                }
+            }
+            if (nFound)
+            {
+                queue_push2(pQueueJobs, SOLVER_SOLVABLE_NAME|SOLVER_LOCK, idPkg);
```
sshedi: use this at line 660 and get rid of nFound variable.

client/goal.c
```diff
@@ -619,3 +622,55 @@ TDNFGoalGetAllResultsIgnoreNoData(
...
+TDNFSolvAddPkgLocks(
+    PTDNF pTdnf,
+    Queue* pQueueJobs,
+    Pool *pPool
+    )
+{
+    uint32_t dwError = 0;
+    char **ppszPackages = NULL;
+    char *pszPkg = NULL;
+    Id idPkg;
```
sshedi: we can reduce the scope of few of these variables by declaring them inside loop where required.

sshedi (changes requested):

client/goal.c
oliverkurth: Done.

sshedi (approved):

Issue #273, opened 2021-12-15
tdnf can't cope with RPM repos that do not contain a `filelists`


andrew-lee-1089:
I have two RPM repos (created by artifactory), comapring the repodata/repomd.xml , in the working case we have:

```
<?xml version="1.0"?>
<repomd xmlns="http://linux.duke.edu/metadata/repo" xmlns:rpm="http://linux.duke.edu/metadata/rpm">
  <data type="other">
    ...
  </data>
  <data type="filelists">
    ...
  </data>
  <data type="primary">
    ...
  </data>
</repomd>
```

Failing case:

```
<?xml version="1.0"?>
<repomd xmlns="http://linux.duke.edu/metadata/repo" xmlns:rpm="http://linux.duke.edu/metadata/rpm">
  ​<data type="other">
    ​...
  </data>
  <data type="primary">
    ...
  </data>
</repomd>
``` 

Looking at the code in https://github.com/vmware/tdnf/blob/v2.1.0/client/repo.c#L878:

```
    dwError = TDNFFindRepoMDPart(
                  pRepo,
                  TDNF_REPOMD_TYPE_FILELISTS,
                  &pRepoMD->pszFileLists);
    BAIL_ON_TDNF_ERROR(dwError);

    dwError = TDNFFindRepoMDPart(
                  pRepo,
                  TDNF_REPOMD_TYPE_UPDATEINFO,
                  &pRepoMD->pszUpdateInfo);
    //updateinfo is not mandatory
    if(dwError == ERROR_TDNF_NO_DATA)
    {
        dwError = 0;
    }
    BAIL_ON_TDNF_ERROR(dwError);
```

It looks like tdnf can't cope with this. Both yum and dnf can cope with this. When I last looked at it ,the specs for what exactly should be in a repomd.xml file were not clear (i.e. http://linux.duke.edu/metadata/rpm is a dead link) - but I think I would welcome a defensive fix in tdnf to cope with this. We already deem 'updateinfo' to not be mandatory.

oliverkurth:
Thank you for reporting this.

Do you happen to know a way to reproduce a repo without a filelist, other than editing the `repomd.xml` file?

The fix is easy, but I would like to write a test for it.

andrew-lee-1089:
The way I did was by creating a brand new RPM repo using JFROG's artifactory. I don't know if you have access to that.

keerthanakalyan:
> Thank you for reporting this.
> 
[...]

PR #275, opened 2021-12-16
ignore missing filelists in metadata (issue #273)

oliverkurth:
closes issue #273 .

### DIFF
--- a/pytests/tests/test_repo_missing_metadata.py
+++ b/pytests/tests/test_repo_missing_metadata.py
@@ -12,10 +12,10 @@ import shutil
 import pytest
 import errno
 
-REPODIR='/root/repofrompath/yum.repos.d'
-REPOFILENAME='repofrompath.repo'
-REPONAME="repofrompath-test"
-WORKDIR='/root/repofrompath/workdir'
+REPODIR='/root/repo_missing_metadata/yum.repos.d'
+REPOFILENAME='repo_missing_metadata.repo'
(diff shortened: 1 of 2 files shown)


tapakund (approved):

pytests/tests/test_repo_missing_metadata.py
```diff
@@ -0,0 +1,60 @@
...
+# of the License are located in the COPYING file of this distribution.
+#
+#   Author: Oliver Kurth <okurth@vmware.com>
+
+import os
+import shutil
+import pytest
+import errno
+
+REPODIR='/root/repofrompath/yum.repos.d'
```
sshedi: Use `REPODIR=/tmp/...` if test fails, our root directory won't contain these files

pytests/tests/test_repo_missing_metadata.py
```diff
@@ -0,0 +1,60 @@
...
+REPOFILENAME='repofrompath.repo'
+REPONAME="repofrompath-test"
+WORKDIR='/root/repofrompath/workdir'
```
sshedi: same here

sshedi (approved):
Minor nit picking comment. LGTM

pytests/tests/test_repo_missing_metadata.py
oliverkurth: Should be okay since we clean it up, but I changed it anyway.

PR #276, opened 2021-12-22
Move tdnf single instance feature to library

sshedi:


sshedi:
@tapakund @oliverkurth please review this fix.
I don't know why tests are not finishing.

The previous locking mechanism doesn't work properly after introducing retries and delay.

sshedi:
Will send it from a branch branch.

PR #277, opened 2021-12-26
fix progress_cb function

sshedi:
This is a nit picking fix.
The existing method of showing progress was too fast and not really
useful.

Let's show the progress in the increments of 10%. Easy to read, less
rapid prints.

### DIFF
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -1,22 +1,11 @@
 /*
- * Copyright (C) 2015-2018 VMware, Inc. All Rights Reserved.
+ * Copyright (C) 2015-2022 VMware, Inc. All Rights Reserved.
  *
  * Licensed under the GNU Lesser General Public License v2.1 (the "License");
  * you may not use this file except in compliance with the License. The terms
  * of the License are located in the COPYING file of this distribution.
  */
 
-/*
- * Module   : remoterepo.c
- *
- * Abstract :
- *
- *            tdnfclientlib
- *
- *            client library
- *
- * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
- */
 #include "includes.h"
 
 typedef struct _hash_op {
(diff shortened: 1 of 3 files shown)


oliverkurth:
I don't think this is the right way to do. If the download is slow (but not stopped) you'd see no progress for a while with this change.

Instead, if you think updates are too frequent, ignore it every 10th time or so. That would probably not slow it down by too much. Or, maybe even better, ignore the update if it comes sooner than maybe one second than the last one.

Unfortunately, looking at https://curl.se/libcurl/c/CURLOPT_XFERINFOFUNCTION.html , there is no way to just configure this with libcurl API, which would be ideal.

sshedi:
> I don't think this is the right way to do. If the download is slow (but not stopped) you'd see no progress for a while with this change.
> 
> Instead, if you think updates are too frequent, ignore it every 10th time or so. That would probably not slow it down by too much. Or, maybe even better, ignore the update if it comes sooner than maybe one second than the last one.
> 
> Unfortunately, looking at https://curl.se/libcurl/c/CURLOPT_XFERINFOFUNCTION.html , there is no way to just configure this with libcurl API, which would be ideal.

Good idea. Done.

oliverkurth:
I just tested this. Something is wrong with the progress display:
```
Total installed size:  25.51M 26745375
Is this ok [y/N]: y

Downloading:
tdnf-cli-libs                       0%
tdnf                                0%
rpm-build                           0%
rpm-devel                           0%
rpm-libs                            0%
rpm                                 0%
...
```
Honestly, I am not sure if it's worthwhile to change this at all.

sshedi:
Nothing wrong in this, there is a 3 second wait time before showing the progress. Within that time, everything gets finished.

To get a feel of this refer https://curl.se/libcurl/c/CURLOPT_MAX_RECV_SPEED_LARGE.html & use

`curl_easy_setopt(curl, CURLOPT_MAX_RECV_SPEED_LARGE, (curl_off_t)100*1024); //100 kb max download speed`

and use it just before https://github.com/vmware/tdnf/blob/727a8248fa3b47ba521df62d611c947fcafaaa46/client/remoterepo.c#L650 (curl_easy_perform)

You will see progress percentage working.

oliverkurth:
It's not acceptable. If it seems to stop with < 100% I am sure people will be wondering if the download was complete. You could make an exception to the 3 seconds rule and display the percentage  (that is 100%) when `dlNow == dlTotal`.

Also, 3 seconds it too long, please make it 1 second. After 3 seconds I already get nervous that the download is stuck.

sshedi:
Done

oliverkurth:
I still see 3 seconds. There is also a conflict now.

client/remoterepo.c
```diff
@@ -62,20 +51,36 @@ progress_cb(
...
+    if (dlTotal <= 0)
+    {
         return 0;
     }
 
-    dPercent = ((double)dlNow / (double)dlTotal) * 100.0;
-    if (!isatty(STDOUT_FILENO)) {
-        pr_info("%s %3.0f%% %ld\n", (char *)pUserData, dPercent, dlNow);
-    } else {
-        pr_info("%-35s %10ld %5.0f%%\r", (char *)pUserData, dlNow, dPercent);
```
oliverkurth, line 78: Why did you remove the display of `dlNow`? I don't think it's useless information.

oliverkurth (commented):
Also, when downloading metadata the display looks weird. Can you have a look at that?

sshedi:
> I still see 3 seconds. There is also a conflict now.

That is the whole point of this PR. Rate limited progress bar display. I reduced it to 1 second.

PR #278, opened 2021-12-26
Refactor locking

sshedi:


ci/Dockerfile.photon
```diff
@@ -1,17 +1,16 @@
 FROM photon:latest
 
-MAINTAINER csiddharth@vmware.com
-
-RUN tdnf update  -q -y
-RUN tdnf remove  -q -y toybox
-RUN tdnf install -q -y --enablerepo=photon-debuginfo \
+RUN tdnf update  -qy
+RUN tdnf remove  -qy toybox
+RUN tdnf install -qy --enablerepo=photon-debuginfo \
```
oliverkurth: I think (while you are at it), we should remove the `-q` flag. I don't see why we shouldn't see the progress and it may make it easier to debug if there are issues.

common/lock.c
```diff
@@ -0,0 +1,238 @@
...
+		{
+			lock->openmode = TDNFLOCK_READ;
+		}
+	}
+	else
+	{
+		lock->openmode = TDNFLOCK_WRITE | TDNFLOCK_READ;
+	}
+
+	if (lock)
```
oliverkurth: Why test for `lock` here? It shouldn't ever be NULL if `TDNFAllocateMemory` above succeeded, which it did if we get here. Besides, we would have crashed in line 20 (`lock->fd = open(...`)

common/lock.c
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return lock;
+
+error:
+	if (lock)
+	{
+		TDNF_SAFE_FREE_MEMORY(lock->path);
+		TDNF_SAFE_FREE_MEMORY(lock->descr);
+		TDNF_SAFE_FREE_MEMORY(lock);
```
oliverkurth: You can call `tdnflock_free()` here instead.

common/lock.c
```diff
@@ -0,0 +1,238 @@
...
+	return lock;
+}
+
+static void tdnflock_free(tdnflock lock)
+{
+	if (lock && --lock->fdrefs == 0)
+	{
+		TDNF_SAFE_FREE_MEMORY(lock->path);
+		TDNF_SAFE_FREE_MEMORY(lock->descr);
+		(void) close(lock->fd);
```
oliverkurth: Check for valid `fd` with `if (lock-fd != -1)`, so you can use it in the error case in `tdnflock_new`.

common/lock.c
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return lock;
+}
+
+static void tdnflock_free(tdnflock lock)
+{
+	if (lock && --lock->fdrefs == 0)
+	{
+		TDNF_SAFE_FREE_MEMORY(lock->path);
```
oliverkurth, line 80: Check for pointers `!= NULL` before freeing, so this can be used in error cases.

ci/Dockerfile.photon
sshedi: Done

common/lock.c
sshedi: Check line 42

oliverkurth: You mean line 47. Got it. But why don't you `goto error` and do the cleanup there?

sshedi: I simplified it. This should suffice.

sshedi:
Shall we merge this?

PR #280, opened 2022-01-04
Refactor minversions logic

oliverkurth:
This change uses libsolv to filter out packages below the configured minimal versions of packages. This makes it possible for libsolv to find alternative solutions instead of blocking a transaction.


client/goal.c
```diff
@@ -667,3 +670,93 @@ TDNFSolvAddPkgLocks(
...
+
+    dwError = TDNFAllocateMemory(
+                          1,
+                          sizeof(Map),
+                          (void**)&pMapMinVersions);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    map_init(pMapMinVersions, pPool->nsolvables);
+
+    for (i = 0; ppszPackages && ppszPackages[i]; i++)
```
sshedi: use `for (int i = 0; ...)` reduce scope of i

client/goal.c
```diff
@@ -667,3 +670,93 @@ TDNFSolvAddPkgLocks(
...
+    {
+        pszPkg = ppszPackages[i];
```
sshedi: same here, `char *pszPkg` and please check for other variables whose cope can be reduced. Seems like too many local variables at top.

sshedi (approved):

PR #283, opened 2022-01-07
implement autoremove feature

oliverkurth:
This implements the autoremove feature. This makes tdnf remove automatically installed packages, that is packages that were installed to satisfy dependencies. This can be done in one of three ways:

* enable it in the config with `clean_requirements_on_remove=true`(This was the default, but it took no effect. This change  removes the flag to prevent unintentional package removals)
* use the `--autoremove` flag for any command that may remove packages
* use the `autoremove` command instead of `remove` or `erase`.

This implementation uses the file `/var/lib/tdnf/autoinstalled` for a list of all automatically installed packages. Whenever a package may be removed, this list is checked and all installed packages that are not in the list will be considered _user installed_. (Therefore if the file does not exist, all installed packages are considered user installed). The list of user installed files is passed on to the libsolv API, and the flag `SOLVER_CLEANDEPS` is set. This will make libsolv add orphaned dependencies to the list of packages to be removed.

Newly installed packages will be marked as user installed, but not any dependencies. If a package is already installed it will be marked as user installed even if it was automatically installed before.

### DIFF
--- a/client/goal.c
+++ b/client/goal.c
@@ -357,6 +357,12 @@ TDNFGoal(
     {
         nFlags = nFlags | SOLVER_FORCEBEST;
     }
+
+    if (pTdnf->pConf->nCleanRequirementsOnRemove)
+    {
+        nFlags = nFlags | SOLVER_CLEANDEPS;
+    }
+
     dwError = SolvAddFlagsToJobs(&queueJobs, nFlags);
     BAIL_ON_TDNF_ERROR(dwError);
 
--- a/client/goal.c
+++ b/client/goal.c
@@ -420,7 +420,7 @@ TDNFGoal(
 
         if (nAlterType == ALTER_UPGRADE && dwExcludeCount != 0 && ppszExcludes)
         {
-            /* if we had packages to exclude, then we'd have disabled ones too */
+            /* if we had packages to exclude, then we'd have inactive ones too */
             dwSkipProblem |= SKIPPROBLEM_DISABLED;
         }
 
@@ -671,17 +671,16 @@ TDNFMarkAutoInstalled(
             fprintf(fp, "%s\n", ppszAutoInstalled[i]);
         }
     }
-    fclose(fp);
 
 cleanup:
-    TDNF_SAFE_FREE_MEMORY(pszAutoFile);
-    TDNF_SAFE_FREE_STRINGARRAY(ppszAutoInstalled);
-    return dwError;
-error:
     if (fp)
     {
         fclose(fp);
     }
+    TDNF_SAFE_FREE_MEMORY(pszAutoFile);
+    TDNF_SAFE_FREE_STRINGARRAY(ppszAutoInstalled);
+    return dwError;
+error:
     goto cleanup;
 }
 
--- a/client/goal.c
+++ b/client/goal.c
@@ -316,6 +316,7 @@ TDNFGoal(
     int nProblems = 0;
     char** ppszExcludes = NULL;
     uint32_t dwExcludeCount = 0;
+    char **ppszAutoInstalled = NULL;
 
     if(!pTdnf || !ppInfo || !pQueuePkgList)
     {
@@ -357,8 +358,9 @@ TDNFGoal(
     {
         nFlags = nFlags | SOLVER_FORCEBEST;
     }
-
-    if (pTdnf->pConf->nCleanRequirementsOnRemove)
+    if ((pTdnf->pConf->nCleanRequirementsOnRemove &&
+         !pTdnf->pArgs->nNoAutoRemove) ||
+        nAlterType == ALTER_AUTOERASE)
     {
         nFlags = nFlags | SOLVER_CLEANDEPS;
     }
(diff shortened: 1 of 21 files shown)


client/goal.c
```diff
@@ -403,7 +417,7 @@ TDNFGoal(
 
         if (nAlterType == ALTER_UPGRADE && dwExcludeCount != 0 && ppszExcludes)
         {
-            /* if we had packages to exclude, then we'd have diabled ones too */
+            /* if we had packages to exclude, then we'd have disabled ones too */
```
sshedi: Chose some other word than `disabled`, some language checking tool complains about this.

client/resolve.c
```diff
@@ -384,6 +383,12 @@ TDNFPrepareSinglePkg(
                       pSack,
                       queueGoal,
                       pszPkgName);
+        if (dwError == ERROR_TDNF_ALREADY_INSTALLED)
+        {
+            dwError = TDNFMarkAutoInstalledSinglePkg(pTdnf, pszPkgName);
+            BAIL_ON_TDNF_ERROR(dwError);
+            dwError = ERROR_TDNF_ALREADY_INSTALLED;
```
sshedi, line 394: Does control reach this line? after bailing out?

client/utils.c
```diff
@@ -759,3 +759,41 @@ TDNFGetCmdOpt(
...
+    char *pszAutoFile = NULL;
+
+    if (!pTdnf || !pppszAutoInstalled)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateStringPrintf(&pszAutoFile, "%s/%s/%s",
+            pTdnf->pArgs->pszInstallRoot,
```
sshedi, line 779: check if install root is "/", string formation might end up in "//.."

common/prototypes.h
```diff
@@ -259,6 +259,12 @@ uint32_t
 TDNFJoinPath(
     char **ppszPath, ...);
 
+uint32_t
+TDNFReadFileToStringArray(
+    const char *pszFile,
+    char ***pppszArray
```
sshedi, line 265: Can we use a double pointer instead?

etc/tdnf/tdnf.conf
```diff
@@ -1,6 +1,5 @@
 [main]
 gpgcheck=1
 installonly_limit=3
-clean_requirements_on_remove=true
```
sshedi, line 4: Make it false by default instead. Let the config entry be there.

pytests/repo/tdnf-test-cleanreq-required.spec
```diff
@@ -0,0 +1,34 @@
...
+
+%install
+mkdir -p %_topdir/%buildroot/lib/cleanreq/
+touch %_topdir/%buildroot/lib/cleanreq/required
+
+%files
+/lib/cleanreq/required
+
+%changelog
+*   Thu Nov 04 2021 Oliver Kurth <okurth@vmware.com> 1.0.1-3
```
sshedi: use 1.0.1-1 no need of other changelogs???

pytests/tests/test_autoremove.py
```diff
@@ -0,0 +1,183 @@
...
+# you may not use this file except in compliance with the License. The terms
+# of the License are located in the COPYING file of this distribution.
+#
+#   Author: Oliver Kurth <okurth@vmware.com>
+
+import os
+import shutil
+import pytest
+
+CONFDIR='/root/cleanreq'
```
sshedi: create this under /tmp

sshedi (changes requested):
Can you add one more test case please?

-> Two installed packages X & Y
-> Both require Z
-> Removing X should not remove Z
-> Removing y should remove  Z
-> Removing both X & Y together should remove Z

client/goal.c
oliverkurth: How about "disenabled"?

client/utils.c
oliverkurth: Thx - that's what `TDNFJoinPath()` is for.

common/prototypes.h
oliverkurth: I don't see an issue with triple pointers.

pytests/repo/tdnf-test-cleanreq-required.spec
oliverkurth: > -> Two installed packages X & Y
-> Both require Z
-> Removing X should not remove Z
-> Removing y should remove Z
-> Removing both X & Y together should remove Z

That's what `test_asutoremove2` does.

oliverkurth:
> -> Two installed packages X & Y
-> Both require Z
-> Removing X should not remove Z
-> Removing y should remove Z
-> Removing both X & Y together should remove Z

That's what `test_autoremove2` does.

client/goal.c
sshedi: `/* if we had packages to exclude, then we'd have inactive ones too */` should be fine.

client/resolve.c
oliverkurth: Yes. If the package is already installed, we want to mark it installed because it might have been automatically installed already, but now it's also wanted on its own. `TDNFMarkAutoInstalledSinglePkg` might return another error though, so we catch that first. After it's done successfully (and set `dwError` to `0`), we still want to return the original error.

pytests/repo/tdnf-test-cleanreq-required.spec
oliverkurth: It's just a test file. It doesn't really matter.

client/goal.c
```diff
@@ -456,6 +476,211 @@ TDNFGoal(
...
+                    break;
+                }
+            }
+        }
+        if (!ppInfo->ppszPkgsUserInstall || ppInfo->ppszPkgsUserInstall[j] == NULL)
+        {
+            fprintf(fp, "%s\n", ppszAutoInstalled[i]);
+        }
+    }
+    fclose(fp);
```
sshedi: move this fclose to cleanup.

sshedi (changes requested):
Minor change needed. Also, fix the copyright year in newly added test files.

PR #285, opened 2022-01-11
fix test_install_as_file_relpath1 test by changing to another directory

oliverkurth:
This test failed locally after merging PR #284 . Apparently the test was broken all along and worked accidentally. The PR mixed things around and made it fail.


sshedi (approved):
Yeah, my PR might have broken it. There were few duplicate function names. I integer suffix fun1, fun2 ... to overcome that.

oliverkurth:
> Yeah, my PR might have broken it. There were few duplicate function names. I integer suffix fun1, fun2 ... to overcome that.

It didn't break it but exposed a bug - I think the function was never run and therefore didn't raise an issue, but your change made it run. So that's a good thing.

PR #286, opened 2022-01-14
add test for tdnf lock

sshedi:


sshedi:
I have tried to keep the design simple for this test.

1. Thread 1 - gets blocked in install
2. Thread 2- runs a search command but waits for lock
3. Thread 3 - runs info command but waits for lock
4. Once all threads have started, send "no" to Thread 1's stdin via procfs
5. This makes Thread 1 end and give up lock
6. Now Thread 2 & 3 will finish and we will compare stdout of both of those and ensure that they were waiting for lock.

This was a tricky one to implement, took me a while to get it working.

PR #288, opened 2022-01-18
Add json output

oliverkurth:
This adds the `-j` option to add json output to a couple of commands. This is intended for integration with https://github.com/ssahani/pmd-nextgen .

~~Work in progress, not intended to be merged yet.~~


jsondump/jsondump.c
```diff
@@ -0,0 +1,333 @@
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+#include "stdarg.h"
+
+#include "jsondump.h"
+
+#define SIZE_INC 10
```
sshedi: Put this in jsondump.h

jsondump/jsondump.c
```diff
@@ -0,0 +1,333 @@
...
+        return NULL;
+
+    jd->buf = (char *)calloc(size, sizeof(char));
+    if (!jd->buf) {
+        jd_destroy(jd);
+        return NULL;
+    }
+    jd->buf_size = size;
+
+    jd->pos = 0;
```
sshedi: redundant, calloc does this by default

jsondump/jsondump.c
```diff
@@ -0,0 +1,333 @@
...
+    
+    return jd;
+}
+
+void jd_destroy(struct json_dump *jd)
+{
+    if (jd) {
+        if (jd->buf) free(jd->buf);
```
sshedi: put free(..) in next line

jsondump/jsondump.c
```diff
@@ -0,0 +1,333 @@
...
+        free(jd);
+    }
+}
+
+static
+int _jd_realloc(struct json_dump *jd, unsigned int add_size)
+{
+    if (jd->pos + add_size >= jd->buf_size-1) {
```
sshedi: nit: space for `jd->buf_size - 1`

jsondump/jsondump.c
```diff
@@ -0,0 +1,333 @@
...
+            return -1;
+        jd->buf_size = jd->buf_size + add_size;
+    }
+    return 0;
+}
+
+char *jsonify_string(const char *str)
+{
+    char *p;
+    const char *q = str;
```
sshedi, line 127: we can use str directly, no need of q

jsondump/jsondump.c
```diff
@@ -0,0 +1,333 @@
...
+    char *buf;
+
+    buf = calloc(strlen(str)*2+1, sizeof(char));
+    if(!buf)
+        return NULL;
+
+    p = buf;
+    while(*q) {
```
sshedi, line 137: use `strspn` ; get length of accept string and loop length times and do the operation.

jsondump/jsondump.c
```diff
@@ -0,0 +1,333 @@
...
+ 
+    free(json_value);
+
+    return 0;
+}
+
+int jd_map_add_int(struct json_dump *jd, char *key, int value)
+{
+    int l;
+    int add_size = strlen(key) + 22 + 4;
```
sshedi: what is this magic number `22+4`?

sshedi: use a common function for jp_map_add_type and use switch case

jsondump/jsondump.c
```diff
@@ -0,0 +1,328 @@
...
+    _jd_map_prep_append(jd);
+
+    int l = snprintf(&(jd->buf[jd->pos]), jd->buf_size - jd->pos, "\"%s\":%s}", key, jd_child->buf);
+    jd->pos += l;
+    return 0;
+}
+
+int jd_list_start(struct json_dump *jd)
+{
+    _jd_realloc(jd, 2);
```
sshedi: validate _jd_realloc ret value

jsondump/test.c
```diff
@@ -0,0 +1,70 @@
...
+    printf("%s\n", jd1->buf);
+    
+    jd_destroy(jd);
+    jd_destroy(jd1);
+    
+    jd = jd_create(10);
+    jd_list_start(jd);
+    for(int i = 0; i < 10; i++) {
+        char buf[3];
+        snprintf(buf, 3, "%d", i);
```
sshedi: init buf to 0 before using

s/3/sizeof(buf)

jsondump/test.c
```diff
@@ -0,0 +1,70 @@
...
+
+    jd_map_start(jd1);
+    jd_map_add_child(jd1, "nested", jd);
+    
+    printf("%s\n", jd1->buf);
+    
+    jd_destroy(jd);
+    jd_destroy(jd1);
+    
+    jd = jd_create(10);
```
sshedi: use a macro for 10 and in loop below

jsondump/test.c
```diff
@@ -0,0 +1,70 @@
...
+    jd_destroy(jd);
+
+    jd = jd_create(10);
+    jd_list_start(jd);
+    jd_list_add_bool(jd, 1);
+    jd_list_add_bool(jd, 0);
+
+    printf("%s\n", jd->buf);
+    
+    jd_destroy(jd);
```
sshedi, line 99: return 0 at end

sshedi (changes requested):
First review, will review again later.

Can we have a wrapper for pr_crit & jsondump calls in a common place and if jsondump flag is set, then handle json operations there?

jsondump/jsondump.c
oliverkurth: Why? It's not used anywhere outside of this file, and shouldn't be.

jsondump/jsondump.c
oliverkurth: We could, but it IMHO it looks better this way. Generally, I don't like changing arguments inside a function.

jsondump/jsondump.c
oliverkurth: 22 is the max length of an 64 bit integer.

jsondump/jsondump.c
oliverkurth: How does that help? The loop has to iterate through the whole string.

jsondump/test.c
oliverkurth: It's not the same `10` though. The arg for `jd_create()` is the initial buffer size. The `10` in the `for` loop is just the number of elements we want in the list.

jsondump/jsondump.c
oliverkurth: Done.

jsondump/jsondump.c
oliverkurth: Done, here and other places.

oliverkurth:
> Can we have a wrapper for pr_crit & jsondump calls in a common place and if jsondump flag is set, then handle json operations there?

That won't work. We cannot just print json without context.

oliverkurth:
Let's move this out of the draft state.

jsondump/jsondump.c
oliverkurth: Done

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+ *
+ * Licensed under the GNU General Public License v2 (the "License");
+ * you may not use this file except in compliance with the License. The terms
+ * of the License are located in the COPYING file of this distribution.
+ */
+
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+#include "stdarg.h"
```
sshedi: nit: use angular brackets

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+        if (jd->buf)
+            free(jd->buf);
+        free(jd);
+    }
+}
+
+static
+int _jd_realloc(struct json_dump *jd, unsigned int add_size)
+{
+    if (jd->pos + add_size >= jd->buf_size - 1) {
```
sshedi, line 107: validate jd for null

jsondump/jsondump.h
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (C) 2022 VMware, Inc. All Rights Reserved.
+ *
+ * Licensed under the GNU General Public License v2 (the "License");
+ * you may not use this file except in compliance with the License. The terms
+ * of the License are located in the COPYING file of this distribution.
+ */
+
+
```
sshedi, line 10: Add header guard or pragma once

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+
+#include "jsondump.h"
+
+#define SIZE_INC 10
+
+struct json_dump *jd_create(unsigned int size)
+{
+    struct json_dump *jd = NULL;
+
+    if (size == 0)
```
sshedi: do this after calloc at line 25

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+#include "stdarg.h"
+
+#include "jsondump.h"
+
+#define SIZE_INC 10
+
+struct json_dump *jd_create(unsigned int size)
```
sshedi, line 69: all the calls to jd_create is done with size as 0.
Why?

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+void jd_destroy(struct json_dump *jd)
+{
+    if (jd) {
+        if (jd->buf)
+            free(jd->buf);
+        free(jd);
+    }
+}
+
+static
```
sshedi, line 101: declare these static functions at the top.

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+            return -1;
+        jd->buf_size = jd->buf_size + add_size;
+    }
+    return 0;
+}
+
+char *jsonify_string(const char *str)
+{
+    char *p;
+    const char *q = str;
```
sshedi, line 127: use str directly, why another variable q?

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+            case '\t':
+                *p++ = '\\';
+                *p++ ='t';
+                break;
+            default:
+                *p++ = *q;
+        }
+        q++;
+    }
+    *p = 0;
```
sshedi, line 173: needed? it is allocated using calloc

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+    jd->buf[jd->pos++] = '}';
+    jd->buf[jd->pos] = 0;
+
+    return 0;
+}
+
+static
+void _jd_map_prep_append(struct json_dump *jd)
+{
+    if (jd->pos > 1 && jd->buf[jd->pos-1] == '}') {
```
sshedi, line 194: check jd for null, maybe need to return a value if jd is null

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+static
+int _jd_map_add_raw(struct json_dump *jd, const char *key, const char *value)
+{
+    int add_size = strlen(key) + strlen(value) + 5;
+
+    if(_jd_realloc(jd, add_size))
+        return -1;
+    _jd_map_prep_append(jd);
+
+    int l = snprintf(&(jd->buf[jd->pos]), jd->buf_size - jd->pos, "\"%s\":%s}", key, value);
```
sshedi: declare all variables at top. I don't like declaring variable in between like this

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+    jd->pos += l;
+    return 0;
+}
+
+int jd_map_add_string(struct json_dump *jd, const char *key, const char *value)
+{
+    if (!value)
```
sshedi, line 233: declare all variable at top

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+        return jd_map_add_null(jd, key);
+
+    char *json_value = jsonify_string(value);
+    int add_size = strlen(key) + strlen(json_value) + 6;
```
sshedi: check json_value for null

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+
+    int l = snprintf(&(jd->buf[jd->pos]), jd->buf_size - jd->pos, "\"%s\":%d}", key, value);
+    jd->pos += l;
+
+    return 0;
+}
+
+int jd_map_add_bool(struct json_dump *jd, const char *key, int value)
+{
+    return _jd_map_add_raw(jd, key, value ? "true": "false");
```
sshedi, line 258: check jd for null

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+
+int jd_map_add_null(struct json_dump *jd, const char *key)
+{
+    return _jd_map_add_raw(jd, key, "null");
+}
+
+int jd_map_add_fmt(struct json_dump *jd, const char *key, const char *format, ...)
+{
+    va_list args;
+    char buf[256];
```
sshedi: init this buffer to {0}

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+    
+    va_start(args, format);
+    vsnprintf(buf, 255, format, args);
```
sshedi: s/255/sizeof(buf)-1

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+        if (jd->buf[jd->pos-1] != '[')
+            jd->buf[jd->pos++] = ',';
+    }
+    return 0;
+}
+
+static
+int _jd_list_add_raw(struct json_dump *jd, const char *value)
+{
+    int add_size = strlen(value) + 3;
```
sshedi: add comment for this math, why +3

jsondump/jsondump.c
```diff
@@ -0,0 +1,344 @@
...
+    jd->pos += l;
+
+    free(json_value);
+
+    return 0;
+}
+
+int jd_list_add_int(struct json_dump *jd, int value)
+{
+    int add_size = 22 + 3;
```
sshedi: same here, 22 + 3, document it

pytests/tests/test_json.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def test_info(utils):
+    ret = utils.run(['tdnf', '-j', 'info'])
+    infolist = json.loads("\n".join(ret['stdout']))
+
+    glibc_found = False
+    for info in infolist:
+        if info['Name'] == "glibc":
+            glibc_found = True
```
sshedi, line 60: just return instead of break, if we go out of loop, then assert false.
Applies in all places, generic comment
```
    for info in infolist:
        if info['Name'] == "glibc":
            return
    assert(False)
```

tools/cli/lib/api.c
```diff
@@ -510,86 +648,181 @@ TDNFCliRepoQueryCommand(
...
-        {
-            for (j = 0; pPkgInfo->ppszFileList[j]; j++);
-            nCount += j;
-        }
-        else if (pPkgInfo->pChangeLogEntries)
-        {
-            PTDNF_PKG_CHANGELOG_ENTRY pEntry;
-            for (pEntry = pPkgInfo->pChangeLogEntries; pEntry; pEntry = pEntry->pNext)
+            int j;
+            struct json_dump *jd_pkg = jd_create(0);
```
sshedi: check return value

tools/cli/lib/api.c
```diff
@@ -510,86 +648,181 @@ TDNFCliRepoQueryCommand(
...
+
+            pPkgInfo = &pPkgInfos[i];
+
+            jd_map_add_fmt(jd_pkg, "Nevra", "%s-%s-%s.%s",
+                                    pPkgInfo->pszName,
+                                    pPkgInfo->pszVersion,
+                                    pPkgInfo->pszRelease,
+                                    pPkgInfo->pszArch);
+
+            jd_map_add_string(jd_pkg, "Name", pPkgInfo->pszName);
```
sshedi: check retval

tools/cli/lib/updateinfocmd.c
```diff
@@ -120,23 +119,37 @@ TDNFCliUpdateInfoSummary(
...
+                pr_crit(
+                    "%d %s notice(s)\n",
+                    pSummary[i].nCount,
+                    TDNFGetUpdateInfoType(pSummary[i].nType));
+            }
+        }
+        if (nCount == 0)
+        {
+            pr_crit(
+                "\n%d updates.\n", nCount);
```
sshedi: why new line?

tools/cli/main.c
```diff
@@ -205,13 +212,30 @@ TDNFCliPrintError(
...
+            jd_map_add_string(jd, "ErrorMessage", pszError);
+            
+            pr_json(jd->buf);
+            jd_destroy(jd);
+        }
     }
     else
     {
-        pr_err("%s\n", pszError);
+        if (dwErrorCode)
```
sshedi, line 232:
```
        if (dwErrorCode)
        {
            pr_err("Error(%u) : ", dwErrorCode);
        }
        pr_err("%s\n", pszError);
```

jsondump/test.c
```diff
@@ -0,0 +1,80 @@
+/*
```
sshedi, line 1: can we move this into pytests?
create this C file from a python script and and run it and check exist status

sshedi (changes requested):
This PR in general needs a lot of fine tuning with validating return values and handling error conditions.

And in jsondump.c it seems to me that there is a lot of code duplication.

For example:
```
int jd_list_add_bool(struct json_dump *jd, int value)
{
    return _jd_list_add_raw(jd, value? "true" : "false");
}

int jd_map_add_bool(struct json_dump *jd, const char *key, int value)
{
    return _jd_map_add_raw(jd, key, value ? "true": "false");
}
```

I think we can improve this.

And most importantly, can you please add API documentation on how to use APIs in jsondump.c?

For example:
In function `jsonify_string` the return value is a memory allocated character pointer which has to be freed by caller function. Such things need to be clearly documented.

jsondump/jsondump.c
oliverkurth: Ooops!

jsondump/jsondump.c
oliverkurth: 0 uses the default, which is defined by `SIZE_INC`. Initial idea was that if you know it's going to be larger you can set a bigger size to safe reallocations.

jsondump/jsondump.c
oliverkurth: Done.

jsondump/jsondump.c
oliverkurth: Done

jsondump/jsondump.c
oliverkurth: I don't like modifying an argument. I think it's bad style.

jsondump/jsondump.c
oliverkurth: It's not needed but it's safer. Maybe one day we decide to use a buffer on the stack, or use `malloc` and then we miss setting this to `0`.

jsondump/jsondump.c
oliverkurth: Converted it allocate a buffer.

jsondump/jsondump.c
oliverkurth: Done

jsondump/jsondump.c
oliverkurth: Changed it to use a dynamically allocated buffer

jsondump/jsondump.c
oliverkurth: `_jd_addr_map_raw()` does the check now, similar for other places.

jsondump/jsondump.c
oliverkurth: Done.

jsondump/jsondump.c
oliverkurth: Done.

jsondump/jsondump.c
oliverkurth: `_jd_map_add_raw()` (which is calling this) does this now.

jsondump/jsondump.c
oliverkurth: No. They are intentionally defined before being used.

pytests/tests/test_json.py
oliverkurth: This makes it less obvious when the assert fails.

tools/cli/lib/updateinfocmd.c
oliverkurth: I don't know, fixed.

jsondump/test.c
oliverkurth: Good idea, done (not exactly as suggested).

jsondump/jsondump.c
```diff
@@ -0,0 +1,397 @@
...
+
+    /* Determine required size */
+
+    size = vsnprintf(p, size, fmt, ap);
+
+    if (size < 0)
+        return NULL;
+
+    size++;             /* For '\0' */
+    p = malloc(size);
```
sshedi: use calloc

tools/cli/lib/installcmd.c
```diff
@@ -275,6 +283,132 @@ TDNFCliAlterCommand(
...
+    for(pPkgInfo = pPkgInfos; pPkgInfo; pPkgInfo = pPkgInfo->pNext)
+    {
+        struct json_dump *jd_pkg = jd_create(0);
+        CHECK_JD_NULL(jd_pkg);
+
+        CHECK_JD_RC(jd_map_start(jd_pkg));
+
+        CHECK_JD_RC(jd_map_add_string(jd_pkg, "Name", pPkgInfo->pszName));
+        CHECK_JD_RC(jd_map_add_string(jd_pkg, "Arch", pPkgInfo->pszArch));
+        CHECK_JD_RC(jd_map_add_fmt(jd_pkg, "Evr", "%s-%s", pPkgInfo->pszVersion, pPkgInfo->pszRelease));
```
sshedi, line 307: We need to destroy created objects upon error.

sshedi (changes requested):
Also, exclude "test" binary file from your PR

sshedi:
Run test.c executable with valgrind enabled and check for leaks please.

tools/cli/lib/api.c
```diff
@@ -159,47 +166,80 @@ TDNFCliListCommand(
...
+            CHECK_JD_RC(jd_map_add_string(jd_pkg, "Name", pPkg->pszName));
+            CHECK_JD_RC(jd_map_add_string(jd_pkg, "Arch", pPkg->pszArch));
+            CHECK_JD_RC(jd_map_add_fmt(jd_pkg, "Evr", "%s-%s", pPkg->pszVersion, pPkg->pszRelease));
+            CHECK_JD_RC(jd_map_add_string(jd_pkg, "Repo", pPkg->pszRepoName));
+
+            CHECK_JD_RC(jd_list_add_child(jd, jd_pkg));
+            jd_destroy(jd_pkg); jd_pkg = NULL;
         }
+        pr_json(jd->buf);
+        jd_destroy(jd); jd = NULL;
```
sshedi: new line please or use a macro.

oliverkurth: I like it this way. It's clear and doesn't waste space.

jsondump/jsondump.c
oliverkurth: Done

tools/cli/main.c
oliverkurth: Thanks, but I don't like it. Makes it too obscure what's happening.

include/tdnf-common-defines.h
```diff
@@ -33,12 +33,36 @@
...
+    if ((jd) == NULL)                  \
+    {                                  \
+        dwError = ERROR_TDNF_JSONDUMP; \
+        BAIL_ON_CLI_ERROR(dwError);    \
+    }                                  \
+}
+
+#define JD_SAFE_DESTROY(jd) \
+{                           \
+    if (jd) jd_destroy(jd); \
```
sshedi:
```
if (jd) {
  jd_destroy(jd);
  jd = NULL;
}
```
saves an unnecessary write

sshedi (approved):

include/tdnf-common-defines.h
oliverkurth: Oh, good catch.

PR #289, opened 2022-01-19
fix copyright headers

sshedi:


sshedi:
Done in https://github.com/vmware/tdnf/pull/294

PR #291, opened 2022-01-20
Move instance running check to TDNFOpenHandle()

sshedi:
This will ensure that any tdnf api usage will create a instance lock file

Implement a new utility for locking.


oliverkurth:
You have two PRs with the same title - this and #290 - which is the correct one?

sshedi:
> You have two PRs with the same title - this and #290 - which is the correct one?

One is for stable-3.1 & other is for stable-3.2 :)

PR #294, opened 2022-02-05
Fix copyright years

oliverkurth:
Change copyright years to the latest year (as from the "CommitDate" in `git log --pretty=fuller`) from the file's commit history.

The script can be found here: https://gist.github.com/oliverkurth/3dced0b2bee94483762bc9d3216868ee . Note that this PR, if merged, would make files modified this year, so future versions of the script need to account for that and filter out this commit.


sshedi:
Can you add the automation script to scripts/ directory? it will be of use in future.

oliverkurth:
> Can you add the automation script to scripts/ directory? it will be of use in future.

Okay, done.

tapakund (approved):

sshedi (approved):

PR #297, opened 2022-02-11
add tox.ini for flake8 options

oliverkurth:
Use tox.ini for flake8 options. Ignore Apple file forks.

### DIFF
--- a/ci/docker-entrypoint.sh
+++ b/ci/docker-entrypoint.sh
@@ -8,7 +8,7 @@ cmake .. && make -j32 && make python -j32
 
 make check -j32
 
-if ! flake8 --ignore E501 ../pytests; then
+if ! flake8 ../pytests ; then
   echo "flake8 tests failed"
   exit 1
 fi
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,7 @@ core*
 
 # vi backup
 *.swp
+
+# Apple forks
+._*
+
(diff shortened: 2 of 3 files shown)


ci/docker-entrypoint.sh
```diff
@@ -8,7 +8,8 @@ cmake .. && make -j32 && make python -j32
 
 make check -j32
 
-if ! flake8 --ignore E501 ../pytests; then
+cd pytests
```
sshedi: No need of this. If we add more  py scripts in some other location, we need to cd to it again which will make it confusing unnecessarily. Either use the method which is already there or explore for an option which can be added tox.ini for directories we want to run this check.

sshedi (changes requested):

ci/docker-entrypoint.sh
oliverkurth: I don't see an issue with that at all. Anyway, I moved `tox.ini` to the top, this way we can use `../pytests` in the command line.

For some reason, `flake8` does not read `tox.ini` in `../pytests` when called with `../pytests`, but it does work if it's in the current directory. That is why I had to make this change.

sshedi:
```For some reason, flake8 does not read tox.ini in ../pytests when called with ../pytests, but it does work if it's in the current directory. That is why I had to make this change.```
I wasn't aware of this. NP, changes LGTM.

sshedi (approved):

PR #299, opened 2022-02-18
build RPMs in github workflow

oliverkurth:
Should make it easy to test PRs.

RPMS can be downloaded by following the "Details" link of the workflow, then "Summary". There is no easy way to do this with curl, unfortunately.

When downloaded, create a directory and unzip:
```
mkdir repo
pushd repo
unzip ../tdnf-rpms.zip
popd
```
Then install with:
`tdnf --repofrompath=tdnf,$(pwd)/repo --repoid=tdnf install tdnf`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,14 +10,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.0 FATAL_ERROR)
 
 project(tdnf VERSION 3.3.0 LANGUAGES C)
 set(VERSION ${PROJECT_VERSION})
-set(PROJECT_YEAR 2021)
-
-# needed for spec file
-execute_process (
-    COMMAND bash -c "uname -r | cut -d'.' -f4"
-    OUTPUT_VARIABLE dist
-)
-set(DIST .${dist})
+set(PROJECT_YEAR 2022)
 
 # Ensure correct directory paths are used for installation
 include(GNUInstallDirs)
--- a/ci/build-rpms.sh
+++ b/ci/build-rpms.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+rm -rf build
+mkdir -p build
+
+pushd build
+cmake .. && ../scripts/build-tdnf-rpms
+popd
+
+createrepo rpms
--- a/.github/workflows/build-tdnf-rpms.yml
+++ b/.github/workflows/build-tdnf-rpms.yml
@@ -1,0 +1,24 @@
+name: tdnf CI
+
+on: [pull_request, push, workflow_dispatch]
+
+jobs:
+    photon-rpms:
+        env:
+            DIST: photon
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - name: build container
+              working-directory: ${{ github.workspace }}
+              run: |
+                docker build -t ${DIST}/tdnf-build -f ci/Dockerfile.${DIST} .
+            - name: build rpms in container
+              working-directory: ${{ github.workspace }}
+              run: |
+                docker run --rm -e DIST -v$(pwd):/build -w/build ${DIST}/tdnf-build ./ci/build-rpms.sh
+            - name: upload RPMs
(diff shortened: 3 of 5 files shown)


sshedi:
Is this really needed? We can build rpms(manually) and test them when required.
This will increase the workflow ci/cd run time.

Building rpms locally is much easier than navigating, downloading and extracting zip file.

oliverkurth:
Workflows run in parallel. When this one is done the others CI workflows are still running, so there is no increase in run time.

I think it's easier to quickly download and test RPMs from someone else's PR and test, without checking it out and build first.

Alsp, this workflow will test the RPM build - I had to spend some time to make it working. And if changes to the spec file are needed we would catch it early, before integrating into Photon. It's also easier with this PR to build RPMs locally now - just do `docker run --rm -e DIST -v$(pwd):/build -w/build ${DIST}/tdnf-build ./ci/build-rpms.sh`.

Another reason is that I created a similar workflow in https://github.com/ssahani/pmd-nextgen/pull/39 . It needs a tdnf version from a special branch.

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set(PROJECT_YEAR 2021)
 
 # needed for spec file
 execute_process (
-    COMMAND bash -c "uname -r | cut -d'.' -f4"
+    COMMAND bash -c "rpm -q glibc | cut -d'.' -f3"
```
sshedi: What's the rationale behind this? I used `uname` because it's available in our docker image by default where as rpm is not.

ci/build-rpms.sh
```diff
@@ -0,0 +1,11 @@
...
+
+rm -rf build
+mkdir -p build
+
+pushd build
+cmake .. && ../scripts/build-tdnf-rpms
+popd
+
+createrepo rpms
+
```
sshedi: nit: remove empty line at end

.github/workflows/rpm.yml
```diff
@@ -0,0 +1,24 @@
+name: tdnf CI
```
sshedi: `.github/workflows/rpm.yml` rename this to `build-tdnf-rpms.yml`

ci/build-rpms.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+rm -rf build
```
sshedi, line 3: Do we need to use different output dir name? `build` is a bit confusing, we are using this name for everything.

I don't know, please give some insights on this.

CMakeLists.txt
oliverkurth: But `uname -r` returns the release of the host. If you run a Photon 3.0 image in a Photon 4.0 host you get `4.0`. Even worse, when running on a Ubuntu host (as this workflow does), you get an empty string. That's how I noticed.

sshedi: Excellent catch. Yeah, this change makes sense. Thanks.

oliverkurth: I see though that this would break `cmake` when trying to build when `rpm` isn't installed. Maybe we should put the determination of the release tag into `ci/build-rpms.sh` and pass it to `rpmbuild` as a parameter.

ci/build-rpms.sh
oliverkurth: I'd like to keep the top dir clean. `build` is used by the CI build too. Here it's actually only used to call `cmake` to fill in the variables in the `*.in` files needed.

CMakeLists.txt
sshedi: Or we can parse /etc/photon-release file

sshedi (approved):

CMakeLists.txt
oliverkurth: `/etc/photon-release` would be Photon specific, but it should work for Fedora and other rpm based distros too. Also, it does not have the suffix. `/etc/os-release` does not work either, unless we would map `ID` and `VERSION`, but that would not be generic.

I am open to other suggestions. `glibc` is installed on most OSes. Should a use case come up that does not have it we can add whatever they use instead.

PR #300, opened 2022-02-25
fix reinstall for packages on the command line

oliverkurth:
When trying to reinstall a package on the command line, `tdnf` exited claiming "Nothing to do". Fixing this.

### DIFF
--- a/pytests/tests/test_rpm_cmdline.py
+++ b/pytests/tests/test_rpm_cmdline.py
@@ -156,3 +156,21 @@ def test_install_same_as_filname(utils):
     ret = utils.run(['tdnf', 'install', '-y', '--nogpgcheck', pkgname])
     assert(ret['retval'] == 0)
     assert(utils.check_package(pkgname))
+
+
+# test "tdnf reinstall /path/to/pkg.rpm". See PR #300.
+def test_reinstall_as_file(utils):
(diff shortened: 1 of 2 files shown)


pytests/tests/test_rpm_cmdline.py
```diff
@@ -156,3 +156,20 @@ def test_install_same_as_filname(utils):
...
+    # prepare by installing package
+    ret = utils.run(['tdnf', 'install', '-y', '--nogpgcheck', path])
+    assert(ret['retval'] == 0)
+    assert(utils.check_package(pkgname))
+
+    # actual test
+    ret = utils.run(['tdnf', 'reinstall', '-y', '--nogpgcheck', path])
+    assert(ret['retval'] == 0)
+    assert(utils.check_package(pkgname))
+    assert("Nothing to do" not in ret['stderr'])
```
sshedi: also check for "Installing ..." string in stdout. One more verification to enure it got installed again

sshedi (approved):

pytests/tests/test_rpm_cmdline.py
oliverkurth: Thanks. I also forgot that `ret['stdout']` and `ret['stderr']` are lists, not strings. I wonder why this only broke in Photon 3.0 though.

PR #306, opened 2022-03-29
simplify argument handling

oliverkurth:
Make argument parsing a little more coherent and simpler; and remove dead code. This is to prepare for more changes to add plugins with their own arguments.

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2021 VMware, Inc. All Rights Reserved.
+ * Copyright (C) 2015-2022 VMware, Inc. All Rights Reserved.
  *
  * Licensed under the GNU Lesser General Public License v2.1 (the "License");
  * you may not use this file except in compliance with the License. The terms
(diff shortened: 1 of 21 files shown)


client/repolist.c
```diff
@@ -609,18 +609,23 @@ TDNFRepoListFinalize(
...
     {
-        if(pSetOpt->nType == CMDOPT_ENABLEREPO ||
-           pSetOpt->nType == CMDOPT_DISABLEREPO)
+        if(strcmp(pSetOpt->pszOptName, "enablerepo") == 0)
+        {
+            dwError = TDNFAlterRepoState(
+                          pTdnf->pRepos,
+                          1,
+                          pSetOpt->pszOptValue);
+            BAIL_ON_TDNF_ERROR(dwError);
```
sshedi: put this BAIL_ON_TDNF_ERROR at the end of loop after all if else conditions, it is common in all cases

client/repolist.c
```diff
@@ -654,25 +657,20 @@ TDNFRepoListFinalize(
...
             if(pRepo->pszBaseUrl)
             {
                 dwError = TDNFConfigReplaceVars(pTdnf, &pRepo->pszBaseUrl);
                 BAIL_ON_TDNF_ERROR(dwError);
             }
-
             if(pRepo->pszMetaLink)
             {
                 dwError = TDNFConfigReplaceVars(pTdnf, &pRepo->pszMetaLink);
                 BAIL_ON_TDNF_ERROR(dwError);
```
sshedi, line 666: same here

sshedi (approved):
LGTM, minor comments.

client/repolist.c
oliverkurth: Good point here.

client/repolist.c
oliverkurth: Not here. This is no `else if`, these lines will be executed in sequence. There may be `dwError` set in the first call to `TDNFConfigReplaceVars`, but this would get overwritten with the next one.

PR #307, opened 2022-04-02
remove trailing ws

oliverkurth:
The command is
```
for f in $(git ls-files) ; do sed -i 's/[ \t]*$//' $f ; done
```


oliverkurth:
trivial, no need to review

PR #308, opened 2022-04-13
do not mark packages as auto installed if only downloading

oliverkurth:
Packages should not be marked auto installed when we are just downloading.


sshedi:
can we add a test for this ?

sshedi (approved):

oliverkurth:
> can we add a test for this ?

Yes, we could. However I am thinking about refactoring the autoremove feature while I work on implementing transaction history. Testing this is not high priority, it's a corner case with little impact: if a package is marked 'autoinstalled' when in fact we are just downloading there is no harm when the package isn't installed. If it already is installed, the status won't be changed.

PR #314, opened 2022-05-06
add history tracking and rollbacks

oliverkurth:
This adds the history tracking and rollback feature. Every transaction will be recorded in a database. The transactions can be rolled back to a past state, or a range of transactions can be undone or redone.

There are five new sub commands to the new `history` command:

### history init/update

This will initialize the history database, and should be called just after tdnf is installed. Any altering command (`install`, `erase`, etc) will also initialize the database if it isn't already.

If called when the database is already initialized, this will add a transaction if any package has been added or removed by another application (rpm command) since the last time a transaction was recorded. Otherwise this will have no effect.

The subcommands `init` and `update` are the same, but this may change in the future.

### history list

This lists the history. Also, the command `history` without any sub command will do the same. It understands these options:

`--info` will list the history in more detail, including packages that were added/removed
`--reverse` will list the history in reverse order
`--from <id>` and `--to <id>` will list a range of transaction ids.

Example output:
```
# tdnf history
ID   cmd line                                 date/time             +added / -removed
   1 (set)                                    Thu May 05 2022 19:14 +152 / -0
   2 -y install less                          Thu May 05 2022 19:14 +1 / -0
   3 -y install lsof                          Thu May 05 2022 19:18 +2 / -0
```
```
# tdnf history --info --from 2 --to 3
ID   cmd line                                 date/time             +added / -removed
   2 -y install less                          Thu May 05 2022 19:14 +1 / -0
added: less-551-2.ph4.aarch64

   3 -y install lsof                          Thu May 05 2022 19:18 +2 / -0
added: libtirpc-1.2.6-2.ph4.aarch64, lsof-4.91-1.ph4.aarch64
```

### history rollback

This will revert to a previous state. The desired state must be specified with `--to`. Example:
```
# tdnf history rollback --to 49

Upgrading:
curl-devel                               aarch64              7.82.0-3.ph4                photon-updates       885.16k 906404
curl                                     aarch64              7.82.0-3.ph4                photon-updates       256.73k 262896
...

Total installed size:   3.52M 3688748
Is this ok [y/N]: y

Downloading:
curl-devel                              793306 100%
curl                                    148725 100%
...
Testing transaction
Running transaction
Installing/Updating: rpm-libs-4.16.1.3-9.ph4.aarch64
Installing/Updating: rpm-4.16.1.3-9.ph4.aarch64
...
Complete!
```

### history undo

This reverts a transaction. The parameter `--from` is mandatory, and the specified transaction will be reverted. Optionally, with the parameter `--to` a range can be specified.  Note that the range is inclusive - for example if the range is given as 2 to 4, the actions done in 2,3 and 4 will be undone.

### history redo

Similar to `undo`, but the actions will be done again.

#### Deltas

The history commands to change history deal with the total deltas between the start and target states. For each range of transactions, the intermediate states are irrelevant. For example, if one transaction installs a package and another removes it, the installed state of that package from start to end will be the same. Therefore any history command across that range will not reinstall or remove that package. If intermediate states are desired, each transaction needs to be executed one by one.

#### Unresolved Packages

If a package cannot be found, `tdnf` will fail with an error message. For example, when rolling back to a state before an update, the installed packages may no longer be available in the repository. In that case, it may be possible to revert by enabling additional repositories. Example:
```
# tdnf history rollback --to 1
The following packages could not be resolved:

curl-libs-7.82.0-1.ph4.aarch64
rpm-libs-4.16.1.3-7.ph4.aarch64
...

The package(s) may have been moved out of the enabled repositories since the
last time they were installed. You may be able to resolve this by enabling
additional repositories.
Error(1011) : No matching packages
```
```
tdnf --enablerepo=photon history rollback --to 1

Downgrading:
curl-devel                               aarch64              7.82.0-1.ph4                photon               885.16k 906404
rpm-build                                aarch64              4.16.1.3-7.ph4              photon               434.00k 444418
...

Total installed size:   4.26M 4463905

Removing:
wget                                     aarch64              1.21.3-1.ph4                @System                3.02M 3168291
tdnf-test-cleanreq-required              aarch64              1.0.1-3                     @System                    0.00b 0
lsof                                     aarch64              4.91-1.ph4                  @System              202.36k 207218
libtirpc                                 aarch64              1.2.6-2.ph4                 @System              193.33k 197970
gdb                                      aarch64              10.1-2.ph4                  @System               12.60M 13214814

Total installed size:  16.01M 16788293
Is this ok [y/N]: 
```

#### Transactions outside tdnf

`tdnf` will keep track of its own transactions. However, packages may be removed or added with other tools, like `rpm`. In this case, the next time `tdnf` does a transaction this will be detected and recorded as a pseudo transaction:
```
# tdnf history --info --from 49 --to 49
ID   cmd line                                 date/time.            +added / -removed
  49 (unknown)                                Thu May 05 2022 23:38 +1 / -0
added: gdb-10.1-2.ph4.aarch64
```

#### Dependencies

For the actions `undo` and `redo`, it may be necessary to install additional dependencies, not just the packages that were changed in a transaction. For example, if a transaction that installed a single package (which was later removed) is redone, but in the meantime one of its dependencies was removed, the dependencies will be attempted to be installed again.

This is not an issue for `rollback` because the entire set of packages will be restored (assuming that dependencies were satisfied at that state).

#### Database Size

A full install if the Photon `minimal` package, pulling 173 packages in total, creates a db with a size of 28672 bytes, 165 bytes per package. Each transaction should add about that amount per package, but there are various factors like adding a new package (space for the new name), length of the command line etc. influencing the size added. However assuming 165 bytes per package per update means that each package can be upgraded 35 times before the database reaches 1MB in size. This a far more than can be expected during the lifetime of a release.

### API and DB

This change adds an internal API to keep track of history, by adding a small library (in the `history` sub directory). The history is tracked in an sqlite3 database. All access to this db is capsuled inside the library. The library does not do any rpm transactions itself.

Every access to the history needs a context object that bis created with `create_history_ctx()`. Optionally, an rpm context can be added to this with `history_set_rpmts()`, which is needed if access to the rpm db is needed (for reading). The rpm db needs to have been opened using `rpmtsOpenDB()`.

#### Updating State

The function `history_sync()` should be called before any rpm transaction is run. It will detect if the state of installed rpms has been changed in respect to what is recorded in the history db, and will update the db if necessary. After that, the context should reflect the current state.

After a successful rpm transaction, the function `history_update_state()` should be called to update the db to the new state.

When everything is done, `destroy_history_ctx()` should be called to close the history db and free resources.

#### Changing History

After calling `history_sync()` specific deltas can then be accessed through `history_get_delta()` and `history_get_delta_range()`. This will give a list of "deltas", that is a lists of added and removed packages from one transaction id to another (this can be across multiple transactions). This can be used to roll back to a previous state, or undo/redo transactions.

#### Listing History

There is no need to call `history_sync()`. `history_get_transactions()` will get a list of transactions for display. The difference to `history_get_delta_range()/history_get_delta()` is that this gives a list of changed packages for each transaction in the range, and more detailed information (like the command that was run).


client/api.c
```diff
@@ -1904,6 +1904,347 @@ TDNFUpdateInfo(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (rpmtsOpenDB(ts, O_RDONLY))
+    {
+        dwError = ERROR_TDNF_RPMTS_OPENDB_FAILED;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if(rpmtsSetRootDir (ts, pTdnf->pArgs->pszInstallRoot))
```
sshedi: nit: delete space before `(`

client/api.c
```diff
@@ -1904,6 +1904,347 @@ TDNFUpdateInfo(
...
+    {
+        dwError = ERROR_TDNF_HISTORY_ERROR;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (pHistoryArgs->nCommand == HISTORY_CMD_INIT)
+    {
+        goto cleanup;
+    }
+    else if (pHistoryArgs->nCommand == HISTORY_CMD_ROLLBACK)
```
sshedi: no need of else, use if. First if is using `goto`

client/api.c
```diff
@@ -1904,6 +1904,347 @@ TDNFUpdateInfo(
...
+        {
+            dwError = ERROR_TDNF_INVALID_PARAMETER;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+    }
+
+    dwError = TDNFGetHistoryCtx(pTdnf, &ctx, 1);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    history_get_transactions(ctx, &tas, &count,
```
sshedi: Check retval

client/goal.c
```diff
@@ -469,14 +412,157 @@ TDNFGoal(
...
+        if (pQueuePkgList->count == 0)
+        {
+            dwError = ERROR_TDNF_ALREADY_INSTALLED;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+
+        for (int i = 0; i < pQueuePkgList->count; i++)
+        {
+            Id dwId = pQueuePkgList->elements[i];
+            TDNFAddGoal(pTdnf, nAlterType, &queueJobs, dwId,
```
sshedi, line 470: Check retval

client/goal.c
```diff
@@ -469,14 +412,157 @@ TDNFGoal(
...
+
+    dwError = TDNFPkgsToExclude(pTdnf, &dwExcludeCount, &ppszExcludes);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    queue_init(&queueJobs);
+
+    for (int i = 0; i < pqInstall->count; i++)
+    {
+        Id id = pqInstall->elements[i];
+        TDNFAddGoal(pTdnf, ALTER_INSTALL, &queueJobs, id,
```
sshedi: check retval

client/history.c
```diff
@@ -0,0 +1,71 @@
...
+    struct history_ctx **ppCtx,
+    int nMustExist
+)
+{
+    uint32_t dwError = 0;
+    char *pszDataDir = NULL;
+    char *pszHistoryDb = NULL;
+    struct history_ctx *ctx = NULL;
+
+    if(!pTdnf || !ppCtx)
```
sshedi, line 23: check for `nMustExist` less than zero

client/history.c
```diff
@@ -0,0 +1,71 @@
...
+
+    dwError = TDNFJoinPath(&pszDataDir,
+                           pTdnf->pArgs->pszInstallRoot,
+                           TDNF_DEFAULT_DATA_LOCATION,
+                           NULL);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFJoinPath(&pszHistoryDb,
+            pszDataDir,
+            "history.db",
```
sshedi: use a macro for this file name

client/utils.c
```diff
@@ -764,3 +764,48 @@ TDNFReadAutoInstalled(
...
+                {
+                    TDNF_SAFE_FREE_MEMORY(pHistoryItems[i].ppszAddedPkgs[j]);
+                }
+                TDNF_SAFE_FREE_MEMORY(pHistoryItems[i].ppszAddedPkgs);
+            }
+            if (pHistoryItems[i].ppszRemovedPkgs != NULL)
+            {
+                for (j = 0; j < pHistoryItems[i].nRemovedCount; j++)
+                {
+                    TDNF_SAFE_FREE_MEMORY(pHistoryItems[i].ppszRemovedPkgs[j]);
```
sshedi, line 792: indent

history/history.h
```diff
@@ -0,0 +1,71 @@
...
+ * you may not use this file except in compliance with the License. The terms
+ * of the License are located in the COPYING file of this distribution.
+ */
+
+#include <sqlite3.h>
+
+#define HISTORY_TRANS_TYPE_BASE 0
+#define HISTORY_TRANS_TYPE_DELTA 1
+
+#define HISTORY_ITEM_TYPE_SET 0
```
sshedi, line 14: use enum here

history/history.h
```diff
@@ -0,0 +1,71 @@
...
+ * Copyright (C) 2022 VMware, Inc. All Rights Reserved.
+ *
+ * Licensed under the GNU Lesser General Public License v2.1 (the "License");
+ * you may not use this file except in compliance with the License. The terms
+ * of the License are located in the COPYING file of this distribution.
+ */
+
+#include <sqlite3.h>
+
+#define HISTORY_TRANS_TYPE_BASE 0
```
sshedi, line 11: use enum

solv/simplequery.c
```diff
@@ -0,0 +1,134 @@
...
+        BAIL_ON_TDNF_LIBSOLV_ERROR(dwError); \
+    } \
+}
+
+/* Split nevra into name, ever, arch. This modifies its input,
+   and returns pointers to the respective components */
+static
+int split_nevra(char *nevra, char **name, char **evr, char **arch)
+{
+    char *p = nevra + strlen(nevra) - 1;
```
sshedi, line 30: validate function args

tools/cli/lib/api.c
```diff
@@ -1043,3 +1043,224 @@ TDNFCliRefresh(
     return TDNFRefresh(pContext->hTdnf);
 }
 
+static
+uint32_t
+TDNFCliHistoryAlter(
+    PTDNF_CLI_CONTEXT pContext,
+    PTDNF_CMD_ARGS pCmdArgs,
+    PTDNF_HISTORY_ARGS pHistoryArgs
+)
```
sshedi, line 1032: validate function args

tools/cli/lib/api.c
```diff
@@ -1043,3 +1043,224 @@ TDNFCliRefresh(
...
+{
+    uint32_t dwError = 0;
+    PTDNF_SOLVED_PKG_INFO pSolvedPkgInfo = NULL;
+
+    dwError = pContext->pFnHistoryResolve(pContext, pHistoryArgs, &pSolvedPkgInfo);
+    BAIL_ON_CLI_ERROR(dwError);
+
+    if (pHistoryArgs->nCommand == HISTORY_CMD_INIT)
+    {
```
sshedi, line 1047: ?

tools/cli/lib/api.c
```diff
@@ -1043,3 +1043,224 @@ TDNFCliRefresh(
...
+error:
+    goto cleanup;
+}
+
+static
+uint32_t
+TDNFCliHistoryList(
+    PTDNF_CLI_CONTEXT pContext,
+    PTDNF_CMD_ARGS pCmdArgs,
+    PTDNF_HISTORY_ARGS pHistoryArgs
```
sshedi, line 1086: validate args

history/history.c
```diff
@@ -0,0 +1,1080 @@
...
+#include <rpm/rpmlib.h>
+#include <rpm/rpmdb.h>
+#include <rpm/rpmlog.h>
+#include <rpm/rpmps.h>
+#include <rpm/rpmts.h>
+
+#include "history.h"
+
+#define check_cond(COND) if(!(COND)) { \
+    fprintf(stderr, "check_cond failed in %s line %d\n", \
```
sshedi, line 23: use `pr_err`

history/history.c
```diff
@@ -0,0 +1,1080 @@
...
+        __FUNCTION__, __LINE__); \
+    rc = -1; \
+    ((void)(rc)); /* suppress "set but not used" warning */ \
+    goto error; \
+}
+
+#define check_ptr(ptr) if(!(ptr)) { \
+    fprintf(stderr, "check_ptr failed in %s line %d\n", \
```
sshedi, line 31: use `pr_err`

history/history.c
```diff
@@ -0,0 +1,1080 @@
...
+static
+void sort_array(int *arr, int count)
+{
+    qsort(arr, count, sizeof(int), _cmp_int);
+}
+
+/*
+ * Diff arr1 and arr2, with sizes count1 and count2. The arrays must be
+ * sorted. The differences
+ * will be stored in ponly1 and ponly2, which are dynamically allocated
```
sshedi: fix comment line formatting

sshedi (changes requested):
history.c is missing function arg checks and return value checks in lot of places.

I will go through this again.

client/api.c
oliverkurth: Right, but it's more readable this way.

client/history.c
oliverkurth: Why?

client/utils.c
oliverkurth: What's wrong with indentation here?

solv/simplequery.c
oliverkurth: This is a static helper function. I don't think we need to check function args for every function.

tools/cli/lib/api.c
oliverkurth: There is nothing to do here for the "init" command. I will add a comment.

client/api.c
```diff
@@ -1904,6 +1904,353 @@ TDNFUpdateInfo(
...
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    rc = history_sync(ctx, ts);
+    if (rc != 0)
+    {
+        dwError = ERROR_TDNF_HISTORY_ERROR;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (pHistoryArgs->nCommand == HISTORY_CMD_INIT)
```
ntsbtz: Can we replace this if else if to switch ?

solv/simplequery.c
```diff
@@ -0,0 +1,134 @@
...
+    int installed
+    )
+{
+    uint32_t dwError = 0;
+    char *n = NULL, *name, *evr, *arch;
+
+    ASSERT_ARG(pool);
+    ASSERT_ARG(qresult);
+
+    n = strdup(nevra);
```
ntsbtz, line 108: TDNFAllocateMemory should be preferred

solv/simplequery.c
```diff
@@ -0,0 +1,134 @@
...
+    Queue* qresult,
+    int installed
+    )
+{
+    uint32_t dwError = 0;
+
+    ASSERT_ARG(pool);
+    ASSERT_ARG(qresult);
+
+    Id p;
```
ntsbtz, line 75: Isn't p should be initialised? Below get used as index

client/api.c
```diff
@@ -1904,6 +1904,353 @@ TDNFUpdateInfo(
...
 }
 
+uint32_t
+TDNFHistoryResolve(
+    PTDNF pTdnf,
+    PTDNF_HISTORY_ARGS pHistoryArgs,
+    PTDNF_SOLVED_PKG_INFO *ppSolvedPkgInfo)
+{
+    uint32_t dwError = 0;
+    int rc;
```
ntsbtz: init rc will be good although below the return value is used

client/api.c
```diff
@@ -1904,6 +1904,353 @@ TDNFUpdateInfo(
...
+    struct history_ctx *ctx = NULL;
+    int rc = 0;
+
+    if(!pTdnf || !pHistoryArgs || !ppHistoryInfo)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (pHistoryArgs->nFrom != 0 && pHistoryArgs->nTo != 0)
```
ntsbtz, line 2162: If i am understanding correctly can we directly check for < 0 in first condition itself and one more separate condition for chekcing nFrom > nTo ?

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+                             int *pcount,
+                             int reverse, int from, int to)
+{
+    sqlite3_stmt *res = NULL;
+    int step, rc = 0;
+    int i, count = 0;
+    struct history_transaction *tas = NULL;
+    char sql[256];
+
+    check_cond(to >= from);
```
ntsbtz, line 837: Redundant check before calling this function we already check From > To

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+        const char *cmdline = (char *)sqlite3_column_text(res, 2);
+        tas[i].cmdline = cmdline ? strdup(cmdline) : NULL;
+
+        tas[i].timestamp = sqlite3_column_int(res, 3);
+        tas[i].type = sqlite3_column_int(res, 4);
+    }
+    count = i;
+
+    for (i = 0; i < count; i++) {
+        history_delta_read(ctx->db, &tas[i].delta, tas[i].id);
```
ntsbtz: Function return status need to check

client/api.c
```diff
@@ -1904,6 +1904,353 @@ TDNFUpdateInfo(
...
+
+    for (int i = 0; i < hd->added_count; i++)
+    {
+        char *pszPkgName = history_get_nevra(hnm, hd->added_ids[i]);
+        if (pszPkgName)
+        {
+            Queue qResult = {0};
+            queue_init(&qResult);
+
+            dwError = SolvFindSolvablesByNevraStr(pTdnf->pSack->pPool,
```
ntsbtz, line 2031: instead of 0 and 1 as installed can we put MACRO

client/goal.c
```diff
@@ -469,14 +412,159 @@ TDNFGoal(
...
+        if (pQueuePkgList->count == 0)
+        {
+            dwError = ERROR_TDNF_ALREADY_INSTALLED;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+
+        for (int i = 0; i < pQueuePkgList->count; i++)
+        {
+            Id dwId = pQueuePkgList->elements[i];
+            TDNFAddGoal(pTdnf, nAlterType, &queueJobs, dwId,
```
ntsbtz, line 470: Return value need to check

client/rpmtrans.c
```diff
@@ -97,14 +99,48 @@ TDNFRpmExecTransaction(
     BAIL_ON_TDNF_ERROR(dwError);
 
     if (!nDownloadOnly) {
+        int rc;
```
ntsbtz, line 102: init rc

client/rpmtrans.c
```diff
@@ -97,14 +99,48 @@ TDNFRpmExecTransaction(
...
+
+        dwError = TDNFGetHistoryCtx(pTdnf, &pHistoryCtx, 0);
```
ntsbtz, line 104: 0 and 1 nMustExist can be replaced by macro

client/utils.c
```diff
@@ -764,3 +764,48 @@ TDNFReadAutoInstalled(
...
     goto cleanup;
 }
+
+void
+TDNFFreeHistoryInfoItems(
+    PTDNF_HISTORY_INFO_ITEM pHistoryItems,
+    int nCount
+)
+{
+    if (pHistoryItems)
```
ntsbtz, line 774: We can put it like args check as in TDNF all places it will maintain the same decorum

common/strings.c
```diff
@@ -175,6 +175,56 @@ TDNFSplitStringToArray(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    nSepSize = strlen(pszSep);
+    nSize = nSepSize * count;
+    for(i = 0; i < count && ppszArray[i]; i++)
+    {
+        nSize += strlen(ppszArray[i]);
+    }
+    nSize++;
```
ntsbtz, line 203: I am assuming this increment for extra null byte. Please correct me if i am wrong

common/strings.c
```diff
@@ -175,6 +175,56 @@ TDNFSplitStringToArray(
...
+
+    dwError = TDNFAllocateMemory(nSize, sizeof(char), (void**)&pszResult);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    for(i = 0; i < count && ppszArray[i]; i++)
+    {
+        strcpy(&pszResult[p], ppszArray[i]);
```
ntsbtz, line 210: if possible can be replaced by alternatives with snprintf or strncpy make sure buffer will well terminated

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+            only1[only1_count++] = arr1[i1++];
+        }
+    }
+    *ponly1 = only1;
+    *ponly1_count = only1_count;
+
+    *ponly2 = only2;
+    *ponly2_count = only2_count;
+
+    return 0;
```
ntsbtz: return rc

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    check_db_rc(db, rc);
+
+    step = sqlite3_step(res);
+    if (step == SQLITE_ROW) {
+        *pcount = sqlite3_column_int(res, 0);
+    }
+
+error:
+    if (res)
+        sqlite3_finalize(res);
```
ntsbtz, line 163: Can we put this in cleanup as we are releasing resource after finalizing statement

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    check_db_rc(db, rc);
+
+    step = sqlite3_step(res);
+    check_cond(step == SQLITE_ROW);
+
+    *pmaxid = sqlite3_column_int(res, 0);
+    sqlite3_finalize(res); res = NULL;
+error:
+    if (res)
+        sqlite3_finalize(res);
```
ntsbtz, line 187: Same as above

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    }
+}
+
+/*
+ * Create a map with all known NEVRAs.
+ */
+static
+struct history_nevra_map *db_nevra_map(sqlite3 *db)
+{
+    int rc;
```
ntsbtz: init rc

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+        check_ptr(nevra);
+        check_cond((id > 0 && id <= count));
+        /* map is zero based, rpm ids start with 1 */
+        hnm->idmap[id - 1] = nevra;
+    }
+error:
+    if (res)
+        sqlite3_finalize(res);
+    if (rc) {
+        db_free_nevra_map(hnm); hnm = NULL;
```
ntsbtz, line 242: goto cleanup

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    int rc = 0;
+    sqlite3_stmt *res = NULL;
+    const char *sql = "SELECT * FROM trans_items WHERE trans_id = ? ORDER BY rpm_id;";
+
+    rc = sqlite3_prepare_v2(db, sql, -1, &res, 0);
+    check_db_rc(db, rc);
+
+    sqlite3_bind_int(res, 1, trans_id);
+    for(int step = sqlite3_step(res); step == SQLITE_ROW; step = sqlite3_step(res)) {
+        int type = sqlite3_column_int(res, 2);
```
ntsbtz, line 349: can we replace 1,2 and 3 with self-explanatory MACRO name

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+        check_cond(type == HISTORY_ITEM_TYPE_ADD || type == HISTORY_ITEM_TYPE_REMOVE);
+        if (type == HISTORY_ITEM_TYPE_ADD) {
+            map_set(installed_map, rpm_id);
+        }
+        else if (type == HISTORY_ITEM_TYPE_REMOVE) {
+            map_unset(installed_map, rpm_id);
+        }
+    }
+
+error:
```
ntsbtz, line 361: missing goto cleanup in file we can maintain that

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    int step, rc = 0;
+
+    /* find most recent HISTORY_TRANS_TYPE_BASE */
+    rc = sqlite3_prepare_v2(db,
+                            "SELECT * FROM transactions "
+                                "WHERE type = ? ORDER BY id DESC;",
+                            -1, &res, 0);
+    check_db_rc(db, rc);
+
+    sqlite3_bind_int(res, 1, HISTORY_TRANS_TYPE_BASE);
```
ntsbtz, line 415: what if init fail

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+int db_update_rpms(rpmts ts, sqlite3 *db, int **pids, int *pcount)
+{
+    int rc = 0;
+    int count = 0, i = 0;
+    int *ids = NULL;
+    Header h;
+    rpmdbMatchIterator mi = NULL;
+    char *nevra = NULL;
+
+    if (pids) {
```
ntsbtz, line 454: can we check for all function parameter

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    return rc;
+}
+
+
+/* Helper to set the ctx cookie, free'ing the old one if needed */
+static
+void history_set_cookie(struct history_ctx *ctx, const char *cookie)
+{
+    safe_free(ctx->cookie);
+    ctx->cookie = strdup(cookie);
```
ntsbtz, line 561: TdnfAllocateMemory instead of strdup if it does the needful

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    check_ptr(installed_ids1);
+
+    rc = diff_arrays(installed_ids1, installed_count1,
+                     installed_ids0, installed_count0,
+                     &hd->removed_ids, &hd->removed_count,
+                     &hd->added_ids, &hd->added_count);
+    check_rc(rc);
+
+error:
+    safe_free(installed_ids0);
```
ntsbtz, line 706: Freeing resources should be in cleanup in all places wherever we release the resources

history/history.c
```diff
@@ -0,0 +1,1079 @@
...
+    check_ptr(installed_map);
+
+    rc = db_play_transaction(ctx->db, trans_id, installed_map, map_size);
+    check_rc(rc);
+
+    history_set_ids_from_map(ctx, installed_map, map_size);
+
+error:
+    safe_free(installed_map);
+
```
ntsbtz, line 755: extra line

history/history.c
ntsbtz, line 837: redundant check  before calling already checked

include/tdnfclitypes.h
```diff
@@ -169,6 +181,8 @@ typedef struct _TDNF_CLI_CONTEXT_
     PFN_TDNF_SEARCH             pFnSearch;
     PFN_TDNF_UPDATEINFO         pFnUpdateInfo;
     PFN_TDNF_UPDATEINFO_SUMMARY pFnUpdateInfoSummary;
+    PFN_TDNF_HISTORY_CMD        pFnHistoryList;
+    PFN_TDNF_HISTORY_RESOLVE_CMD  pFnHistoryResolve;
```
ntsbtz, line 184: Alignment of members

tools/cli/lib/api.c
```diff
@@ -1043,3 +1043,237 @@ TDNFCliRefresh(
...
+        dwError = ERROR_TDNF_CLI_INVALID_ARGUMENT;
+        BAIL_ON_CLI_ERROR(dwError);
+    }
+
+    dwError = pContext->pFnHistoryList(pContext, pHistoryArgs, &pHistoryInfo);
+    BAIL_ON_CLI_ERROR(dwError);
+
+    pItems = pHistoryInfo->pItems;
+
+    if (!pCmdArgs->nJsonOutput)
```
ntsbtz, line 1108: can we put both if and else separately with thier respective function . Although it requires  two calling but looks more readable

tools/cli/lib/api.c
```diff
@@ -1043,3 +1043,237 @@ TDNFCliRefresh(
...
+    {
+        int nConsoleWidth, nCmdWidth;
+
+        dwError = GetConsoleWidth(&nConsoleWidth);
+        BAIL_ON_CLI_ERROR(dwError);
+
+        /* We want to give as much space as possible to the command line.
+           All other fields have fixed lengths. */
+        /* 4 = ID, 21 = date/time, 9 = '+add/-rem', 7 = spaces */
+        nCmdWidth = nConsoleWidth - 4 - 21 - 9 - 7;
```
ntsbtz, line 1118: Single MACRO name something like CMD_ID_DATE_ACTION_FIXED_LEN 41

solv/simplequery.c
oliverkurth: No need, see https://github.com/openSUSE/libsolv/blob/6ffe49cdd5c4f53b9135a51179e3b7c160a472aa/src/pool.h#L409

client/api.c
oliverkurth: Doesn't hurt to initialize, but it's never used uninitialized.

client/api.c
oliverkurth: Good point, done.

history/history.c
oliverkurth: Yes, it's redundant in this case, but maybe future changes add callers that omit the check. This is an API function where we should do some checking, but if this was just a helper function we could skip it.

history/history.c
oliverkurth: Done

client/rpmtrans.c
oliverkurth: Not needed.

client/utils.c
oliverkurth: We don't do that for the `TDNFree*` functions because they return `void`.

common/strings.c
oliverkurth: Yes.

common/strings.c
oliverkurth: `strcpy()` copies the `NUL` terminator. And we have checked the size before. I don't see how `snprintf()` or `strncpy()` would improve this.

tools/cli/lib/api.c
oliverkurth: I think having this here with explanation makes it much clearer.

tools/cli/lib/api.c
oliverkurth: Hm, I don't agree that it would make it more readable. It's one more step if you want to follow the code.

history/history.c
oliverkurth: Is this a duplicate comment? Already answered above.

history/history.c
oliverkurth: I am trying another coding style in this API. Here we clean up everything below `error`.

history/history.c
oliverkurth: This API uses another coding style than the main TDNF code.

history/history.c
oliverkurth: It's a static helper function. The caller is responsible for giving valid args.

Note that we check `pids` only because `NULL` is valid (it means we are not interested in them on return).

history/history.c
oliverkurth: Different coding style in this API.

solv/simplequery.c
oliverkurth: Different coding style in this API.

client/api.c
oliverkurth: Agreed, looks better.

history/history.c
oliverkurth: Again, different style. There is no `cleanup` label.

history/history.c
oliverkurth: I want to keep the option open to use this code in a patched `rpm`, so I want to avoid dependencies on `tdnf`.

history/history.c
oliverkurth: Thanks - I will do that in an upcoming PR (I made the change, but it's too much effort to backport the commit)

history/history.c
oliverkurth: I can do that in a later PR. Changing it in this PR will just create too many conflicts for my next change.

client/api.c
oliverkurth: Done.

client/rpmtrans.c
oliverkurth: I thought about it, but do you have any suggestion for a name for the "does not have to exist" case?

common/strings.c
```diff
@@ -175,6 +175,56 @@ TDNFSplitStringToArray(
...
+    char *pszResult = NULL;
+
+    if(!ppszArray || !pszSep || !ppszResult)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    nSepSize = strlen(pszSep);
+    nSize = nSepSize * count;
```
sshedi, line 198: Let's do it this way:
`nSize = (nSepSize * count) + 1024`

get rid of first loop of getting total size and start copying string to buffer, keep count of total copied bytes, by if we exceed nSize, then realloc. We may never need realloc or rarely realloc but  this will optimize a lot of function calls

solv/simplequery.c
```diff
@@ -0,0 +1,134 @@
...
+    if (!(x)) { \
+        dwError = ERROR_TDNF_OUT_OF_MEMORY; \
+        BAIL_ON_TDNF_LIBSOLV_ERROR(dwError); \
+    } \
+}
+
+/* Split nevra into name, ever, arch. This modifies its input,
+   and returns pointers to the respective components */
+static
+int split_nevra(char *nevra, char **name, char **evr, char **arch)
```
sshedi, line 28: use this instead:
```
#include <stdio.h>
#include <string.h>

char x[] = "tar-xyz-1.34-1.ph4";

struct nevra {
   char *name;
   char *evr;
   char *arch;
};

static int split_nevra(char *nevra, struct nevra *n)
{
    char *p = strrchr(nevra, '.');
    if (p == NULL) {
        return -1;
    }

    *p = 0;
    n->arch = p + 1;

    p = strrchr(nevra, '-');
    if (p == NULL){
        return -2;
    }

    while (*(--p) != '-' && p > nevra);
     if (p > nevra) {
         return -3;
     }

    *p = 0;
    n->evr = p + 1;

    n->name = nevra;

    return 0;
}

int main(void)
{
  struct nevra n = {};

  split_nevra(x, &n);
  printf("%s %s %s\n", n.name, n.evr, n.arch);

  return 0;
}
```

solv/simplequery.c
```diff
@@ -0,0 +1,134 @@
...
+    int installed
+    )
+{
+    uint32_t dwError = 0;
+    char *n = NULL, *name, *evr, *arch;
+
+    ASSERT_ARG(pool);
+    ASSERT_ARG(qresult);
+
+    n = strdup(nevra);
```
sshedi, line 108: I would suggest doing this in split_nevra function and freeing the return pointer in caller. That way split_nevra would work for any string.

Or free the name pointer, `name = nevra` anyways.

solv/simplequery.c
```diff
@@ -0,0 +1,134 @@
...
+    Id id_arch = pool_str2id(pool, arch, 0);
+
+    if (id_name && id_evr && id_arch)
+    {
+        dwError = SolvFindSolvablesByNevraId(pool, id_name, id_evr, id_arch, qresult, installed);
+        BAIL_ON_TDNF_LIBSOLV_ERROR(dwError);
+    }
+
+cleanup:
+    if (n)
```
sshedi, line 128: TDNF_SAFE_FREE()

solv/simplequery.c
```diff
@@ -0,0 +1,134 @@
...
+    n = strdup(nevra);
+    ASSERT_MEM(n);
+
+    if (split_nevra(n, &name, &evr, &arch) != 0)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_LIBSOLV_ERROR(dwError);
+    }
+
+    Id id_name = pool_str2id(pool, name, 0);
```
sshedi, line 117: declare at top

history/history.c
sshedi: In that case use appropriate header guards and use it.

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    goto error; \
+}
+
+#define check_rc(rc) if((rc) != 0) { \
+    fprintf(stderr, "check_rc failed in %s line %d\n", \
+        __FUNCTION__, __LINE__); \
+    goto error; \
+}
+
+#define check_db_rc(db, rc) if((rc) != SQLITE_OK) { \
```
sshedi, line 44: nit: `if<space>((rc) !`

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    fprintf(stderr, \
+        "check_db_rc failed with %s in %s line %d\n", \
+        sqlite3_errmsg(db), __FUNCTION__, __LINE__); \
+    ((void)(rc)); /* suppress "set but not used" warning */ \
+    goto error; \
+}
+
+#define check_db_cmd(db, cmd) { \
+    rc = (cmd); \
```
sshedi, line 53: this is a bad macro. Use some thing like:
```
{ \
  int __rc = (cmd); \
   check_db_rc(db, __rc); \
}
```

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    check_db_rc(db, rc); \
+}
+
+#define safe_free(ptr) { if ((ptr) != NULL) { free(ptr); ptr = NULL; }}
```
sshedi, line 57: Define a generic tdnf macro for this, we might need it in some other module as well.

I think lot of other macros as well, move them into a common place.
They will be useful in future.

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+
+/*
+ * Diff arr1 and arr2, with sizes count1 and count2. The arrays must be
+ * sorted. The differences will be stored in ponly1 and ponly2, which are
+ * dynamically allocated and should be free'd by the caller. Counts of the
+ * created arrays are stored in ponly1_count and ponly2_count.
+ */
+static
+int diff_arrays(int *arr1, int count1,
+                 int *arr2, int count2,
```
sshedi, line 83: use a structure to keep this. That way it looks neat.
```
struct A {
   int *arr;
   int count;
   int **ponly;
   int *ponly_count;
}
```

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    int only1_count = 0, only2_count = 0;
+
+    check_ptr(only1 = (int *)calloc(count1, sizeof(int)));
+    check_ptr(only2 = (int *)calloc(count2, sizeof(int)));
+
+    while(i1 < count1 && i2 < count2) {
+        if (arr1[i1] != arr2[i2]) {
+            if(arr1[i1] > arr2[i2]) {
+                only2[only2_count++] = arr2[i2++];
+            } else {
```
sshedi, line 99: `else if arr1[i1] < arr2[i2])` read is cheaper than write

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+                only1[only1_count++] = arr1[i1++];
+            }
+        } else {
+            i1++;
+            i2++;
+        }
+    }
+
+    if(i1 == count1) {
```
sshedi, line 108: no need of this if.
```
while (i2 < count2) {
            only2[only2_count++] = arr2[i2++];
}
while (i1 < count1) {
            only1[only1_count++] = arr1[i1++];
}
```

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    }
+    return rc;
+}
+
+static
+sqlite3 *init_db(const char *filename)
+{
+    sqlite3 *db = NULL;
+
+    int rc = sqlite3_open(filename, &db);
```
sshedi, line 136: declare and use separately, validate filename

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+
+error:
+    return db;
+}
+
+/* count known rpms in db */
+static
+int db_rpms_count(sqlite3 *db, int *pcount)
+{
+    int rc = 0, step;
```
sshedi, line 147: I really prefer having variables declared in independent lines

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    sqlite3_stmt *res = NULL;
+    const char *sql = "SELECT count(*) FROM rpms;";
```
sshedi, line 149: Use a nice macro for these sql queries

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    return rc;
+}
+
+/* max rpm id db (should be same as count unless there are gaps) */
+static
+int db_rpms_maxid(sqlite3 *db, int *pmaxid)
+{
+    int rc = 0, step;
+    sqlite3_stmt *res = NULL;
+    const char *sql = "SELECT * FROM rpms ORDER BY id DESC;";
```
sshedi, line 173: use macro for this query

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    rc = sqlite3_prepare_v2(db,
+                            sql,
+                            -1, &res, 0);
+    check_db_rc(db, rc);
+
+    step = sqlite3_step(res);
+    check_cond(step == SQLITE_ROW);
+
+    *pmaxid = sqlite3_column_int(res, 0);
+    sqlite3_finalize(res); res = NULL;
```
sshedi, line 184: do `res = NULL` in new line or use a macro for this

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+error:
+    if (res)
+        sqlite3_finalize(res);
+    return rc;
+}
+
+static
+void db_free_nevra_map(struct history_nevra_map *hnm)
+{
+    if (hnm){
```
sshedi, line 194:
```
if (hnm == NULL || hnm->idmap == NULL) {
  return;
}
...
rest of the logic
```
Saves you indentation level

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+        if (hnm->idmap) {
+            for (int i = 0; i < hnm->count; i++) {
```
sshedi, line 196: declare `i` at the function beginning

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+/*
+ * Create a map with all known NEVRAs.
+ */
+static
+struct history_nevra_map *db_nevra_map(sqlite3 *db)
+{
+    int rc = 0;
+    struct history_nevra_map *hnm = NULL;
+    sqlite3_stmt *res = NULL;
+    const char *sql_find = "SELECT * FROM rpms;";
```
sshedi, line 214: use macro for query

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+/*
+ * Add nevra to db if it's not there already. The id will be returned
+ * in *pid if that's not NULL.
+*/
+static
+int db_add_nevra(sqlite3 *db, const char *nevra, int *pid)
+{
+    int rc = 0, step;
+    sqlite3_stmt *res = NULL;
+    const char *sql_find = "SELECT * FROM rpms WHERE nevra = ?;";
```
sshedi, line 256: macro please

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+        sqlite3_finalize(res);
+    if (rc) {
+        db_free_nevra_map(hnm); hnm = NULL;
+    }
+    return hnm;
+}
+
+/*
+ * Add nevra to db if it's not there already. The id will be returned
+ * in *pid if that's not NULL.
```
sshedi, line 249: I didn't get this comment

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    return rc;
+}
+
+/* read transaction items into ht */
+static
+int history_delta_read(sqlite3 *db, struct history_delta *hd, int id)
+{
+    int rc = 0;
+    sqlite3_stmt *res = NULL;
+    const char *sql = "SELECT * FROM trans_items WHERE trans_id = ? ORDER BY rpm_id;";
```
sshedi, line 297: use macro

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    char *err_msg = NULL;
+    int trans_id = 0;
+
+    /* avoid unfinished transaction record on failure or crash */
+    rc = sqlite3_exec(ctx->db, "BEGIN TRANSACTION;", 0, 0, NULL);
+    check_db_rc(ctx->db, rc);
+
+    rc = sqlite3_exec(ctx->db,
+        "CREATE TABLE IF NOT EXISTS "
+            "transactions(Id INTEGER PRIMARY KEY AUTOINCREMENT,"
```
sshedi, line 774: all these queries should be kept in a macro

history/history.c
```diff
@@ -0,0 +1,1081 @@
+/*
```
sshedi, line 1: Few generic comments for this file:
- Validate function args
- Use macros for sql queries
- Move useful macros to common place

It would be good to have a simple wrapper functions for `sqlite3_*`
function calls. If we want to migrate to a different db someday, we will have to change the code in one place which will make maintenance easy.

For example:
```
typedef sqlite3 db;

int db_open(const char *fn, db *_db)
{
    int rc;

    rc = sqlite3_open(fn, _db);
    return rc;
}

static db_init_db(const char *fn)
{
    int rc;
    db _db;

    rc = db_open(fn, &_db);
    check_db_rc(db, rc);

    return _db;
}
```
Like this, we need to have simple wrappers for each sqlite3 api.

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+    int rc = 0;
+    int i1 = 0, i2 = 0;
+    int *only1 = NULL, *only2 = NULL;
+    int only1_count = 0, only2_count = 0;
+
+    check_ptr(only1 = (int *)calloc(count1, sizeof(int)));
+    check_ptr(only2 = (int *)calloc(count2, sizeof(int)));
+
+    while(i1 < count1 && i2 < count2) {
+        if (arr1[i1] != arr2[i2]) {
```
sshedi, line 96: this condition is not needed.

history/history.c
```diff
@@ -0,0 +1,1081 @@
...
+}
+
+static
+void sort_array(int *arr, int count)
+{
+    qsort(arr, count, sizeof(int), _cmp_int);
+}
+
+/*
+ * Diff arr1 and arr2, with sizes count1 and count2. The arrays must be
```
sshedi, line 76: We can simplify this further:
```
static
int diff_arrays(int *arr1, int count1,
                 int *arr2, int count2,
                 int **ponly1, int *ponly1_count,
                 int **ponly2, int *ponly2_count)
{
    int rc = 0;
    int i1 = 0, i2 = 0;
    int *only1 = NULL, *only2 = NULL;
    int only1_count = 0, only2_count = 0;

    check_ptr(only1 = (int *)calloc(count1, sizeof(int)));
    check_ptr(only2 = (int *)calloc(count2, sizeof(int)));

    for (; i1 < count1 && i2 < count2; i1++, i2++) {
        if (arr1[i1] != arr2[i2]) {
            only1[only1_count++] = arr1[i1];
            only2[only2_count++] = arr2[i2];
        }
    }

    while (i2 < count2) {
        only2[only2_count++] = arr2[i2++];
    }

    while (i1 < count1) {
        only1[only1_count++] = arr1[i1++];
    }

    *ponly1 = only1;
    *ponly1_count = only1_count;

    *ponly2 = only2;
    *ponly2_count = only2_count;

error:
    if(rc) {
        safe_free(only1);
        safe_free(only2);
    }
    return rc;
}
```

history/history.c
oliverkurth: For the `SELECT` queries I disagree. It just creates more work trying to figure out what we are doing.

history/history.c
oliverkurth: I agree for the `CREATE` queries, and I did that in PR #327 already.

history/history.c
oliverkurth: I don't think we should check args for every function call, that's just overkill. We can do that for the API calls, and I think I have done that, but not for every helper function.

As for alternative dbs, let's address that when needed, and I don't believe it will be soon. It will always be a local db, no point in using for example postgres. Besides, we have control over it. Even if RPM decides to use something else we do not need to switch.

common/strings.c
oliverkurth: I don't see how this improves anything, it just makes it more complex. Currently we calculate the exact size. The performance impact by looping is negligible, and I wouldn't be surprised if the `realloc()` would be even more costly.

solv/simplequery.c
oliverkurth: Why?

history/history.c
oliverkurth: I don't think that works. If `arr1` is` [1, 2]` and `arr2` is `[2, 3]`, we'd add `2` to `only2` on the first iteration of the loop, and that's wrong.

PR #317, opened 2022-05-18
add --testonly option

oliverkurth:
Add the option `--testonly` to only test a transaction but not actually running it. This is similar to the `--test` for `rpm`.


sshedi (approved):

oliverkurth:
This actually breaks the `autoremove` feature - dependencies would be marked as automatically installed although they are not actually installed.

Since I am going to shuffle code around in other PRs I am putting this on hold for now.

PR #318, opened 2022-05-26
common/strings.c: simplify TDNFTrimSuffix function

sshedi:

### DIFF
--- a/common/strings.c
+++ b/common/strings.c
@@ -422,42 +422,21 @@ error:
     goto cleanup;
 }
 
-uint32_t
-TDNFTrimSuffix(
-    char* pszSource,
-    const char* pszSuffix
-    )
+uint32_t TDNFTrimSuffix(char *pszSource, const char *pszSuffix)
 {
-    uint32_t dwError = 0;
-    int nSourceStrLen = 0, nSuffixStrLen = 0;
+    char *ptr;
 
-    if (IsNullOrEmptyString(pszSource) || IsNullOrEmptyString(pszSuffix))
-    {
-        dwError = ERROR_TDNF_INVALID_PARAMETER;
-        BAIL_ON_TDNF_ERROR(dwError);
+    if (IsNullOrEmptyString(pszSource) || IsNullOrEmptyString(pszSuffix)) {
+        return ERROR_TDNF_INVALID_PARAMETER;
     }
 
-    nSourceStrLen = strlen(pszSource);
-    nSuffixStrLen = strlen(pszSuffix);
-
-    if (nSuffixStrLen > nSourceStrLen)
-    {
-        dwError = ERROR_TDNF_INVALID_PARAMETER;
-        BAIL_ON_TDNF_ERROR(dwError);
+    ptr = strstr(pszSource, pszSuffix);
+    if (!ptr || strcmp(ptr, pszSuffix)) {
+        return ERROR_TDNF_INVALID_PARAMETER;
     }
 
-    while (nSuffixStrLen > 0 &&
-          (pszSource[nSourceStrLen - 1] == pszSuffix[nSuffixStrLen - 1]))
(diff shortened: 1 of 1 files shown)


common/strings.c
```diff
@@ -422,42 +422,24 @@ TDNFReplaceString(
...
-        dwError = ERROR_TDNF_INVALID_PARAMETER;
-        BAIL_ON_TDNF_ERROR(dwError);
+        return ERROR_TDNF_INVALID_PARAMETER;
     }
 
-    nSourceStrLen = strlen(pszSource);
-    nSuffixStrLen = strlen(pszSuffix);
-
-    if (nSuffixStrLen > nSourceStrLen)
+    ptr = strcasestr(pszSource, pszSuffix);
```
oliverkurth: This ignores case. Is that intended? The original function is case sensitive.

common/strings.c
```diff
@@ -422,42 +422,24 @@ TDNFReplaceString(
...
+    if (!ptr)
     {
-        dwError = ERROR_TDNF_INVALID_PARAMETER;
-        BAIL_ON_TDNF_ERROR(dwError);
+        return ERROR_TDNF_INVALID_PARAMETER;
     }
 
-    while (nSuffixStrLen > 0 &&
```
oliverkurth, line 449: It looks to me that the original function would strip anything from the end that is matching, not only if the whole suffix matches. For example `TDNFTrimSuffix("foobar", "feebar")` would result in `"foo"`, but the new one would return an error.

There is only one caller (in `client/repo.c`), and looks like this is being checked before, so that's probably okay (maybe a bug in the old implementation even).

sshedi: No it doesn't. Newer version of code also return `foo`
Try it here: https://onlinegdb.com/GKQq_joZD

common/strings.c
sshedi: Will fix it. Thanks.

common/strings.c
oliverkurth: For `TDNFTrimSuffix("foobar", "feebar")` it returns an error, I just tried. I modified the code also to show the error, see https://onlinegdb.com/F1_2QHAmK. But your version of the code there is different from this PR.

This, from this PR:
```
    ptr = strstr(pszSource, pszSuffix);
    if (!ptr) {
        return ERROR_TDNF_INVALID_PARAMETER;
```
will also return an error if the suffix is not in the string (`ptr` will be `NULL`).

But that's good.  It's the old version that misbehaves. We don't want to change the string if the suffix doesn't match. And the caller in `client/repo.c` already checks if the suffix matches.

oliverkurth: I actually found a bug - this does not return an error although the suffix doesn't match:
```
    char s1[] = "foobar1";
    char s2[] = "bar";

    ret = TDNFTrimSuffix(s1, s2);
    printf("ret = %d, %s:%lu\n", ret, s1, strlen(s1));
```
Here is a version that works - but please verify:
```
uint32_t TDNFTrimSuffix(char *pszSource, const char *pszSuffix)
{
    int l1, l2;
    uint32_t dwError = 0;

    if (!pszSource || !pszSuffix)
    {
        BAIL_ON_TDNF_ERROR((dwError = ERROR_TDNF_INVALID_PARAMETER));
    }

    l1 = strlen(pszSource);
    l2 = strlen(pszSuffix);
    if (l1 >= l2 && strcmp(&pszSource[l1-l2], pszSuffix) == 0) {
        pszSource[l1-l2] = 0;
    } else {
        printf("error\n");
        BAIL_ON_TDNF_ERROR((dwError = ERROR_TDNF_INVALID_PARAMETER));
    }

cleanup:
    return dwError;

error:
    goto cleanup;
}
```

oliverkurth:
LGTM.

PR #320, opened 2022-05-31
move autoremove flags to history db transactions

oliverkurth:
PR #314 will break the status of the `autoremove` flags for package. For example, a rollback will not restore the flags. This change move tracking of the flag from the flat file `/var/lib/tdnf/autoinstalled` to the db, and tracks it for transactions.

Currently this is a change against the `dev` branch, so includes PR #314 . I will rebase when that is merged.


oliverkurth:
Closing, let's go right to PR #327 .

Issue #34, opened 2015-10-21
Check for available size before download/install


ppadmavilasom:
@YustasSwamp reports this. Check for available size before download/install and throw appropriate errors if not enough space to perform action.

oliverkurth:
There are two issues to address:
* space for downloading the package before doing the download. We can just check the space for the partition where we store the downloads (`/var/tdnf/cache` unless configured differently).
* space for the actual install. This can be tricky if there are multiple partitions involved.

prestonsn:
PR https://github.com/vmware/tdnf/pull/321 provides a cache directory space check. However, it does not address a potential out of disk space error during RPM installs.

oliverkurth:
I think we can close this. The first item is addressed. The second is actually also addressed - it will be detected by the rpm backend, when the transaction is attempted.

PR #321, opened 2022-06-10
Iss34 check avail cache size for download

prestonsn:
Pull-request for issue https://github.com/vmware/tdnf/issues/34. Addresses missing check for available disk space prior to installing package(s). This PR **only** provides a cache directory space check, and does not check for available disk space at package install paths. 

- [x] Adds tracking of package download sizes. 
- [x] Reports download size of package(s) to the user prior to accepting an install.
- [x] Reports download size of package(s) through json formatting when enabled.
- [x] Adds sufficient error handling for `statfs` syscall.  
- [x] Adds tests to ensure accuracy of reported download package size(s).
- [x] Adds test to intentionaly trigger an out of disk space error.
- [x] Fix Copyright year in file headers. 

An install will now appear as follows: 
```
root [ /build ]# ./bin/tdnf install htop gdb vim

Installing:
vim                    x86_64           8.2.4925-1.ph4         photon-updates     3.44M           1.53M
gdb                    x86_64           10.1-2.ph4             photon-updates    10.54M           4.21M
htop                   x86_64           3.1.2-1.ph4            photon-updates   383.54k         163.31k

Total installed size:  14.36M
Total download size:   5.89M
Is this ok [y/N]: y

Downloading:
vim                                    1600344 100%
gdb                                    4410176 100%
htop                                    167227 100%
Testing transaction
Running transaction
Installing/Updating: htop-3.1.2-1.ph4.x86_64
Installing/Updating: gdb-10.1-2.ph4.x86_64
Installing/Updating: vim-8.2.4925-1.ph4.x86_64

Complete!
```


client/packageutils.c
```diff
@@ -933,6 +944,89 @@ TDNFCheckProtectedPkgs(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if(!pConf || !pConf->pszCacheDir)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    struct statfs tmpStatfsBuffer;
```
sshedi: declare at top

common/utils.c
```diff
@@ -156,8 +156,7 @@ TDNFUtilsFormatSize(
     dwError = TDNFAllocateMemory(1, nMaxSize, (void**)&pszFormattedSize);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    if(sprintf(pszFormattedSize, "%6.2f%c %lu", dSize, pszSizes[nIndex],
-        (unsigned long)unSize) < 0)
+    if(sprintf(pszFormattedSize, "%6.2f%c", dSize, pszSizes[nIndex]) < 0)
```
sshedi: use snprintf

tools/cli/lib/installcmd.c
```diff
@@ -675,18 +680,27 @@ PrintAction(
...
+            ppszInfoToPrint[4],
+            nColWidths[5],
+            ppszInfoToPrint[5]);
         pPkgInfo = pPkgInfo->pNext;
     }
 
     TDNFUtilsFormatSize(dwTotalInstallSize, &pszTotalInstallSize);
     pr_info("\nTotal installed size: %s\n", pszTotalInstallSize);
 
+    TDNFUtilsFormatSize(dwTotalDownloadSize, &pszTotalDownloadSize);
```
sshedi: validate retval? here and at line 689 as well

tools/cli/lib/installcmd.c
```diff
@@ -675,18 +680,27 @@ PrintAction(
...
+    pr_info("Total download size: %s\n", pszTotalDownloadSize);
+
 cleanup:
     if (pszTotalInstallSize)
     {
         free(pszTotalInstallSize);
     }
+    if (pszTotalDownloadSize)
```
sshedi: Use `TDNFFreeMemory` here and in previous if

client/api.c
```diff
@@ -1624,6 +1624,12 @@ TDNFResolve(
         pSolvedPkgInfo->pPkgsToDowngrade ||
         pSolvedPkgInfo->pPkgsToReinstall;
 
+    dwError = TDNFCalculateTotalDownloadSize(pSolvedPkgInfo);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFCheckAvailableCache(pTdnf->pConf, pSolvedPkgInfo);
```
sshedi: A small design change here.

```
1. First calculate available cache
2. Then while looping and getting individual rpm size, keep checking if accumulated size exceeded total cache size.
````

File: client/packageutils.c [hunk 6]
```diff
@@ -933,6 +944,89 @@ TDNFCheckProtectedPkgs(
...
+    for (byPkgIndex = 0; byPkgIndex < ARRAY_SIZE(ppPkgsNeedDownload); byPkgIndex++)
+    {
+        pPkgInfo = ppPkgsNeedDownload[byPkgIndex];
+        while(pPkgInfo) {
+            dwTotalDownloadSizeBytes += pPkgInfo->dwDownloadSizeBytes;
+            pPkgInfo = pPkgInfo->pNext;
+        }
+    }
+
+    pSolvedPkgInfo->dwTotalDownloadSizeBytes = dwTotalDownloadSizeBytes;
```
sshedi: may be we don't need this member `dwTotalDownloadSizeBytes` if we calculate total available size first and then if check if we crossed the size upon each iteration.

sshedi (changes requested):

client/packageutils.c
prestonsn: Moved!

common/utils.c
prestonsn: switched to `snprintf()`. Made use of the `nMaxSize` integer as well.

client/api.c
prestonsn: I like this better! I thought maybe having the accumulated size stored somewhere would be useful for something later on. But for now, this is cleaner.

sshedi: For what future requirements we can use this stored total size?

I think my idea is simple and realistic here.

When shopping for something we first think how much amount we have in hand and then later we shop accordingly.
Likewise, first calculate total size available and then check if we crossed it. It might save some iterations when available size is actually lesser than total size of rpms to be downloaded.

It's okay if you want to keep the total download size in a structure member for future needs but design-wise, I'm not really convinced with the existing approach.

client/api.c
```diff
@@ -1557,6 +1557,7 @@ TDNFResolve(
     Queue queueGoal = {0};
     char** ppszPkgsNotResolved = NULL;
     PTDNF_SOLVED_PKG_INFO pSolvedPkgInfo = NULL;
+    uint64_t qwAvailCacheBytes;
```
tapakund: pls initialise to 0.

client/packageutils.c
```diff
@@ -933,6 +944,87 @@ TDNFCheckProtectedPkgs(
...
+    PTDNF_SOLVED_PKG_INFO pSolvedPkgInfo,
+    uint64_t qwAvailCacheBytes
+    )
+{
+    uint32_t dwError = 0;
+    uint64_t qwTotalDownloadSizeBytes = 0;
+    uint8_t byPkgIndex = 0;
+    PTDNF_PKG_INFO pPkgInfo = NULL;
+
+    if(!pSolvedPkgInfo)
```
tapakund, line 992: can we club?
if(!pSolvedPkgInfo || !pSolvedPkgInfo->nNeedDownload)

tapakund (commented):
few comments.

client/api.c
prestonsn: Oh no I meant what you suggested was much cleaner. I agree with your suggestions and in fact added it to the latest commit earlier today: https://github.com/vmware/tdnf/pull/321/commits/4a7d1aff2cbe1b83473bd13f6867e345cd106d2b. That's why I said I liked your approach better! 

I should have been more clear, sorry about that.

As for a future use case for storing download sizes, I'm not exactly sure. So I removed it as suggested in the same commit mentioned previously: https://github.com/vmware/tdnf/pull/321/commits/4a7d1aff2cbe1b83473bd13f6867e345cd106d2b

sshedi: Understood, thanks for the clarification.

sshedi:
I have two more things.

1. Can we somehow create a test for negative scenario? Where available size is less than download size? It might be difficult but it's good to have it.
2. Do we need to add + 5 or 10% to total download size ? rpms are zstd compressed, after extraction size will be more. So we might be able to download but not extract and install the rpm because space won't be available. Consider linux rpm for example,
```
➜ $ ls
boot  etc  lib  linux-5.10.78-6.ph4.x86_64.rpm  var
[Tue 14.Jun.2022 10:51:39] sshedi@ ~/work/photon-dev/t (dev)➤
➜ $ du *
14M	boot
4.0K	etc
23M	lib
29M	linux-5.10.78-6.ph4.x86_64.rpm
4.0K	var
```
rpm is 29M, after extraction it's ~37M.

Both these are my queries, I'm okay with not having these things.

Changes LGTM.

oliverkurth:
> 2. Do we need to add + 5 or 10% to total download size ? rpms are zstd compressed, after extraction size will be more. So we might be able to download but not extract and install the rpm because space won't be available. Consider linux rpm for example,

> rpm is 29M, after extraction it's ~37M.

Shreenidhi, we don't have to guess, we know the exact install size from the metadata. The problem is that the package may be installed across multiple filesystems, and we don't know how the sizes of installed files (AFAIK, maybe we can finds out).

However, this is not a common case, so maybe we can make the check optional, or even detect it automatically.

client/packageutils.c
prestonsn: I agree and would like to put both condition in one `if` statement, but one of those cases, namely `if(!pSolvedPkgInfo)`, is an error and should return a non-zero error code. But the second condtion, `if(!pSolvedPkgInfo->nNeedDownload)` means that there are no packages to download and so we can simply return without error. That's why I placed them separately.

client/api.c
prestonsn: Added, thank you!

prestonsn:
> 1. Can we somehow create a test for negative scenario? Where available size is less than download size? It might be difficult but it's good to have it.
> 
I'm currently working on adding some functionality to the test code to mount a tmpfs filesystem with intentionally limited diskspace. Then supply its mounted path as the cache directory path to tdnf. This should trigger the test case as you mentioned.

tools/cli/lib/installcmd.c
prestonsn: Addressed in latest commit: https://github.com/vmware/tdnf/pull/321/commits/04949049afbb502572395207430dd169bde9bde2. Thanks!

client/defines.h
```diff
@@ -236,6 +236,7 @@ typedef enum
     {ERROR_TDNF_OPERATION_ABORTED, "ERROR_TDNF_OPERATION_ABORTED", "Operation aborted."},\
     {ERROR_TDNF_INVALID_INPUT, "ERROR_TDNF_INVALID_INPUT", "Invalid input."},\
     {ERROR_TDNF_CACHE_DISABLED, "ERROR_TDNF_CACHE_DISABLED", "cache only is set, but no repo data found"},\
+    {ERROR_TDNF_CACHE_DIR_OUT_OF_DISK_SPACE, "ERROR_TDNF_CACHE_DIR_OUT_OF_DISK_SPACE", "Insufficient disk space at cache directory /var/cache/tdnf (unless specified differently in config). Try freeing space first!"},\
```
oliverkurth: Nit - no exclamation mark please.

client/packageutils.c
oliverkurth: The second `if()` has no effect. When `dwError` is 0, `BAIL_ON_TDNF_ERROR` does nothing, you are not setting it before. So the rest of the code in the function is still executed.

I think you should put this check into the caller - don't call this function if `pSolvedPkgInfo->nNeedDownload` is `false`.

include/tdnferror.h
```diff
@@ -163,6 +165,14 @@ extern "C" {
...
 #define ERROR_TDNF_ACCESS_DENIED        (ERROR_TDNF_SYSTEM_BASE + EACCES)
 #define ERROR_TDNF_ALREADY_EXISTS       (ERROR_TDNF_SYSTEM_BASE + EEXIST)
+#define ERROR_TDNF_INVALID_ADDRESS      (ERROR_TDNF_SYSTEM_BASE + EFAULT)
+#define ERROR_TDNF_CALL_INTERRUPTED     (ERROR_TDNF_SYSTEM_BASE + EINTR)
+#define ERROR_TDNF_FILESYS_IO           (ERROR_TDNF_SYSTEM_BASE + EIO)
+#define ERROR_TDNF_SYM_LOOP             (ERROR_TDNF_SYSTEM_BASE + ELOOP)
+#define ERROR_TDNF_NAME_TOO_LONG        (ERROR_TDNF_SYSTEM_BASE + ENAMETOOLONG)
+#define ERROR_TDNF_CALL_NOT_SUPPORTED   (ERROR_TDNF_SYSTEM_BASE + ENOSYS)
+#define ERROR_TDNF_INVALID_DIR          (ERROR_TDNF_SYSTEM_BASE + ENOTDIR)
+#define ERROR_TDNF_OVERFLOW             (ERROR_TDNF_SYSTEM_BASE + EOVERFLOW)
```
oliverkurth, line 175: These are only needed if we set them ourselves. You can leave them there, it doesn't hurt.

pytests/tests/test_cache.py
```diff
@@ -111,3 +111,40 @@ def test_enable_repo_make_cache_verbose(utils):
...
+    clean_cache(utils)
+    enable_cache(utils)
+
+    cache_dir = utils.tdnf_config.get('main', 'cachedir')
+    pkgname = utils.config["mulversion_pkgname"]
+    utils.erase_package(pkgname)
+    ret = utils.run(['tdnf', 'install', '-y', '--nogpgcheck', pkgname])
+
+    down_bytes = utils.download_size_to_bytes(ret['stdout'])
+    cached_rpm_bytes = sum(utils.get_cached_package_sizes(cache_dir).values())
```
oliverkurth, line 168: Are you setting the flag to keep cached rpms somewhere? Or how do the files not get removed after installation? Maybe I am missing something.

pytests/tests/test_cache.py
```diff
@@ -111,3 +111,40 @@ def test_enable_repo_make_cache_verbose(utils):
...
+def test_download_vs_cache_size_multiple_packages(utils):
+    clean_cache(utils)
+    enable_cache(utils)
+
+    cache_dir = utils.tdnf_config.get('main', 'cachedir')
+    run_args = ['tdnf', 'install', '-y', '--nogpgcheck']
+    pkg_list = [
+        utils.config["mulversion_pkgname"],
+        utils.config["sglversion_pkgname"],
+        utils.config["sglversion2_pkgname"],
```
oliverkurth, line 182: You should remove all packages in `teardown_test`.

oliverkurth (commented):
Please also add the package size to the json info - should be trivial.

include/tdnferror.h
prestonsn: I initially had used linux's `errno` to get the error code, but that's been removed in a previous commit. As suggested, I'll leave them in.

client/defines.h
prestonsn: Removed, thanks!

pytests/tests/test_cache.py
oliverkurth: Nm, Preston clarified. This is done with `enable_cache()`.

prestonsn: `enable_cache()` call at the start of each test enable persistence of packages after the install. https://github.com/vmware/tdnf/blob/10857284658aeeb10b9fcd45270ef8fa05ff8a34/pytests/tests/test_cache.py#L28-L31

pytests/tests/test_cache.py
prestonsn: Added, Thanks!

client/packageutils.c
prestonsn: Addressed, Thanks!

prestonsn:
> Please also add the package size to the json info - should be trivial.

Added in latest commit: https://github.com/vmware/tdnf/pull/321/commits/e9c6d07b2f1611912950ff984fe5d59a1627a37d

prestonsn:
Latest commit https://github.com/vmware/tdnf/pull/321/commits/95605345421e6b525db59ecf89de5876f20afdb2 adds a test to intentionally trigger an out-of-diskspace error. It uses the new script `pytests/mount-small-cache.in` to mount the small directory. If this mount fails, then the test is skipped.

oliverkurth:
This looks good to me. Nice work! Please squash the commits together into a smaller number of commits that make sense on their own. [fb9b995](https://github.com/vmware/tdnf/pull/321/commits/fb9b995125cf75f0518ebb96eafc47a60bb0411a) could be worth its own PR, but should be left as a separate commit at least. Squashing all other commits into a single one is fine too.

I am wondering about corner cases:
- if the file is already downloaded to cache, would this check for remaining space even though download isn't needed?
- similar, if we do `tdnf install ./foo-1.2.3.rpm` or `tdnf install file:///some/path/foo-1.2.3.rpm` we don't download, so no need to check for the size. I don't think this is addressed.

This could be fixed in a separate PR, but if not addressed we should create an issue.

prestonsn:
> This looks good to me. Nice work! Please squash the commits together into a smaller number of commits that make sense on their own. [fb9b995](https://github.com/vmware/tdnf/pull/321/commits/fb9b995125cf75f0518ebb96eafc47a60bb0411a) could be worth its own PR, but should be left as a separate commit at least. Squashing all other commits into a single one is fine too.

Thank you for the advice! Everything has been squashed into 3 commits, and those mentioned changes have been merged in their own PR [#333](https://github.com/vmware/tdnf/pull/333).

> 
> I am wondering about corner cases:
> 
>     * if the file is already downloaded to cache, would this check for remaining space even though download isn't needed?
> 
>     * similar, if we do `tdnf install ./foo-1.2.3.rpm` or `tdnf install file:///some/path/foo-1.2.3.rpm` we don't download, so no need to check for the size. I don't think this is addressed.
> 
> 
> This could be fixed in a separate PR, but if not addressed we should create an issue.

Good point, I've opened issue https://github.com/vmware/tdnf/issues/334 and added some more details there.

PR #323, opened 2022-06-15
fix repoquery segfault

oliverkurth:
The `repoquery` command crashed when given more than one spec (arguments). For example:
```
# tdnf repoquery foo bar
Error(902) : Invalid argument.
Segmentation fault (core dumped)
```
Reason was trying to free an uninitialized `PTDNF_REPOQUERY_ARGS` pointer.

### DIFF
--- a/pytests/tests/test_repoquery.py
+++ b/pytests/tests/test_repoquery.py
@@ -206,3 +206,12 @@ def test_list1(utils):
                               'tdnf-repoquery-{}'.format(dep)
                               ])
     assert(ret['retval'] == 0)
+
+
+# fail with ore than one pkg spec
+def test_two_args(utils):
+    ret = utils.run(['tdnf',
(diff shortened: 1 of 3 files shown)


tapakund (commented):
I would also recommend initialising with NULL.  
PTDNF_REPOQUERY_ARGS pRepoqueryArgs. = NULL;
https://github.com/vmware/tdnf/blob/40851e79841fc3913a449309e49e5a1b934a46db/tools/cli/lib/api.c#L674

oliverkurth:
> I would also recommend initialising with NULL. PTDNF_REPOQUERY_ARGS pRepoqueryArgs. = NULL;
> 
> https://github.com/vmware/tdnf/blob/40851e79841fc3913a449309e49e5a1b934a46db/tools/cli/lib/api.c#L674

Yes, very good.

PR #326, opened 2022-06-17
ci: exit if one of the cmake/make commands fail

oliverkurth:
Somehow the `set -` was not enough to make `ci/docker-entrypoint.sh` fail on failure. adding an extra `exit` of one of the commands to build and check fails.


oliverkurth:
Proven to work, see PR #327 .

PR #327, opened 2022-06-17
add "mark" command 

### LINKED ISSUE #319: `autoremove` without args is not implemented
The `autoremove` feature will remove dependencies when a package is removed, but it does nothing when called without arguments, even when there are packages automatically installed:

```
# tdnf install lsof
Refreshing metadata for: 'VMware Photon Linux 4.0 (aarch64) Updates'
Refreshing metadata for: 'VMware Photon Linux 4.0 (aarch64)'
Refreshing metadata for: 'VMware Photon Extras 4.0 (aarch64)'
photon-extras                              123 100%
Installing:
libtirpc                                 aarch64              1.2.6-2.ph4                 photon-updates       193.33k 197970
lsof                                     aarch64              4.91-1.ph4                  photon-release       202.36k 207218
...

```

```
# tdnf remove --noautoremove lsof

Removing:
lsof                                     aarch64              4.91-1.ph4                  @System              202.36k 207218

Total installed size: 202.36k 207218
Is this ok [y/N]: y
Testing transaction
Running transaction
Removing: lsof-4.91-1.ph4.aarch64

Complete!
```

This should remove `libtirpc`:
```
# tdnf autoremove     
Nothing to do.
# 
```

oliverkurth:
PR https://github.com/vmware/tdnf/pull/314 will break the status of the autoremove flags for package. For example, a rollback will not restore the flags. This change moves tracking of the flag from the flat file `/var/lib/tdnf/autoinstalled `to the db, and tracks it for transactions.

This  also adds the `mark` command to mark a packages as "autoinstalled".

And it adds the `--userinstalled` option to the `repoquery` command to list all user installed packages.

And it implements the `autoremove` command without args to uninstall all autoinstalled packages. Closes #319 .

The command `tdnf mark remove foo` will mark a package as automatically installed, and `tdnf mark install foo` will reset that flag. The flag setting will be part of transactions, so can be rolled back or undone/redone with the `history` commands.

The syntax is the same as for the dnf command, see https://dnf.readthedocs.io/en/latest/command_ref.html#mark-command-label , only we do not have the `group` label.

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -55,6 +55,7 @@ target_link_libraries(${LIB_TDNF}
     ${METALINK_LIBRARIES}
     ${OPENSSL_LIBRARIES}
     ${LIBXML2_LIBRARIES}
+    ${SQLITE3_LIBRARIES}
 )
 
 set_target_properties(${LIB_TDNF} PROPERTIES
--- a/client/api.c
+++ b/client/api.c
@@ -183,13 +183,40 @@ TDNFAlterCommand(
     )
 {
     uint32_t dwError = 0;
+
+    UNUSED(nAlterType);
+
     if(!pTdnf || !pSolvedInfo)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
-    dwError = TDNFRpmExecTransaction(pTdnf, pSolvedInfo, nAlterType);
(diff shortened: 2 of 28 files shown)


oliverkurth:
Ugh, last commit (`autoremove`) needs to be revised - it would erase user installed packages too if they depend on auto installed packages. I will revise this.

oliverkurth:
> Ugh, last commit (`autoremove`) needs to be revised - it would erase user installed packages too if they depend on auto installed packages. I will revise this.

Fixed now.

Issue #209, opened 2021-01-11
tdnf clean options not implemented


keerthanakalyan:
tdnf clean has only all option implemented. We have others such as packages, metadata, dbcache, plugins, expire-cache. Would be good to have each command implemented.

oliverkurth:
PR #329 is merged, closing issue.

PR #329, opened 2022-06-23
add clean options and refactor

oliverkurth:
Addresses issue #209 . This adds the options `metadata, dbcache, packages, keys and expire-cache` to the `clean` command. Example output:
```
# ./bin/tdnf clean all
cleaning photon: metadata dbcache packages keys expire-cache
cleaning photon-debuginfo: metadata dbcache packages keys expire-cache
cleaning photon-iso: metadata dbcache packages keys expire-cache
cleaning photon-updates: metadata dbcache packages keys expire-cache
cleaning photon-extras: metadata dbcache packages keys expire-cache
cleaning photon-release: metadata dbcache packages keys expire-cache
Done.
```
This will clean all repos, including the disabled repositories. This is the same behavior as the `dnf clean` command. The `keys` option however is unique to `tdnf`.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -410,7 +410,7 @@ TDNFCheckUpdates(
 uint32_t
 TDNFClean(
     PTDNF pTdnf,
-    TDNF_CLEANTYPE nCleanType
+    uint32_t nCleanType
     )
 {
     uint32_t dwError = 0;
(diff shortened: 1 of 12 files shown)


client/api.c
```diff
@@ -410,76 +410,67 @@ TDNFCheckUpdates(
...
+    for (pRepo = pTdnf->pRepos; pRepo; pRepo = pRepo->pNext)
     {
-        while(*ppszReposUsed)
+        if (strcmp(pRepo->pszId, CMDLINE_REPO_NAME) == 0)
         {
-            dwError = TDNFRepoRemoveCache(pTdnf,*ppszReposUsed);
+            continue;
+        }
+        pr_info("cleaning %s:", pRepo->pszId);
+        if (nCleanType == CLEANTYPE_METADATA || nCleanType == CLEANTYPE_ALL)
```
sshedi: Instead of enum for `CLEANTYPE_*` let's use bit for values:
```
CLEANTYPE_NONE = 0x0
CLEANTYPE_PACKAGES = 0x1
CLEANTYPE_METADATA = 0x2
CLEANTYPE_DBCACHE = 0x4
CLEANTYPE_PLUGINS = 0x8
CLEANTYPE_EXPIRE_CACHE = 0x10
CLEANTYPE_RPMDB = 0x20
CLEANTYPE_ALL = 0xFF
```
And here in this function:
```
if (nCleanType & CLEANTYPE_METADATA) {
...
}
if (nCleanType & CLEANTYPE_DBCACHE) {
...
}
...
```

This is cleaner and scales well. And no need to compare every time with `CLEANTYPE_ALL`

sshedi (changes requested):
Changes LGTM, there is room for further improvement.

client/api.c
oliverkurth: Sure, I can do that. I thought about this before, but thought it's not really worth the effort because we either have just one item or all, never a set of 2 or more.

sshedi (approved):

prestonsn (approved):

Issue #253, opened 2021-08-04
`--releasever` uses cached repo data from other releases


oliverkurth:
For `repolist`, it seems to do the right thing:
```
# tdnf --releasever=3.0 repolist
repo id             repo name                               status    
photon-extras       VMware Photon Extras 3.0 (x86_64)       enabled   
photon-release      VMware Photon Linux 3.0 (x86_64)        enabled   
photon-updates      VMware Photon Linux 3.0 (x86_64) Updatesenabled
```
However, when trying to install, assuming you have a cache that's up-to-date it would use whatever is the cached data, which is most likely but not necessarily the release in the host:
```
# tdnf --releasever=3.0 install lsof

Upgrading:
lsof                                        x86_64                4.91-1.ph4                    photon-release        198.47k 203234

Total installed size: 198.47k 203234
Is this ok [y/N]: 
```

Even worse, if I refresh the cache for another release using `--releasever`, it would install from that release afterwards even if the option isn't given (all example are in Photon 4):
```
root@photon-edc45aa81da2 [ ~ ]# tdnf clean all
Cleaning repos: photon-updates photon-release photon-extras
Cleaning up everything
root@photon-edc45aa81da2 [ ~ ]# tdnf --releasever=3.0 install lsof
Refreshing metadata for: 'VMware Photon Extras 3.0 (x86_64)'
Refreshing metadata for: 'VMware Photon Linux 3.0 (x86_64)'
Refreshing metadata for: 'VMware Photon Linux 3.0 (x86_64) Updates'

...
root@photon-edc45aa81da2 [ ~ ]# tdnf install lsof

Installing:
lsof                                        x86_64                4.91-1.ph3                    photon-release        196.10k 200810
```

oliverkurth:
Fixed with PR #330.

PR #330, opened 2022-07-02
use a unique path for cache data

oliverkurth:
Fixes issue #253 . Wit this change, `tdnf` uses a unique name for the top level cache directory, derived from a `sha256` hash of the `baseurl` or `metalink` URL, after expansion of variables (like `$releasever`). This ensures different paths if `$releasever` is set using the `--releasever` option.

Example from the original issue report:

```
root [ /build/bld-ph4 ]# ./bin/tdnf makecache
Refreshing metadata for: 'VMware Photon Linux 4.0 (aarch64) Updates'
Refreshing metadata for: 'VMware Photon Linux 4.0 (aarch64)'
Refreshing metadata for: 'VMware Photon Extras 4.0 (aarch64)'
Metadata cache created.                    123 100%
root [ /build/bld-ph4 ]# ./bin/tdnf --releasever=3.0 install lsof
Refreshing metadata for: 'VMware Photon Linux 3.0 (aarch64) Updates'
Refreshing metadata for: 'VMware Photon Linux 3.0 (aarch64)'
Refreshing metadata for: 'VMware Photon Extras 3.0 (aarch64)'
photon-extras                              123 100%
Installing:
libtirpc                                            aarch64                   1.1.4-1.ph3                        photon-release               180.60k 184935
lsof                                                aarch64                   4.91-1.ph3                         photon-release               195.94k 200642

Total installed size: 376.54k 385577
Is this ok [y/N]: N
```
From the above, two different sets of cache directories were created:
```
# ls /var/cache/tdnf/
photon-extras-03310507f9e69f4ffd6371f07359c05de4d14c47b8ec056958a35cf2691430a5   photon-release-c58bea6dd9ed4dcf138eff5168526c61eb7a2cebe84d7292a6c97bc0c1d9a41a
photon-extras-89385989dce942383163eebfddbc353146801fa3f62b0dbfa11a1c6ca86a5811   photon-updates-ba5e6bf96d92bf6b703fccd7b1816b7ac57e83d7abcdb631ab2208dc4032af32
photon-release-2d2a0e933fae8240c8f9d53bc467764813771d913d32f7b80a03a415b2fb60f7  photon-updates-bec13383a1df049249f68a4ffff75eec7304d8464739a758831dd1956e1fdde0
```
This behavior is also similar to dnf.

Additionally, `tdnf clean all` will remove the top level repo cache dir as well.

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -679,7 +679,6 @@ TDNFRepoListFinalize(
                                                   &pRepo->pszCacheName);
             }
             BAIL_ON_TDNF_ERROR(dwError);
-            printf("pszCacheName = %s\n", pRepo->pszCacheName);
         }
     }
 cleanup:
@@ -803,6 +802,7 @@ TDNFFreeReposInternal(
         TDNF_SAFE_FREE_STRINGARRAY(pRepo->ppszUrlGPGKeys);
         TDNF_SAFE_FREE_MEMORY(pRepo->pszUser);
         TDNF_SAFE_FREE_MEMORY(pRepo->pszPass);
+        TDNF_SAFE_FREE_MEMORY(pRepo->pszCacheName);
         pRepos = pRepo->pNext;
         TDNF_SAFE_FREE_MEMORY(pRepo);
     }
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -645,42 +645,38 @@ TDNFRepoListFinalize(
     }
 
     /* Now that the overrides are applied, replace config vars
-       for the repos that are enabled. */
+       for all repos. */
     for(pRepo = pTdnf->pRepos; pRepo; pRepo = pRepo->pNext)
     {
-//        if(pRepo->nEnabled)
-        if (1)
+        if(pRepo->pszName)
         {
-            if(pRepo->pszName)
-            {
-                dwError = TDNFConfigReplaceVars(pTdnf, &pRepo->pszName);
-                BAIL_ON_TDNF_ERROR(dwError);
-            }
-            if(pRepo->pszBaseUrl)
-            {
-                dwError = TDNFConfigReplaceVars(pTdnf, &pRepo->pszBaseUrl);
-                BAIL_ON_TDNF_ERROR(dwError);
-            }
-            if(pRepo->pszMetaLink)
-            {
-                dwError = TDNFConfigReplaceVars(pTdnf, &pRepo->pszMetaLink);
-                BAIL_ON_TDNF_ERROR(dwError);
-            }
-
-            if (pRepo->pszMetaLink)
-            {
-                dwError = SolvCreateRepoCacheName(pRepo->pszId,
-                                                  pRepo->pszMetaLink,
-                                                  &pRepo->pszCacheName);
-            }
-            else if (pRepo->pszBaseUrl)
-            {
-                dwError = SolvCreateRepoCacheName(pRepo->pszId,
-                                                  pRepo->pszBaseUrl,
-                                                  &pRepo->pszCacheName);
-            }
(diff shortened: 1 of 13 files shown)


solv/tdnfrepo.c
```diff
@@ -377,6 +379,55 @@ SolvCalculateCookieForFile(
...
+SolvCreateRepoCacheName(
+    const char *pszName,
+    const char *pszUrl,
+    char **ppszCacheName
+    )
+{
+    uint32_t dwError = 0;
+    Chksum *pChkSum = NULL;
+    unsigned char pCookie[SOLV_COOKIE_LEN] = {0};
+    char pszCookie[SOLV_COOKIE_LEN*2+1];
```
sshedi: initialize this buffer to zero

client/repolist.c
```diff
@@ -648,7 +648,8 @@ TDNFRepoListFinalize(
        for the repos that are enabled. */
     for(pRepo = pTdnf->pRepos; pRepo; pRepo = pRepo->pNext)
     {
-        if(pRepo->nEnabled)
+//        if(pRepo->nEnabled)
+        if (1)
```
sshedi: ???why

client/repoutils.c
```diff
@@ -229,27 +226,62 @@ TDNFRepoGetRpmCacheDir(
...
+    }
+
+    dwError = TDNFGetCachePath(pTdnf, pRepo,
+                               NULL, NULL,
+                               &pszRepoCacheDir);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (rmdir(pszRepoCacheDir) != 0)
+    {
+        /* ignore ENOTEMPTY, let's keep the dir if it's not empty */
```
sshedi: I think we should not ignore not empty, it might lead to partial clean ups

sshedi (approved):
Good catch. But I think the use case is not valid though. If we use --installroot along with --releasever it works without any issues.

I haven't come across a single use case so far where we install ph3 package in ph4 and vice versa.

LGTM.

client/repolist.c
oliverkurth: Oops, I forgot to clean that up.

We need to loop through all repos, even disabled ones, to expand the URLs which we need to determine the cache path.

client/repoutils.c
oliverkurth: It only makes sense to remove the directory when it's empty. And this *is* executed when we do partial cleanups (something other than `all`), so leaving stuff behind is intended.

Maybe we can print a warning when the arg is `all` but the dir is not empty. Because this means either that we missed something, or the user put some files there.

oliverkurth:
> Good catch. But I think the use case is not valid though. If we use --installroot along with --releasever it works without any issues.

It works because now we use the `installroot` for the cache directory. Before it was in the host, IIRC that is when I found this issue. Fixing it is less urgent now but still makes sense.

There is also the corner case where the URL is changed, pointing to another repository, with potentially different metadata. With this change, we create another cache directory instead of using the old one.

oliverkurth:
A request for comment: how about shortening the hash that is used? Currently it uses the the full sha256 checksum, making a very long path. We could shorten this to the first 8 chars (4 bytes). The chance of a collision is low (1 in 2^32 ~ 4 billion). This is not a security issue.

Note that git uses the first 7 chars to identify commits.

sshedi:
> A request for comment: how about shortening the hash that is used? Currently it uses the the full sha256 checksum, making a very long path. We could shorten this to the first 8 chars (4 bytes). The chance of a collision is low (1 in 2^32 ~ 4 billion). This is not a security issue.
> 
> Note that git uses the first 7 chars to identify commits.

Makes sense. +1

solv/tdnfrepo.c
oliverkurth: Done

PR #331, opened 2022-07-07
python: Fix linking to built libraries

Conan-Kudo:
We incorrectly use RPATH here when we actually want to link
to the libraries we produced earlier in the build.

Usage of RPATH also causes package builds in Fedora to fail,
since RPATHs are forbidden.

### DIFF
--- a/pytests/CMakeLists.txt
+++ b/pytests/CMakeLists.txt
@@ -14,7 +14,7 @@ configure_file(
 )
 
 add_custom_target(check
-    COMMAND pytest
+    COMMAND ${CMAKE_COMMAND} -E env LD_LIBRARY_PATH="${CMAKE_LIBRARY_OUTPUT_DIRECTORY}" pytest
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     COMMENT "Running tests.."
 )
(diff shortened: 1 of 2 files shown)


oliverkurth:
Thanks @Conan-Kudo . Can you give a use case that fails and is fixed with this change?

Conan-Kudo:
Since Fedora Linux 35, package builds of TDNF itself fail because of an invalid RPATH for the Python bindings library: https://fedoraproject.org/wiki/Changes/Broken_RPATH_will_fail_rpmbuild

oliverkurth (approved):

oliverkurth:
Thanks for your contribution!

@Conan-Kudo , I noticed that there is a Fedora build for tdnf at https://copr.fedorainfracloud.org/coprs/ngompa/tdnf/build/4605225/ - is this available in any Fedora repo? I would like to play with it.

Conan-Kudo:
> Thanks for your contribution!
> 
> @Conan-Kudo , I noticed that there is a Fedora build for tdnf at https://copr.fedorainfracloud.org/coprs/ngompa/tdnf/build/4605225/ - is this available in any Fedora repo? I would like to play with it.

It is available by enabling the COPR:

```
$ sudo dnf copr enable ngompa/tdnf
$ sudo dnf install tdnf
```

PR #332, opened 2022-07-10
Correctly mark internal libraries to be built as static libraries

Conan-Kudo:
We were relying on CMake having a particular default preferred
behavior, but this causes problems when building tdnf RPMs on
Fedora, where CMake's default behavior is to produce shared libraries.

Explicitly set up these internal libraries as static libraries to
resolve the issue.


oliverkurth:
This looks good to me.

PR #333, opened 2022-07-11
Fix missing error when repo build script setup-repo.sh fails.

prestonsn:
Adds error-checking to the script `pytests/repo/setup-repo.sh` as well as a `pytest.exit()` call when the script fails to run. Currently, no error is printed and the tests run regardless of failing to build the test repository. 

Added as a separate PR as per https://github.com/vmware/tdnf/pull/321#issuecomment-1179221201.


oliverkurth:
Very nice!

PR #336, opened 2022-07-22
Fix coverity issues

oliverkurth:
Fixing most issues found by coverity.

Also working around a coverity bug triggered by gcc >= 10.0 - the version isn't parsed correctly. 

Remaining issues:
* 5 instances of `PARSE_ERROR` in `stdlib.h` (we have no control over)
* 7 instances of `PW.BRANCH_PAST_INITIALIZATION` that have no impact
* 2 instances of `PW.INCLUDE_RECURSION` in `libxml` headrers, again not under our control

### DIFF
--- a/ci/coverity.sh
+++ b/ci/coverity.sh
@@ -16,6 +16,14 @@ CC=cc
 
 cov-configure --config ${COV_CONFIG} --compiler ${CC} --comptype gcc --template
 cov-build --dir ${COV_DIR} --config ${COV_CONFIG} make
+
+# gcc >= 10.0 workaround, see https://www.mail-archive.com/gcc@gcc.gnu.org/msg93814.html
+if grep -qr 'version>0' . --include=*.xml ; then
+    for f in $(grep -lr 'version>0' . --include=*.xml) ; do sed -i 's/version>0/version>10/' $f; done
+    make clean
+    cov-build --dir ${COV_DIR} --config ${COV_CONFIG} make
+fi
+
 cov-analyze --dir ${COV_DIR} --config ${COV_CONFIG} --all
 
 mkdir html
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -112,7 +112,8 @@ set_progress_cb(
     BAIL_ON_TDNF_CURL_ERROR(dwError);
 
     memset(&pData, 0, sizeof(pcb_data));
-    strcpy(pData.pszData, pszData);
+    strncpy(pData.pszData, pszData, sizeof(pData.pszData) - 1);
+    /* coverity[bad_sizeof] */
     dwError = curl_easy_setopt(pCurl, CURLOPT_XFERINFODATA, &pData);
     BAIL_ON_TDNF_CURL_ERROR(dwError);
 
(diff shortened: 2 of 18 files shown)


ci/coverity.sh
```diff
@@ -16,6 +16,14 @@ CC=cc
 
 cov-configure --config ${COV_CONFIG} --compiler ${CC} --comptype gcc --template
 cov-build --dir ${COV_DIR} --config ${COV_CONFIG} make
+
+# gcc >= 10.0 workaround, see https://www.mail-archive.com/gcc@gcc.gnu.org/msg93814.html
+if grep -qr 'version>0' . --include=*.xml ; then
```
tapakund, line 21: shouldn't we check for gcc version before doing this? Also, can you share the xml file path.

client/remoterepo.c
```diff
@@ -368,12 +369,19 @@ TDNFCheckRepoMDFileHashFromMetalink(
...
+    {
+        dwError = TDNFChecksumFromHexDigest(hashInfo->value, digest);
+        BAIL_ON_TDNF_ERROR(dwError);
 
+        dwError = TDNFCheckHash(pszFile, digest, hash_Type);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+    else
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
```
tapakund: should we put more accurate error here?

history/history.c
```diff
@@ -1469,9 +1465,10 @@ int history_update_state(struct history_ctx *ctx, rpmts ts, const char *cmdline)
     history_set_cookie(ctx, cookie);
     ctx->trans_id = trans_id;
 error:
-    if (rc)
+    if (rc) {
         sqlite3_exec(ctx->db, "ROLLBACK;", 0, 0, NULL);
-    else
+        safe_free(current_ids);
+    } else
```
tapakund, line 1471: nit: better to add braces for else as well.

tools/cli/main.c
```diff
@@ -262,19 +262,25 @@ TDNFCliShowVersion(
     PTDNF_CMD_ARGS pCmdArgs
     )
 {
+    uint32_t dwError = 0;
```
tapakund, line 265: why do we need dwError here at all?

tools/cli/main.c
```diff
@@ -262,19 +262,25 @@ TDNFCliShowVersion(
...
+        CHECK_JD_RC(jd_map_add_string(jd, "Name", TDNFGetPackageName()));
+        CHECK_JD_RC(jd_map_add_string(jd, "Version", TDNFGetVersion()));
         pr_json(jd->buf);
-        jd_destroy(jd);
     }
     else
     {
         pr_info("%s: %s\n", TDNFGetPackageName(), TDNFGetVersion());
     }
+error:
```
tapakund, line 281: are we doing goto error anywhere from this fn?

tapakund (commented):
few comments and queries.

ci/coverity.sh
oliverkurth: The `if` checks if the version that coverity found starts with `0`, which is the bug when the version is actually `10`.

The xml files are in `${COV_DIR}/emit`, `coverity-intermediate/emit/` under the build directory with our settings in the script.

tools/cli/main.c
oliverkurth: It's used in the `CHECK_JD_NULL/CHECK_JD_RC` macros.

tools/cli/main.c
oliverkurth: Yes, again in the macros.

client/remoterepo.c
oliverkurth: Changing it to `ERROR_TDNF_INVALID_REPO_FILE` - I think that's better, from looking at the code preceding this.

history/history.c
oliverkurth: Different coding style in that file.

ci/coverity.sh
tapakund: instead of directly searching the version in xml file, why don't we do this if the gcc version if greater than or equal to 10?

oliverkurth: The `if grep -qr 'version>0' . --include=*.xml` checks for the actual bug. So if coverity fixes the bug we wouldn't do the workaround either. So this is more reliable than just testing for the gcc version.

tapakund (approved):

oliverkurth:
Thanks @tapakund for reviewing!

PR #340, opened 2022-08-11
various minor fixes

oliverkurth:
A few smaller issues.

### DIFF
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -668,10 +668,16 @@ TDNFDownloadFile(
     }
     else
     {
-        dwError = rename(pszFileTmp, pszFile);
-        BAIL_ON_TDNF_ERROR(dwError);
-        dwError = chmod(pszFile, S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH);
-        BAIL_ON_TDNF_ERROR(dwError);
+        if (rename(pszFileTmp, pszFile) == -1)
+        {
+            dwError = errno;
+            BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+        }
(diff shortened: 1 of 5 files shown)


client/remoterepo.c
```diff
@@ -668,8 +668,12 @@ TDNFDownloadFile(
...
     {
-        dwError = rename(pszFileTmp, pszFile);
-        BAIL_ON_TDNF_ERROR(dwError);
+        if (rename(pszFileTmp, pszFile) == -1)
+        {
+            dwError = errno;
+            BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+        }
+
         dwError = chmod(pszFile, S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH);
```
tapakund: we should also do the same for chmod. if we get ret val as -1, look at errno and set dwError accordingly and do BAIL_ON_TDNF_SYSTEM_ERROR

tools/cli/main.c
```diff
@@ -162,6 +162,7 @@ int main(int argc, char **argv)
         {
             TDNFCliShowUsage();
         }
+        dwError = ERROR_TDNF_CLI_NO_SUCH_CMD;
     }
```
tapakund, line 167: we do not need to BAIL_ON_CLI_ERROR after setting dwError?

oliverkurth: Thank you, you are correct. I'll fix it.

client/remoterepo.c
oliverkurth: How could I miss that? Thank you! Fixed.

tapakund (approved):
looks good.

PR #343, opened 2022-09-08
handle error when install root is relative

oliverkurth:
The install root needs to be relative. Print a better error message than "Unknown error".


oliverkurth:
Test failures are due to rpm key import issues and unrelated to this change.

sshedi (approved):

PR #348, opened 2022-09-23
support multiple base URLs

oliverkurth:
This adds support for multiple base URLs in repository configurations. The URLs will be tried in the order given.

Other changes are
* passing a pointer to the repository structure instead of the repo id for a lot of functions - this is more efficient than searching all repositories for the matching name
* simplifying some of the metalink logic when fetching repo metadata

This change is a prerequisite for an upcoming change to put the metalink functionality into a plugin.


ntsbtz (commented):
LGTM as review comments has been already taken care

PR #349, opened 2022-09-26
move metalink functionality to a plugin

oliverkurth:
This moves the `metalink` functionality out of the main code to a plugin. This moves the dependency on `libxml2` to the plugin as well.

This PR includes changes from https://github.com/vmware/tdnf/pull/348 .

Commits need to be squashed.

TODO:
* handle unique repo cache directory naming (without baseurl) (does not need to in this change)


ntsbtz (commented):
LGTM as review comments has been already taken care

PR #357, opened 2022-10-18
add utility to inititialize history and mark packages

oliverkurth:
During an install of `tdnf` we cannot run `tdnf` in its own `rpm` scriptlets. This adds a simple utility that can be used to initialize the history db and convert the old file `/var/lib/tdnf/autoinstalled`.


oliverkurth:
This was part of a patch in the Photon package which was reviewed.

PR #346, opened 2022-09-17
add --skip-broken option

oliverkurth:
Add the `--skip-broken` option, addresses issue #342 .

This will also enable `--skipconflicts` and `--skipobsoletes` for all commands, not just `check`.

PR #361, opened 2022-10-27
fix skip options

oliverkurth:
The options `--skipobsoletes` and `--skipconflicts` options were being ignored. This has been broken with https://github.com/vmware/tdnf/pull/346.


oliverkurth:
Part of a patch in the Photon package, reviewed.

PR #363, opened 2022-11-04
Fix crash in rpmtsCheck() if rpmtsRun() fails

reubeno:
The preexisting logic in `doCheck` is very helpful for extracting and displaying problems, but the call to `rpmtsCheck` itself should not be re-executed if the failure occurred after the initial check. Once `rpmtsClean()` has been called, data within `pTS` has been emptied and `rpmtsCheck` will segfault on NULL dereference.

Also address a few potential issues with the lifetime of the problem objects:
* The `rpmps` object is ref-counted internally and should be released with `rpmpsFree`; the latter function frees all of the problems within it when the ref count reaches 0.
* If multiple problems are reported, then `pErrorStr` needs to be freed multiple times.

---

_NOTE: This is my first (attempted) contribution to this project, so I wasn't sure how/which tests to re-run. I'd appreciate any and all feedback you have for me to make this PR acceptable._

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -549,12 +549,31 @@ int
 doCheck(PTDNFRPMTS pTS)
 {
     int nResult = 0;
-    rpmpsi psi = NULL;
-    rpmProblem prob = NULL;
+
     nResult = rpmtsCheck(pTS->pTS);
-    char *pErrorStr = NULL;
 
     rpmps ps = rpmtsProblems(pTS->pTS);
+    if(ps)
+    {
+        int nProbs = rpmpsNumProblems(ps);
+        if(nProbs > 0)
+        {
+            nResult = ERROR_TDNF_RPM_CHECK;
+        }
+        rpmpsFree(ps);
+    }
+
+    return nResult;
+}
+
+void
+reportProblems(PTDNFRPMTS pTS)
+{
+    rpmps ps = NULL;
+    rpmpsi psi = NULL;
+    char *pErrorStr = NULL;
+
+    ps = rpmtsProblems(pTS->pTS);
     if(ps)
     {
         int nProbs = rpmpsNumProblems(ps);
@@ -565,7 +584,7 @@ doCheck(PTDNFRPMTS pTS)
             psi = rpmpsInitIterator(ps);
             while(rpmpsNextIterator(psi) >= 0)
             {
-                prob = rpmpsGetProblem(psi);
+                rpmProblem prob = rpmpsGetProblem(psi);
                 char *msg = rpmProblemString(prob);
                 if (strstr(msg, "no digest") != NULL)
                 {
@@ -577,24 +596,32 @@ doCheck(PTDNFRPMTS pTS)
                     if (rpmProblemGetType(prob) == RPMPROB_REQUIRES)
                     {
                         uint32_t dwError = 0;
+
                         dwError = TDNFAllocateString(rpmProblemGetStr(prob), &pErrorStr);
                         BAIL_ON_TDNF_ERROR(dwError);
 
                         dwError = TDNFDetectPreTransFailure(pTS->pTS, pErrorStr);
                         BAIL_ON_TDNF_ERROR(dwError);
+
+                        TDNF_SAFE_FREE_MEMORY(pErrorStr);
                     }
                 }
-                rpmProblemFree(prob);
             }
-            rpmpsFreeIterator(psi);
-            nResult = ERROR_TDNF_RPM_CHECK;
         }
     }
 cleanup:
     TDNF_SAFE_FREE_MEMORY(pErrorStr);
-    return nResult;
+    if (psi)
+    {
+        rpmpsFreeIterator(psi);
+    }
+    if (ps)
+    {
+        rpmpsFree(ps);
+    }
+    return;
+
 error:
-    nResult = ERROR_TDNF_RPM_CHECK;
     goto cleanup;
 }
 
@@ -678,9 +705,9 @@ cleanup:
     return dwError;
 
 error:
-    if(pTS && dwError != ERROR_TDNF_RPM_CHECK)
+    if(pTS)
     {
-        doCheck(pTS);
+        reportProblems(pTS);
     }
     goto cleanup;
 }


client/rpmtrans.c
```diff
@@ -576,26 +593,23 @@ doCheck(PTDNFRPMTS pTS)
...
                     if (rpmProblemGetType(prob) == RPMPROB_REQUIRES)
                     {
+                        char *pErrorStr = NULL;
                         uint32_t dwError = 0;
+
                         dwError = TDNFAllocateString(rpmProblemGetStr(prob), &pErrorStr);
-                        BAIL_ON_TDNF_ERROR(dwError);
+                        if (!dwError) {
+                            dwError = TDNFDetectPreTransFailure(pTS->pTS, pErrorStr);
+                        }
```
oliverkurth: Did you make this change to avoid calling `reportProblems()` if this fails?

client/rpmtrans.c
```diff
@@ -678,9 +692,9 @@ TDNFRunTransaction(
     return dwError;
 
 error:
-    if(pTS && dwError != ERROR_TDNF_RPM_CHECK)
+    if(pTS)
     {
-        doCheck(pTS);
+        reportProblems(pTS);
```
oliverkurth, line 710: I wonder if we should move this to right after the two `rpmtsRun()` calls. If it's true that only `rpmtsRun()` and `rpmtsOrder()` generate problems.

oliverkurth (commented):
Do you have steps to reproduce the segfault issue?

reubeno:
This isn't exactly what I was doing--and is a bit artificial to produce an error--but below is a self-contained way to reproduce the segfault. I'm intentionally making sure the root is readonly (such that package install will necessarily fail) and mounting in a tmpfs for /run and /var so that `tdnf` can get far enough to see the install failure. (Since I'm throwing caution to the wind and mounting an empty fs over /var, I know this destroys the RPM database; that's not what I'm doing in the scenario that caused me to find this issue, but it was a convenient way to reproduce the problem for you.)

```
$ docker run -it --rm --read-only --mount type=tmpfs,destination=/run --mount type=tmpfs,destination=/var photon bash -c "mkdir -p /var/run; tdnf install -y vim --releasever=4.0; bash"
Refreshing metadata for: 'VMware Photon Extras 4.0 (x86_64)'
Refreshing metadata for: 'VMware Photon Linux 4.0 (x86_64) Updates'
photon-release                          342894 100%
Installing:
pcre-libs                                                  x86_64                        8.44-4.ph4                              photon-updates                     480.89k 492434
libselinux                                                 x86_64                        3.3-1.ph4                               photon-updates                     170.36k 174450
libgcc                                                     x86_64                        10.2.0-2.ph4                            photon-updates                     100.82k 103240
grep                                                       x86_64                        3.4-1.ph4                               photon-release                     229.47k 234978
gmp                                                        x86_64                        6.2.0-1.ph4                             photon-release                     532.45k 545232
openssl                                                    x86_64                        3.0.7-1.ph4                             photon-updates                      6.19M 6486800
coreutils-selinux                                          x86_64                        9.1-1.ph4                               photon-updates                      6.97M 7311209
filesystem                                                 x86_64                        1.1-4.ph4                               photon-release                         7.19k 7363
glibc                                                      x86_64                        2.32-11.ph4                             photon-updates                      8.72M 9142716
readline                                                   x86_64                        7.0-3.ph4                               photon-release                     371.60k 380520
bash                                                       x86_64                        5.0-3.ph4                               photon-updates                      1.85M 1935927
ncurses-libs                                               x86_64                        6.2-6.ph4                               photon-updates                     625.83k 640848
vim                                                        x86_64                        8.2.5169-1.ph4                          photon-updates                      3.47M 3643511

Total installed size:  29.66M 31099228

Downloading:
pcre-libs                               182887 100%
importing key from file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
importing key from file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY-4096
libselinux                               80699 100%
libgcc                                   56407 100%
grep                                    124107 100%
gmp                                     272614 100%
openssl                                2313969 100%
coreutils-selinux                      1415902 100%
filesystem                               21351 100%
glibc                                  2749414 100%
readline                                147551 100%
bash                                    587946 100%
ncurses-libs                            259996 100%
vim                                    1611111 100%
Testing transaction
bash: line 1:     8 Segmentation fault      (core dumped) tdnf install -y vim --releasever=4.0
root [ / ] #
```

client/rpmtrans.c
reubeno: With this function now just reporting problems and no longer being responsible for validation/checking, it seems to me that it could now be "best effort", i.e., report on as many problems as it can manage to format instead of bailing early if something, say, goes wrong in `TDNFDetectPreTransFailure`.

What's your thinking?

client/rpmtrans.c
reubeno: I'm not sure I follow. Are you suggesting moving it to just before the `cleanup` label? Or else placing a call to `reportProblems` in each of the individual failure paths near the failed call sites? Is there a reason that we need to avoid calling it in other cases? I do appreciate that, with where it is now, changes to the core logic in the body of the function won't introduce cases where something could fail above and generate problems without the problems being reported on.

With that said, if you have guidance as to what you're looking for or a change that you'd like me to make, please do let me know.

oliverkurth: > I'm not sure I follow. Are you suggesting moving it to just before the cleanup label?

No

> Or else placing a call to reportProblems in each of the individual failure paths near the failed call sites?

Yes, this.

> Is there a reason that we need to avoid calling it in other cases?

Yes. We get to the error label for all sorts of errors, including memory allocation errors or other system errors. Thus my comment in the code ("Did you make this change to avoid...). But we only want to report problems when there are actually rpm problems.

My comment is outside of the scope of the issue you are addressing though - it's independent, we could address it in another change.

reubeno: Okay, I understand now. Thanks for clarifying. Would you like me to file an issue for separate tracking of this feedback? (I'm not sure what process this project follows for out-of-scope PR feedback like this.)

client/rpmtrans.c
oliverkurth: I think this needs reverting - use `BAIL_ON_TDNF_ERROR(dwError);` and remove the `if`, as it was before. We can later address calling `reportProblems()` when it's not an RPM error.

client/rpmtrans.c
oliverkurth: Sure, you can do that.

client/rpmtrans.c
reubeno: Okay, will do. I will try to fix the free/cleanup logic on the bail case, then, to make sure the iterator et al. aren't potentially leaked.

client/rpmtrans.c
reubeno: Okay, filed as #364.

client/rpmtrans.c
reubeno: I've pushed an update that reverts this change and moves around the cleanup logic in that function to address the leak I mention above. Thanks for the feedback.

client/rpmtrans.c
```diff
@@ -577,24 +596,32 @@ doCheck(PTDNFRPMTS pTS)
...
-    nResult = ERROR_TDNF_RPM_CHECK;
+    TDNF_SAFE_FREE_MEMORY(pErrorStr);
+    if (psi)
+    {
+        rpmpsFreeIterator(psi);
+    }
+    if (ps)
+    {
+        rpmpsFree(ps);
+    }
```
oliverkurth: I think we should always free these, not just in the error case.

reubeno: You're absolutely correct. My apologies, that was definitely unintended. I missed that with the churn from the last round of changes--thanks for catching it.

I've pushed an update addressing this.

reubeno:
FYI, it looks like I don't have permission to run the workflows, so I can't tell if the latest version of the PR will pass checks. If one of the maintainers is able to run them on behalf, I'd appreciate it!

oliverkurth:
No problem, I just approved the workflow runs. I think as soon as you have contributed it runs automatically. Not sure if I can do anything before.

oliverkurth:
Thanks for your contribution @reubeno !

Issue #373, opened 2022-11-22
metalink plugin uses wrong checksum for verification


oliverkurth:
### Describe the bug

A metalink file can have multiple checksums. If I understand correctly, these can be tried in order. For example, this is the file for current Fedora 37:
```
<?xml version="1.0" encoding="utf-8"?>
<metalink version="3.0" xmlns="http://www.metalinker.org/" type="dynamic" pubdate="Tue, 22 Nov 2022 00:01:23 GMT" generator="mirrormanager" xmlns:mm0="http://fedorahosted.org/mirrormanager">
 <files>
  <file name="repomd.xml">
   <mm0:timestamp>1668992591</mm0:timestamp>
   <size>8307</size>
   <verification>
    <hash type="md5">d7338c528c9612025dddd74e4467c0e6</hash>
    <hash type="sha1">bdf0de43e3551855d42af6f6a5eac4552ed20ca7</hash>
    <hash type="sha256">c096acb14db2b66ef4f696db98e7a64cdfa5cd8f251bc34998e767346515e620</hash>
    <hash type="sha512">b4e08a810afad386776ebb7f96e6b765c1f27a596776d0d3ac2e793f6139f5635e1ec6af5f8068b8e7c2ed064cf9509c3d837b43f373dee48d0bdb99ffc33356</hash>
   </verification>
   <mm0:alternates>
    <mm0:alternate>
      <mm0:timestamp>1668906925</mm0:timestamp>
      <size>8307</size>
      <verification>
       <hash type="md5">d74c5f729e207da802ce4c1b84d37f1c</hash>
       <hash type="sha1">30d4806e87ffbe6438373065b3184d3c615a5aa3</hash>
       <hash type="sha256">00d4d4cdbe19d238dd5b01dcd009b46fb692101c9b306044a4efcb0f56297961</hash>
       <hash type="sha512">914c3722e2b6826379d8ef2772b8d2fad6a2c4238bba58edd58875b579cfa5226b6f5517f05eada400737ef9d40490c49a52f697d53bb26713b7176ec90d2799</hash>
      </verification>
    </mm0:alternate>
    <mm0:alternate>
      <mm0:timestamp>1668825721</mm0:timestamp>
      <size>8304</size>
      <verification>
       <hash type="md5">54cb1f85848092bcdb15f444cab218f1</hash>
       <hash type="sha1">3ac34a6f871c99b5161b7aa7b23294085e7565be</hash>
[...]

PR #374, opened 2022-11-28
Fix metalink checksum verification (issue #373)

oliverkurth:
Find 'best' available checksum type and use all available checksums of this type in metalink file for verification. Fixes issue #373.

Also move a few bits from header files in core to the plugin.

### DIFF
--- a/plugins/metalink/utils.c
+++ b/plugins/metalink/utils.c
@@ -371,7 +371,6 @@ TDNFCheckRepoMDFileHashFromMetalink(
 {
     uint32_t dwError = 0;
     TDNF_ML_HASH_LIST *hashList = NULL;
-    TDNF_ML_HASH_INFO *hashInfo = NULL;
     unsigned char digest[EVP_MAX_MD_SIZE] = {0};
     int hash_Type = -1;
     TDNF_ML_HASH_INFO *currHashInfo = NULL;
@@ -383,6 +382,7 @@ TDNFCheckRepoMDFileHashFromMetalink(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    /* find best (highest) available hash type */
     for(hashList = ml_ctx->hashes; hashList; hashList = hashList->next)
     {
         int currHashType = TDNF_HASH_SENTINEL;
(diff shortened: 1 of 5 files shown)


plugins/metalink/utils.c
```diff
@@ -397,28 +397,42 @@ TDNFCheckRepoMDFileHashFromMetalink(
...
+        int currHashType = TDNF_HASH_SENTINEL;
+        currHashInfo = hashList->data;
+
+        dwError = TDNFGetResourceType(currHashInfo->type, &currHashType);
         BAIL_ON_TDNF_ERROR(dwError);
+
+        /* filter for our best type and also check that the value is valid */
+        if (hash_Type == currHashType &&
+            TDNFCheckHexDigest(currHashInfo->value, hash_ops[currHashType].length)) {
+            dwError = TDNFChecksumFromHexDigest(currHashInfo->value, digest);
```
ntsbtz, line 424: If current highest priority checksum is not valid then we should continue. That might be the case if we have multiple same priority cheksum the next one might be valid one.

Except above LGTM

ntsbtz (changes requested):

plugins/metalink/utils.c
oliverkurth: Okay, good point. I will change it.

oliverkurth: Thinking about this again - I think now the code is correct. The function `TDNFCheckHexDigest()` checks if the hash is syntactically correct, but not if it matches. We skip if not (this only happens if the metalink file is not correct), and go on to the next one (if any). The function `TDNFChecksumFromHexDigest()` retrieves the hash. If it fails, something serious is wrong and we bail out.

The function `TDNFCheckHash()` then finally checks if the checksum matches. In case it does not, it returns `ERROR_TDNF_CHECKSUM_VALIDATION_FAILED`, and we go on to the next one (if any). If it matches, it returns `0` and we break out of the loop - it means we found a match, no need to check any further. If it returns any thing else, something is wrong and we bail.

So, I don't see anything wrong with this logic.

PR #378, opened 2022-12-16
install a providing and obsoleting package when found

oliverkurth:
This uses `pool_whatprovides()` to find package providers and uses libsolv's `pool_best_solvables()` to determine the best available package when doing an `install` or `upgrade`. If no package is found using `pool_whatprovides()` (because the desired package spec contains a version or a relation), fall back to the old method, but still use `pool_best_solvables()` to determine the best result.

### DIFF
--- a/client/goal.c
+++ b/client/goal.c
@@ -332,7 +332,7 @@ TDNFSolv(
     dwError = TDNFSolvAddPkgLocks(pTdnf, pQueueJobs, pTdnf->pSack->pPool);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    dwError = TDNFSolvAddMinVersions(pTdnf, pQueueJobs, pTdnf->pSack->pPool);
+    dwError = TDNFSolvAddMinVersions(pTdnf, pTdnf->pSack->pPool);
     BAIL_ON_TDNF_ERROR(dwError);
 
     pSolv = solver_create(pTdnf->pSack->pPool);
@@ -893,7 +893,6 @@ error:
 uint32_t
 TDNFSolvAddMinVersions(
     PTDNF pTdnf,
-    Queue* pQueueJobs,
     Pool *pPool
     )
 {
@@ -903,7 +902,7 @@ TDNFSolvAddMinVersions(
     Map *pMapMinVersions = NULL;
     char *pszTmp = NULL;
 
-    if(!pTdnf || !pQueueJobs || !pPool)
+    if(!pTdnf || !pPool)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
--- a/client/prototypes.h
+++ b/client/prototypes.h
@@ -553,7 +553,6 @@ TDNFSolvAddPkgLocks(
 uint32_t
 TDNFSolvAddMinVersions(
     PTDNF pTdnf,
-    Queue* pQueueJobs,
     Pool *pPool
     );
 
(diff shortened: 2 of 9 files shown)


mlschroe:
1) never change the pool flags. They must reflect what your low-level package manager (i.e. rpm) does. Unless you're using rpm5, setting POOL_FLAG_IMPLICITOBSOLETEUSESPROVIDES is certainly wrong.

2) the pull request only looks at provides, not at obsoletes. I don't think this is what you're looking for.

mlschroe:
I would do it like this (untested):
```
/* check if package s obsoletes is */
int
is_obsoleting(Pool *pool, Solvable *s, Solvable *is)
{
  Id obs, *obsp;
  Id p, pp;
  Id n = is - pool->solvables;
  if (!s->obsoletes)
    return 0;
  obsp = s->repo->idarraydata + s->obsoletes;
  while ((obs = *obsp++) != 0)  /* for all obsoletes */
    {
      FOR_PROVIDES(p, pp, obs)   /* and all matching providers of the obsoletes */
        if (p == n && pool_match_nevr(pool, pool->solvables + p, obs))
          return 1;
    }
  return 0;
}

Id
find_best_package_for_name(Pool *pool, Id name)
{
  Queue q;
  Id p, pp;

  queue_init(&q);

  /* step 1: look at packages with the specified name */
  FOR_PROVIDES(p, pp, name)
    if (pool->solvables[p].name == name)
      queue_push(&q, p);
  if (!q.count)
    {
      queue_free(&q);
      return 0;
    }
  pool_best_solvables(pool, &q, 0);
  queue_truncate(&q, 1);        /* use the first element */

  /* step 2: check if a package has a provides/obsoletes combination
   * for the best package */
  FOR_PROVIDES(p, pp, name)
    {
      if (pool->solvables[p].name == name)
        continue;               /* already done in step 1 */
      if (!is_obsoleting(pool, pool->solvables + p, pool->solvables + q.elements[0]))
        continue;
      queue_push(&q, p);
    }
  pool_best_solvables(pool, &q, 0);
  p = q.elements[0];
  queue_free(&q);
  return p;
}
```

oliverkurth:
Thanks @mlschroe for the suggestion. I am not sure if I am missing something - I think the original code from the PR *does* look at the obsoletes, via `pool_best_solvables()` which calls ultimately `prune_obsoleted()` (see `src/policy.c`). I was assuming that this would do the right thing. If I understand your suggestion correctly, the additional check using `is_obsoleting()` is then not needed.

But you are the expert on libsolv and I am not. So I may well be wrong and I am missing something.

I did test the code from your comment, and it does work as expected (thanks!).

mlschroe:
pool_best_solvables() looks at the obsoletes for pruning purposes, it will not remove unrelated packages. I.e. it will remove a package from the list if there is another package in the list that obsoletes the package. So it's something very different.

oliverkurth:
Okay, got it. In other words, my PR would find provides for `name` even if they don't obsolete it, while your suggestion would only add those packages that also obsolete it. `pool_best_solvables()` would prune obsoleted packages, but would keep packages that don't obsolete `name`.

oliverkurth:
Updated the PR with @mlschroe 's suggestion. I will squash before merging.

oliverkurth:
I have made a simplification in `is_obsoleting()`. The inner loop isn't needed because we already know that `s` provides `is` because the loop in the caller goes through just the provides. @mlschroe let me know if I am missing anything. @sshedi , please test.

I made a few tests myself, testing various obsoleting/providing packages, and it works as expected.

PR #387, opened 2023-01-17
test error when trying to install pkg with missing require

oliverkurth:
Make sure we are getting a good error when trying to install a package that has a missing requirement. See also https://github.com/vmware/tdnf/pull/145 .


ntsbtz (approved):
The error was already reported so only adding test was missing hence LGTM

PR #393, opened 2023-01-30
make max number of open files configurable

oliverkurth:
Improve the change in PR #391 to make it configurable, or disable it. Also, check if `/proc` can be used because in that case we will not hit the problem, instead of just checking if we are installing to another installroot. By default the limit is still set to 1024.

### DIFF
--- a/client/rpmtrans.c
+++ b/client/rpmtrans.c
@@ -599,6 +599,60 @@ error:
     goto cleanup;
 }
 
+/*
+ * Restrict number of open files. When rpm cannot access /proc
+ * it tries to set the close on exec flag for every possible
+ * fd, which may take a long time if the limit is very high.
+ * See also https://github.com/rpm-software-management/rpm/issues/2081.
+ * This can be disabled by setting "openmax=0" in the configuration.
+ */
+
+uint32_t
+TDNFSetOpenMax(PTDNF pTdnf)
+{
+    uint32_t dwError = 0;
+    char *pszProcPath = NULL;
+    int nIsDir = 0;
+
+    if(!pTdnf || !pTdnf->pConf || !pTdnf->pArgs)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    /* First, check if /proc is available - if rpm can
+       open it there is no issue. */
+    dwError = TDNFJoinPath(&pszProcPath,
+                           pTdnf->pArgs->pszInstallRoot ?
+                               pTdnf->pArgs->pszInstallRoot : "",
+                           "/proc/self/fd",
+                           NULL);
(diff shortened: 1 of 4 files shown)


client/rpmtrans.c
```diff
@@ -599,6 +599,58 @@ reportProblems(PTDNFRPMTS pTS)
...
+       open it there is no issue. */
+    dwError = TDNFJoinPath(&pszProcPath,
+                           pTdnf->pArgs->pszInstallRoot ?
+                               pTdnf->pArgs->pszInstallRoot : "",
+                           "/proc/self/fd",
+                           NULL);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFIsDir(pszProcPath, &nIsDir);
+    if (dwError == ERROR_TDNF_SYSTEM_BASE + ENOENT) {
```
sshedi, line 633: check `errno` instead, I think it will make it easy

client/rpmtrans.c
```diff
@@ -599,6 +599,58 @@ reportProblems(PTDNFRPMTS pTS)
...
+        nIsDir = 0;
```
sshedi, line 634: nIsDir already 0 by default

client/rpmtrans.c
```diff
@@ -599,6 +599,58 @@ reportProblems(PTDNFRPMTS pTS)
...
+        dwError = 0;
+    }
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (!nIsDir) {
+        int nOpenMax = pTdnf->pConf->nOpenMax;
+        struct rlimit rl = {nOpenMax, nOpenMax};
+        if (setrlimit(RLIMIT_NOFILE, &rl) != 0) {
+            /* shouldn't be fatal */
+            pr_err("warning: could not set rlimit: %s (%d)\n", strerror(errno), errno);
```
sshedi: can we print a warning if the rlimit value is huge and inform users that it will take a long time to finish the task?

sshedi (approved):
Changes are fine but I think would be better with some verbose regarding huge file limit values.

client/rpmtrans.c
oliverkurth: Checking `errno` is risky here in case there is some syscall  in `TDNFIsDir()` after checking the directory. For example a log message, clearing `errno`. This may cause obscure bugs.

client/rpmtrans.c
oliverkurth: Similar argument as above. The value of `nIsDir` is really undefined in case of an error, but the return code from `TDNFIsDir()` is defined. Again, this can cause obscure bugs.

client/rpmtrans.c
oliverkurth: Yes, we can do that.

oliverkurth: Done.

PR #394, opened 2023-02-03
Add `--source` and `--builddeps` options

oliverkurth:
This adds the two options `--source` and `--builddeps` to download source packages and build dependencies for a package.

Usage example:
```
# cat /etc/yum.repos.d/photon-srpms.repo                                                                                                                               
[photon-srpms]
name=VMware Photon Linux $releasever ($basearch)
baseurl=https://packages.vmware.com/photon/$releasever/photon_srpms_$releasever_$basearch
gpgkey=file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY-4096
gpgcheck=1
enabled=1
skip_if_unavailable=1
skip_md_filelists=1
```
```
tdnf install --source lsof
tdnf install --builddeps lsof
rpmbuild -bb /usr/src/photon/SPECS/lsof.spec
```
See built packages:
```
# ls -lR /usr/src/photon/RPMS/                                                                                                                                         
/usr/src/photon/RPMS/:
total 4
drwxr-xr-x 2 root root 4096 Feb  3 20:14 x86_64

/usr/src/photon/RPMS/x86_64:
total 144
-rw-r--r-- 1 root root 127910 Feb  3 20:14 lsof-4.95.0-1.x86_64.rpm
-rw-r--r-- 1 root root  15817 Feb  3 20:14 lsof-debuginfo-4.95.0-1.x86_64.rpm
```

Showing output of the commands:
```
# tdnf install --source lsof                                                                                                                                     
photon-srpms                            392205 100%
Installing:
lsof                                   src                          4.95.0-1.ph4                           photon-srpms                   1.28M                           1.29M

Total installed size:   1.28M
Total download size:   1.29M
Is this ok [y/N]: y
lsof                                   1351526 100%
Installing/Updating: lsof-4.95.0-1.ph4 (source)
Testing transaction
Running transaction
```
```
# tdnf install --builddeps lsof                                                                                                                                  

Installing:
krb5-devel                             x86_64                       1.17-9.ph4                             photon-updates               711.98k                         130.77k
libtirpc                               x86_64                       1.2.6-4.ph4                            photon-updates               197.58k                          91.54k
libtirpc-devel                         x86_64                       1.2.6-4.ph4                            photon-updates               654.85k                         200.36k

Total installed size:   1.53M
Total download size: 422.67k

Upgrading:
krb5                                   x86_64                       1.17-9.ph4                             photon-updates                 3.34M                           1.15M

Total installed size:   3.34M
Total download size:   1.15M
Is this ok [y/N]: y
krb5-devel                              133904 100%
libtirpc                                 93737 100%
libtirpc-devel                          205172 100%
krb5                                   1201085 100%
Testing transaction
Running transaction
Installing/Updating: krb5-1.17-9.ph4.x86_64
Installing/Updating: krb5-devel-1.17-9.ph4.x86_64
Installing/Updating: libtirpc-1.2.6-4.ph4.x86_64
Installing/Updating: libtirpc-devel-1.2.6-4.ph4.x86_64
Removing: krb5-1.17-7.ph4.x86_64
```

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -772,12 +772,30 @@ TDNFAddCmdLinePackages(
          * to our virtual @cmdline repo */
         pszPkgName = pCmdArgs->ppszCmds[nCmdIndex];
 
+        /* check if it's a package file or URL */
+        if (fnmatch("*.rpm", pszPkgName, 0) != 0) {
+            continue;
+        }
+
+        if (fnmatch("*.src.rpm", pszPkgName, 0) == 0) {
+            if (!pCmdArgs->nSource) {
+                pr_err("package '%s' appears to be a source rpm - use --source to install\n", pszPkgName);
+                dwError = ERROR_TDNF_INVALID_PARAMETER;
+            }
+        } else {
+            if (pCmdArgs->nSource) {
+                pr_err("package '%s' appears not to be a source rpm but --source was used\n", pszPkgName);
+                dwError = ERROR_TDNF_INVALID_PARAMETER;
+            }
+        }
+        BAIL_ON_TDNF_ERROR(dwError);
+
         dwError = TDNFIsFileOrSymlink(pszPkgName, &nIsFile);
         BAIL_ON_TDNF_ERROR(dwError);
 
-        /* if it's a file and matches *.rpm it's to be installed
+        /* if it's a file it's to be installed
          * directly as a file. No need to download. */
-        if (nIsFile && (fnmatch("*.rpm", pszPkgName, 0) == 0))
+        if (nIsFile)
         {
             pszRPMPath = realpath(pszPkgName, NULL);
             if (pszRPMPath == NULL)
--- a/client/api.c
+++ b/client/api.c
@@ -1611,11 +1611,18 @@ TDNFResolve(
                   &queueGoal);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    dwError = TDNFGoal(
-                  pTdnf,
-                  &queueGoal,
-                  &pSolvedPkgInfo,
-                  nAlterType);
+    if (!pTdnf->pArgs->nSource) {
+        dwError = TDNFGoal(
+                      pTdnf,
+                      &queueGoal,
+                      &pSolvedPkgInfo,
+                      nAlterType);
+    } else {
+        dwError = TDNFGoalNoDeps(
+                      pTdnf,
+                      &queueGoal,
+                      &pSolvedPkgInfo);
+    }
     BAIL_ON_TDNF_ERROR(dwError);
 
     dwError = TDNFCheckProtectedPkgs(pSolvedPkgInfo);
--- a/client/api.c
+++ b/client/api.c
@@ -778,13 +778,13 @@ TDNFAddCmdLinePackages(
         }
 
         if (fnmatch("*.src.rpm", pszPkgName, 0) == 0) {
-            if (!pCmdArgs->nSource) {
-                pr_err("package '%s' appears to be a source rpm - use --source to install\n", pszPkgName);
+            if (!pCmdArgs->nSource && !pCmdArgs->nBuildDeps) {
+                pr_err("package '%s' appears to be a source rpm - use --source to install, or --builddeps to install its build depenfdencies\n", pszPkgName);
                 dwError = ERROR_TDNF_INVALID_PARAMETER;
             }
         } else {
-            if (pCmdArgs->nSource) {
-                pr_err("package '%s' appears not to be a source rpm but --source was used\n", pszPkgName);
+            if (pCmdArgs->nSource || pCmdArgs->nBuildDeps) {
+                pr_err("package '%s' appears not to be a source rpm but --source or --builddeps was used\n", pszPkgName);
                 dwError = ERROR_TDNF_INVALID_PARAMETER;
             }
         }
@@ -1587,6 +1587,9 @@ TDNFResolve(
     char** ppszPkgsNotResolved = NULL;
     PTDNF_SOLVED_PKG_INFO pSolvedPkgInfo = NULL;
     uint64_t qwAvailCacheBytes = 0;
+    char **ppszPkgNames = NULL;
+    char **ppszPkgFiles = NULL; /* cmd line packages */
+    int i, iFiles = 0, iPkgs = 0;
 
     if(!pTdnf || !ppSolvedPkgInfo)
     {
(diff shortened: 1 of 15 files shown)


client/init.c
```diff
@@ -61,6 +61,8 @@ TDNFCloneCmdArgs(
     pCmdArgs->nJsonOutput    = pCmdArgsIn->nJsonOutput;
     pCmdArgs->nTestOnly      = pCmdArgsIn->nTestOnly;
     pCmdArgs->nSkipBroken    = pCmdArgsIn->nSkipBroken;
+    pCmdArgs->nSource        = pCmdArgsIn->nSource;
+    pCmdArgs->nBuildDeps     = pCmdArgsIn->nBuildDeps;
```
sshedi, line 65: Two things:
1. I think it's high time to use bit fields for command line args. Let's use uint62_t and let each bit represent an input arg. Have all the options in a different enum. Like:
```
typdef enum {
  ALL_DEPS = (1 << 0),
  ALLOW_ERASING = (1 << 1),
  ....
} input_args;
```

2. Instead of assigning each member of struct, just do `pCmdArgs = pCmdArgsIn`
And adjust the pointers accordingly. Or at least, move these options to a new structure.

```
typedef struct _TDNF_CMD_ARGS {
    cmd_opts opts;
  
    char* pszDownloadDir;  //directory for download, if nDownloadOnly is set
    char* pszInstallRoot;  //set install root
    char* pszConfFile;     //set conf file location
    char* pszReleaseVer;   //Release version

    //Commands and args that do not fall in options
    char** ppszCmds;
    int nCmdCount;
    PTDNF_CMD_OPT pSetOpt;
    char **ppszArgv;
} TDNF_CMD_ARGS;
```

something like this.

tools/cli/lib/parseargs.c
```diff
@@ -342,6 +343,8 @@ TDNFCopyOptions(
     pArgs->nJsonOutput    = pOptionArgs->nJsonOutput;
     pArgs->nTestOnly      = pOptionArgs->nTestOnly;
     pArgs->nSkipBroken    = pOptionArgs->nSkipBroken;
+    pArgs->nSource        = pOptionArgs->nSource;
+    pArgs->nBuildDeps     = pOptionArgs->nBuildDeps;
```
sshedi, line 347: same here, try with
```
pArgs = pOptionArgs
```

solv/tdnfpackage.c
```diff
@@ -1176,6 +1181,56 @@ SolvFindAvailablePkgByName(
...
+    if(!pSack || IsNullOrEmptyString(pszName) || !ppPkgList)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = SolvCreateQuery(pSack, &pQuery);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    pQuery->nScope = SCOPE_SOURCE;
```
sshedi, line 1204: there is one line difference between SolvFindAvailableSrcPkgByName & SolvFindAvailablePkgByName; fix it. Use a single function.

solv/tdnfpackage.c
```diff
@@ -980,6 +980,7 @@ uint32_t
 SolvCountPkgByName(
     PSolvSack pSack,
     const char* pszName,
+    int nSource,
```
sshedi, line 983: can we make this a part of some data structure? or else we will end up passing this function arg everywhere

solv/simplequery.c
```diff
@@ -132,3 +132,25 @@ SolvFindSolvablesByNevraStr(
 error:
     goto cleanup;
 }
+
+uint32_t
+SolvRequiresFromQueue(
+    Pool *pool,
+    Queue *pq_pkgs,  /* solvable ids */
```
sshedi, line 139: null check the args

solv/simplequery.c
```diff
@@ -132,3 +132,25 @@ SolvFindSolvablesByNevraStr(
...
+    Queue *pq_deps   /* string ids */
+)
+{
+    uint32_t dwError = 0;
+    int i,j;
```
sshedi, line 144: j's scope can be reduced

solv/simplequery.c
```diff
@@ -132,3 +132,25 @@ SolvFindSolvablesByNevraStr(
...
+
+    for (i = 0; i < pq_pkgs->count; i++) {
+        Queue q_tmp = {0};
+        Solvable *p_solv = pool_id2solvable(pool, pq_pkgs->elements[i]);
```
sshedi, line 148: validate the return values, generic comment

solv/simplequery.c
```diff
@@ -132,3 +132,25 @@ SolvFindSolvablesByNevraStr(
 error:
     goto cleanup;
 }
+
+uint32_t
+SolvRequiresFromQueue(
```
sshedi, line 137: code duplication:

https://github.com/vmware/tdnf/blob/dev/solv/tdnfpackage.c#L2064

please use a single function

pytests/repo/setup-repo.sh
```diff
@@ -123,6 +128,15 @@ gpgcheck=0
 enabled=1
 EOF
 
+cat << EOF > ${TEST_REPO_DIR}/yum.repos.d/photon-test-src.repo
```
sshedi, line 131: one question:
what happens if my binary rpm and source rpm are in same repo?

include/tdnftypes.h
```diff
@@ -227,6 +228,8 @@ typedef struct _TDNF_CMD_ARGS
     int nJsonOutput;       //output in json format
     int nTestOnly;         //run test transaction only
     int nSkipBroken;
+    int nSource;
```
sshedi, line 231: this structure should be refactored and cleaned up.

client/resolve.c
```diff
@@ -472,3 +457,88 @@ TDNFPrepareSinglePkg(
...
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    if (ppszPackageNameSpecs[0]) {
+        dwError = SolvCreateQuery(pTdnf->pSack, &pQuery);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        dwError = TDNFApplyScopeFilter(pQuery, SCOPE_SOURCE);
```
sshedi, line 486: take a look at these entries:
```
0 api.c TDNFListInternal  547 dwError = TDNFApplyScopeFilter(pQuery, nScope);
1 api.c TDNFRepoSync     1109 dwError = TDNFApplyScopeFilter(pQuery, SCOPE_ALL);
2 api.c TDNFRepoQuery    1431 dwError = TDNFApplyScopeFilter(pQuery, nScope);
3 api.c TDNFMark         2315 dwError = TDNFApplyScopeFilter(pQuery, SCOPE_INSTALLED);
```

duplicate code. I think this can be moved to a function of it's own

client/api.c
```diff
@@ -770,12 +770,30 @@ TDNFAddCmdLinePackages(
          * to our virtual @cmdline repo */
         pszPkgName = pCmdArgs->ppszCmds[nCmdIndex];
 
+        /* check if it's a package file or URL */
+        if (fnmatch("*.rpm", pszPkgName, 0) != 0) {
```
sshedi, line 776: fname.src.rpm matches with this regex
```
#include <stdio.h>
#include <fnmatch.h>

int main()
{
    char pattern[] = "*.rpm";
    char string[] = "file.src.rpm";
    int r;

    r = fnmatch(pattern,string,0);
    if( r==0 )
        printf("'%s' matches '%s'\n",
                pattern,
                string
              );
    else if( r==FNM_NOMATCH )
        printf("'%s' does not match '%s'\n",
                pattern,
                string
              );
    else
        fprintf(stderr,"fnmatch() error\n");

    return(0);
}
````

sshedi (changes requested):

client/init.c
oliverkurth: Yes, I am aware that the argument handling in `tdnf` needs to be refactored. I am thinking about a good design, that would also work with plugins supplying their own commands and arguments.

But that's not part of this PR.

pytests/repo/setup-repo.sh
oliverkurth: That should be totally fine.

client/api.c
oliverkurth: Yes, it does :-). It's supposed to match. If it matches, `fnmatch()` returns `0` and then we check if it also matches `*.src.rpm`.

client/resolve.c
oliverkurth: Yes, I have seen that. Another tech debt.

solv/simplequery.c
oliverkurth: I think it was my intention to do that, but it needs more changes. They are not quite the same - this one creates a `Queue` of solvable ids, `SolvGetDependenciesFromId` creates a list of names. `SolvGetDependenciesFromId` should call this one, or the code that uses it needs to be redesigned.

solv/simplequery.c
oliverkurth: Look at the code in `libsolv`. They never do that. `pool_id2solvable` is actually an inline function that won't return any error.

client/api.c
sshedi: Oops, got it now. I missed the `!=0` part

PR #401, opened 2023-02-16
BugFix segfault caused due to missing name param

### LINKED ISSUE #400: Missing Name= definition in repository definition causes segmentation fault
### Describe the bug

If a new repository does not contain a valid definition name= tdnf segfaults.

### Reproduction steps

1. echo "[foo]" > /etc/yum.repos.d/foo.repo
2. tdnf install mypacakge
Segmentation fault...

### Expected behavior

tdnf should not segfault

### Additional context

Looking at the code, it looks like the default behaviour when reading the repository files is to set a default value of NULL if the name definition is not specified. 

repolist.c
`
-bash-5.1# git blame -L487,499 ../client/repolist.c
 487)         dwError = TDNFReadKeyValue(
 488)                       pSections,
 489)                       TDNF_REPO_KEY_NAME,
 490)                       NULL,
 491)                       &pRepo->pszName);
 492)         BAIL_ON_TDNF_ERROR(dwError);
`

When later reading the name definition in, a strpcmp() is done on a null

AZaugg:
If a repository definition is missing the name parameter the repositroy is later processed and a segfault occurs.

Adding a check to ensure that the repository is not processed in the event that the name parameter is missing.

fixes #400

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -487,7 +487,7 @@ TDNFLoadReposFromFile(
         dwError = TDNFReadKeyValue(
                       pSections,
                       TDNF_REPO_KEY_NAME,
-                      NULL,
+                      pszRepo,
                       &pRepo->pszName);
         BAIL_ON_TDNF_ERROR(dwError);
 


client/init.c
```diff
@@ -251,8 +251,9 @@ TDNFRefreshSack(
         /*
          * skip the @cmdline repo - options do not apply, and it is
          * initialized.
+         * In addition skip any repository that has a null repo name set
          */
-        if (!strcmp(pRepo->pszName, CMDLINE_REPO_NAME) || !pRepo->nEnabled)
+        if ((!pRepo->pszName) || !strcmp(pRepo->pszName, CMDLINE_REPO_NAME) || !pRepo->nEnabled)
```
oliverkurth: I think we should not just ignore and skip it. I think we should check for it in line 238, and bail if NULL.

Maybe we should also add a check right when we read the configuration.

oliverkurth: Can you please address this?

oliverkurth (changes requested):

client/init.c
oliverkurth: @AZaugg , are going to address my comment? It's fine if not, I can take over.

AZaugg:
OK, I changed my approach, as the repo stanzas need to be unique, I set the name value to the stanza definition as the default value. I think this is the best of both worlds. Not ignoring it, but not failing if a repo file is missing the name param

oliverkurth:
> OK, I changed my approach, as the repo stanzas need to be unique, I set the name value to the stanza definition as the default value.

That makes a lot of sense.

oliverkurth (approved):

PR #403, opened 2023-02-17
Ensure tdnf lock file is removed on application exit

### LINKED ISSUE #402: tdnf lock file does not get removed
### Describe the bug

When running tdnf, the lock file that is written to is never cleaned up /var/run/.tdnf-instance-lockfile

### Reproduction steps

-bash-5.1# ls -l  /var/run/.tdnf-instance-lockfile
ls: cannot access '/var/run/.tdnf-instance-lockfile': No such file or directory
-bash-5.1#
-bash-5.1#  bin/tdnf --releasever 2.0 install nginx
1. package nginx-1:1.18.0-2.el7.ngx.x86_64 requires libcrypto.so.10()(64bit), but none of the providers can be installed
Found 1 problem(s) while resolving
Error(1301) : Solv general runtime error
-bash-5.1#
-bash-5.1# ls -l  /var/run/.tdnf-instance-lockfile
-rw-r--r-- 1 root root 0 Feb 17 01:46 /var/run/.tdnf-instance-lockfile
-bash-5.1#
-bash-5.1# ps -ef|grep dnf
root     3997785 3867423  0 06:48 pts/14   00:00:00 grep --color=auto dnf
-bash-5.1# ls -l  /var/run/.tdnf-instance-lockfile
-rw-r--r-- 1 root root 0 Feb 17 06:46 /var/run/.tdnf-instance-lockfile
-bash-5.1#

### Expected behavior

Lock file should be removed when application exists

### Additional context

_No response_

AZaugg:
On completion of a tdnf run, ensure that lock file is removed from /var/run.

Insert the PID of the running tdnf process into the lock file.

fixes #402

### DIFF
--- a/common/defines.h
+++ b/common/defines.h
@@ -33,3 +33,9 @@
     } while(0)
 
 #define TDNF_DEFAULT_MAX_STRING_LEN       16384000
+
+/*
+ * creating this under /var/run because /var/run/lock doesn't exist
+ * in fedora docker images and as a result ci fails
+ */
+#define TDNF_INSTANCE_LOCK_FILE     "/var/run/.tdnf-instance-lockfile"
--- a/common/lock.c
+++ b/common/lock.c
@@ -54,6 +54,21 @@ static tdnflock tdnflock_new(const char *lock_path, const char *descr)
 	else
 	{
 		lock->openmode = TDNFLOCK_WRITE | TDNFLOCK_READ;
+
+		/* Write out PID into lock file */
+		char pid_buffer[128] = {0};
+		int rsnpf;
+		rsnpf = snprintf(pid_buffer, sizeof(pid_buffer), "%ld\n", (long) getpid());
+
+		if (rsnpf > 0)
+		{
+			int wr;
+			wr = write(lock->fd, pid_buffer, strlen(pid_buffer));
+			if (wr == 0)
+			{
+				sync();
+			}
+		}
 	}
 
 	lock->fdrefs = 1;
@@ -234,6 +249,12 @@ tdnflock tdnflockFree(tdnflock lock)
 	{
 		tdnflock_release(lock);
 		tdnflock_free(lock);
+	}
+
+	if (remove(TDNF_INSTANCE_LOCK_FILE))
+	{
+			pr_err("WARNING: Unable to remove lockfile(%s)\n",
+							TDNF_INSTANCE_LOCK_FILE);
 	}
 
 	return NULL;
--- a/client/defines.h
+++ b/client/defines.h
@@ -22,12 +22,6 @@
 
 #include "config.h"
 
-/*
- * creating this under /var/run because /var/run/lock doesn't exist
- * in fedora docker images and as a result ci fails
- */
-#define TDNF_INSTANCE_LOCK_FILE     "/var/run/.tdnf-instance-lockfile"
-
 typedef enum
 {
     /* this should be a bitmask */


oliverkurth:
@sshedi can you look at this please?

common/defines.h
```diff
@@ -33,3 +33,9 @@
     } while(0)
 
 #define TDNF_DEFAULT_MAX_STRING_LEN       16384000
+
+/*
+ * creating this under /var/run because /var/run/lock doesn't exist
+ * in fedora docker images and as a result ci fails
+ */
+#define TDNF_INSTANCE_LOCK_FILE     "/var/run/.tdnf-instance-lockfile"
```
sshedi, line 41: Don't declare this macro here again. There should be only one declaration.

It's already there in `client/defines.h`

common/lock.c
```diff
@@ -54,6 +54,11 @@ static tdnflock tdnflock_new(const char *lock_path, const char *descr)
 	else
 	{
 		lock->openmode = TDNFLOCK_WRITE | TDNFLOCK_READ;
+
+		/* Write out PID into lock file */
+		char pid_buffer[128];
+		snprintf(pid_buffer, 128, "%ld\n", (long) getpid());
```
sshedi: Validate snprintf output

common/lock.c
```diff
@@ -54,6 +54,11 @@ static tdnflock tdnflock_new(const char *lock_path, const char *descr)
 	else
 	{
 		lock->openmode = TDNFLOCK_WRITE | TDNFLOCK_READ;
+
+		/* Write out PID into lock file */
+		char pid_buffer[128];
```
sshedi: initialize buf to 0

common/lock.c
```diff
@@ -54,6 +54,11 @@ static tdnflock tdnflock_new(const char *lock_path, const char *descr)
...
+		snprintf(pid_buffer, 128, "%ld\n", (long) getpid());
+		write(lock->fd, pid_buffer, strlen(pid_buffer));
```
sshedi: validate write output and call sync

common/lock.c
```diff
@@ -236,5 +241,11 @@ tdnflock tdnflockFree(tdnflock lock)
 		tdnflock_free(lock);
 	}
 
+	if (remove(TDNF_INSTANCE_LOCK_FILE))
+	{
+			pr_err("ERROR: Unable to remove lockfile(%s)\n",
```
sshedi: Make this a warning print? even though this file is not removed, there is no problem.

sshedi (changes requested):

common/defines.h
oliverkurth: I think code from `common/` does not include fles from`client/`. So the macro should be moved from `client/defines.h` to `common/defines.h`, if it's needed in `common/`

AZaugg: Good call, moved the defines to common.

oliverkurth:
There is a build error, see https://github.com/vmware/tdnf/actions/runs/4208360003/jobs/7304317763:
```
[ 21%] Building C object solv/CMakeFiles/tdnfsolv.dir/simplequery.c.o
/build/common/lock.c: In function 'tdnflock_new':
/build/common/lock.c:69:11: error: expected ';' before '}' token
   69 |     sync()
      |           ^
      |           ;
   70 |    }
      |    ~       
make[2]: *** [common/CMakeFiles/common.dir/build.make:160: common/CMakeFiles/common.dir/lock.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

Please test your changes before pushing.

oliverkurth:
Thanks for this!

PR #404, opened 2023-02-17
Add new 'dnf_check_update_compat' config file option. Defaults to disabled.

mminkus:
Add new 'dnf_check_update_compat' config file option. Defaults to disabled.
When enabled, return exit status 100 if there are package updates available.
Updated formatting of check-update to match that of list so that it can be
parsed by puppet correctly.

Background:
Puppet and other tooling expects yum/dnf/tdnf check-update to return exit code 100 if there are updates available. (If you merely want a list of the update you can call yum/dnf/tdnf list updates. There is a reason these are different commands I suppose.)

This PR adds this functionality to tdnf.
I was not sure of the best way to add this functionality given the current error handling and error code tables.

Here is an example call from puppet:

```
Debug: Prefetching tdnf resources for package
Debug: Executing: '/bin/rpm --version'
Debug: Executing '/bin/rpm -qa --nosignature --nodigest --qf '%{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n' | sort'
Debug: Executing: '/bin/tdnf check-update'
Debug: Puppet::Type::Package::ProviderTdnf: /bin/tdnf check-update exited with 0; no package updates available.
```

tdnf also prints the table columns in reverse for some reason (see: `tdnf check-update | less`) which also break's puppet's parsing of the output.
Modified check-update (`TDNFCliCheckUpdateCommand`) to use the same output as list (`TDNFCliListCommand`).

Testing:
Builds and runs for me and provides the correct status code.
Puppet is also now able to parse the output to determine which packages need updating and invokes tdnf to update the packages.

```
Debug: Prefetching tdnf resources for package
Debug: Executing: '/bin/rpm --version'
Debug: Executing '/bin/rpm -qa --nosignature --nodigest --qf '%{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n' | sort'
Debug: Executing: '/bin/tdnf check-update'
Debug: Package[LNKD-ucm-enc](provider=tdnf): Yum didn't find updates, current version (2.0.9-1.el7) is the latest
Debug: /Stage[main]/Coreucm/Package[LNKD-ucm-hieraretrieve]/ensure: LNKD-ucm-hieraretrieve "2.0.24-1.el7" is installed, latest is "0:2.0.25-1.el7"
Debug: Package[LNKD-ucm-hieraretrieve](provider=tdnf): Ensuring => latest
Debug: Executing: '/bin/rpm -q LNKD-ucm-hieraretrieve --nosignature --nodigest --qf %{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n'
Debug: Package[LNKD-ucm-hieraretrieve](provider=tdnf): Ensuring latest, so using upgrade
Debug: Executing: '/bin/tdnf -e 1 -y upgrade LNKD-ucm-hieraretrieve'
Notice: /Stage[main]/Coreucm/Package[LNKD-ucm-hieraretrieve]/ensure: ensure changed '2.0.24-1.el7' to '0:2.0.25-1.el7'
Debug: /Package[LNKD-ucm-hieraretrieve]: The container Class[Coreucm] will propagate my refresh event
Debug: Finishing transaction 48526180
```

### DIFF
--- a/tools/cli/lib/api.c
+++ b/tools/cli/lib/api.c
@@ -112,18 +112,17 @@ error:
     goto cleanup;
 }
 
+static
 uint32_t
-TDNFCliListCommand(
-    PTDNF_CLI_CONTEXT pContext,
-    PTDNF_CMD_ARGS pCmdArgs
-    )
+TDNFCliListPackagesPrint(
+    PTDNF_PKG_INFO pPkgInfo,
+    uint32_t dwCount,
+    int nJsonOutput
+)
 {
     uint32_t dwError = 0;
-    PTDNF_PKG_INFO pPkgInfo = NULL;
-    PTDNF_PKG_INFO pPkg = NULL;
-    uint32_t dwCount = 0;
     uint32_t dwIndex = 0;
-    PTDNF_LIST_ARGS pListArgs = NULL;
+    PTDNF_PKG_INFO pPkg = NULL;
     struct json_dump *jd = NULL;
     struct json_dump *jd_pkg = NULL;
 
@@ -136,23 +135,7 @@ TDNFCliListCommand(
     int nColPercents[LIST_COL_COUNT] = {55, 25, 15};
     int nColWidths[LIST_COL_COUNT] = {0};
 
-    if(!pContext || !pContext->hTdnf || !pContext->pFnList)
-    {
-        dwError = ERROR_TDNF_CLI_INVALID_ARGUMENT;
-        BAIL_ON_CLI_ERROR(dwError);
-    }
-
-    dwError = TDNFCliParseListArgs(pCmdArgs, &pListArgs);
-    BAIL_ON_CLI_ERROR(dwError);
-
-    dwError = pContext->pFnList(pContext, pListArgs, &pPkgInfo, &dwCount);
-    if (pCmdArgs->nJsonOutput && dwError == ERROR_TDNF_NO_MATCH)
-    {
-        dwError = 0;
-    }
-    BAIL_ON_CLI_ERROR(dwError);
-
-    if (pCmdArgs->nJsonOutput)
+    if (nJsonOutput)
     {
         jd = jd_create(0);
         CHECK_JD_NULL(jd);
@@ -223,6 +206,45 @@ TDNFCliListCommand(
     }
 
 cleanup:
+    return dwError;
+
+error:
+    JD_SAFE_DESTROY(jd);
+    JD_SAFE_DESTROY(jd_pkg);
+    goto cleanup;
+}
+
+uint32_t
+TDNFCliListCommand(
+    PTDNF_CLI_CONTEXT pContext,
+    PTDNF_CMD_ARGS pCmdArgs
+    )
+{
+    uint32_t dwError = 0;
+    PTDNF_PKG_INFO pPkgInfo = NULL;
+    uint32_t dwCount = 0;
+    PTDNF_LIST_ARGS pListArgs = NULL;
+
+    if(!pContext || !pContext->hTdnf || !pContext->pFnList)
+    {
+        dwError = ERROR_TDNF_CLI_INVALID_ARGUMENT;
+        BAIL_ON_CLI_ERROR(dwError);
+    }
+
+    dwError = TDNFCliParseListArgs(pCmdArgs, &pListArgs);
+    BAIL_ON_CLI_ERROR(dwError);
+
+    dwError = pContext->pFnList(pContext, pListArgs, &pPkgInfo, &dwCount);
+    if (pCmdArgs->nJsonOutput && dwError == ERROR_TDNF_NO_MATCH)
+    {
+        dwError = 0;
+    }
+    BAIL_ON_CLI_ERROR(dwError);
+
+    dwError = TDNFCliListPackagesPrint(pPkgInfo, dwCount, pCmdArgs->nJsonOutput);
+    BAIL_ON_CLI_ERROR(dwError);
+
+cleanup:
     if(pListArgs)
     {
         TDNFCliFreeListArgs(pListArgs);
@@ -234,8 +256,6 @@ cleanup:
     return dwError;
 
 error:
-    JD_SAFE_DESTROY(jd);
-    JD_SAFE_DESTROY(jd_pkg);
     goto cleanup;
 }
 
@@ -899,14 +919,15 @@ TDNFCliCheckUpdateCommand(
     uint32_t dwIndex = 0;
     char** ppszPackageArgs = NULL;
     int nPackageCount = 0;
-    struct json_dump *jd = NULL;
-    struct json_dump *jd_pkg = NULL;
+    int nCheckUpdateCompat = 0;
 
     if(!pContext || !pContext->hTdnf || !pCmdArgs || !pContext->pFnCheckUpdate)
     {
         dwError = ERROR_TDNF_CLI_INVALID_ARGUMENT;
         BAIL_ON_CLI_ERROR(dwError);
     }
+
+    nCheckUpdateCompat = GlobalGetDnfCheckUpdateCompat();
 
     dwError = TDNFCliParsePackageArgs(
                   pCmdArgs,
@@ -920,33 +941,7 @@ TDNFCliCheckUpdateCommand(
                                        &dwCount);
     BAIL_ON_CLI_ERROR(dwError);
 
-    if (pCmdArgs->nJsonOutput)
-    {
-        jd = jd_create(0);
-        CHECK_JD_NULL(jd);
-
-        CHECK_JD_RC(jd_list_start(jd));
-
-    	for(dwIndex = 0; dwIndex < dwCount; ++dwIndex)
-        {
-            jd_pkg = jd_create(0);
-            CHECK_JD_NULL(jd_pkg);
-
-            jd_map_start(jd_pkg);
-            pPkg = &pPkgInfo[dwIndex];
-
-            CHECK_JD_RC(jd_map_add_string(jd_pkg, "Name", pPkg->pszName));
-            CHECK_JD_RC(jd_map_add_string(jd_pkg, "Arch", pPkg->pszArch));
-            CHECK_JD_RC(jd_map_add_fmt(jd_pkg, "Evr", "%s-%s", pPkg->pszVersion, pPkg->pszRelease));
-            CHECK_JD_RC(jd_map_add_string(jd_pkg, "Repo", pPkg->pszRepoName));
-
-            CHECK_JD_RC(jd_list_add_child(jd, jd_pkg));
-            JD_SAFE_DESTROY(jd_pkg);
-        }
-        pr_json(jd->buf);
-        JD_SAFE_DESTROY(jd);
-    }
-    else
+    if (!nCheckUpdateCompat && !pCmdArgs->nJsonOutput)
     {
         for(dwIndex = 0; dwIndex < dwCount; ++dwIndex)
         {
@@ -957,6 +952,11 @@ TDNFCliCheckUpdateCommand(
             pr_crit("\n");
         }
     }
+    else
+    {
+        dwError = TDNFCliListPackagesPrint(pPkgInfo, dwCount, pCmdArgs->nJsonOutput);
+        BAIL_ON_CLI_ERROR(dwError);
+    }
 
 cleanup:
     TDNF_CLI_SAFE_FREE_STRINGARRAY(ppszPackageArgs);
@@ -964,13 +964,16 @@ cleanup:
     {
         TDNFFreePackageInfoArray(pPkgInfo, dwCount);
     }
+    /* yum and dnf return 100 if there are package updates available.
+       puppet depends on this behaviour, even with tdnf. */
+    if(nCheckUpdateCompat && dwCount > 0 && dwError == 0)
+        return ERROR_TDNF_CLI_CHECK_UPDATES_AVAILABLE;
     return dwError;
 
 error:
-    JD_SAFE_DESTROY(jd);
-    JD_SAFE_DESTROY(jd_pkg);
     goto cleanup;
 }
+
 
 uint32_t
 TDNFCliMakeCacheCommand(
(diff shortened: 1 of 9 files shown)


oliverkurth:
Thanks for that, we'll review.

Have you considered using the `-j` option? That will give you output in easy to parse `json` format.

mminkus:
> Have you considered using the `-j` option? That will give you output in easy tp parse `json` format.

I'm a huge fan of using `-j` json output where it's available and use it in our internal tooling. But in this instance, I am trying to fix a bug wrt tdnf and puppet. It would require even more drastic changes for puppet to parse json output in its package provider code (and would require tdnf to deviate from the dnf/yum code), it seemed easier to get tdnf to provide output in line with dnf and yum. Addtionally, tdnf's `check-update` output didn't even match it's `list updates` output so there was inconsistency within tdnf itself.

oliverkurth:
Can you please give examples of the output, before and after this change?

If the output changes, this may break scripts that assume the old format.

oliverkurth:
@sshedi can you please check? I am pretty this change would break `bin/tdnf-automatic.in`, both because of the return code and maybe because of different output. We would need to fix that if we merge this.

The CI test did not catch any errors though.

sshedi:
IMO, any non zero exit status from a bash command is an error condition and I don't know if we change the exit status of this option, how much of ripple effect it will create. So, I'm more inclined towards keeping it as is because I don't want our existing consumers to rework on their modules.

I think the fix should be in puppet, is it not possible to to parse the output of `check-update` and decide rather than relying on exit status? And exit status can be changed, there is no rule that says we should not. Relying on a definite exit status alone entirely to carry out an operation is a fragile implementation.

AZaugg:
The behaviour in yum/dnf [mentions](https://man7.org/linux/man-pages/man8/dnf.8.html) that exiting with exit code 100 is a standard error. If you are worried about it breaking behaviour of exiting deployments. Would a compromise be to add an option in tdnf.conf to enable exit code behaviors more similar to yum/dnf? ie

dnf_exitcode_behaviour = 1

And have it disabled by default.

mminkus:
> Can you please give examples of the output, before and after this change?
> 
> If the output changes, this may break scripts that assume the old format.

Yessir.

Previously:

```
[mminkus@ltx1-app5907 ~]$ sudo tdnf check-update
LNKD-coreutils.x86_64                         8.32-6.cm2 linkedin.stg.extra.clis
dbus.x86_64                                1.10.24-15.el7           linkedin-elr
openssl.x86_64                              1.0.2k-12.el7           linkedin-elr
openssl-libs.x86_64                         1.0.2k-12.el7           linkedin-elr
osmetrics.x86_64                            0.5.87-1.el7           linkedin-prod
[mminkus@ltx1-app5907 ~]$ echo $?
0
[mminkus@ltx1-app5907 ~]$
```

After:
```
[mminkus@ltx1-app5907 build]$ sudo bin/tdnf check-update
LNKD-coreutils.x86_64                                                                                                          8.32-6.cm2                                                           linkedin.stg.extra.clis
dbus.x86_64                                                                                                                    1.10.24-15.el7                                                                  linkedin-elr
openssl.x86_64                                                                                                                 1.0.2k-12.el7                                                                   linkedin-elr
openssl-libs.x86_64                                                                                                            1.0.2k-12.el7                                                                   linkedin-elr
osmetrics.x86_64                                                                                                               0.5.87-1.el7                                                                   linkedin-prod
[mminkus@ltx1-app5907 build]$ echo $?
100
[mminkus@ltx1-app5907 build]$
```

Note, this new output matches the `tdnf list updates` output of existing versions:
```
[mminkus@ltx1-app5907 ~]$ sudo tdnf list updates
LNKD-coreutils.x86_64                                                                                                         8.32-6.cm2                                                          linkedin.stg.extra.clis
dbus.x86_64                                                                                                                   1.10.24-15.el7                                                                 linkedin-elr
openssl.x86_64                                                                                                                1.0.2k-12.el7                                                                  linkedin-elr
openssl-libs.x86_64                                                                                                           1.0.2k-12.el7                                                                  linkedin-elr
osmetrics.x86_64                                                                                                              0.5.87-1.el7                                                                  linkedin-prod
[mminkus@ltx1-app5907 ~]$
```

mminkus:
> IMO, any non zero exit status from a bash command is an error condition and I don't know if we change the exit status of this option, how much of ripple effect it will create. So, I'm more inclined towards keeping it as is because I don't want our existing consumers to rework on their modules.

As Andy pointed out, yum (and dnf) returning 100 for `check-update` if there are updates is standard behavior and has been the case for at least 10+ years. All tooling in the wild expects this. If you intend for tdnf to be a drop in replacement, it should respect this existing behavior.

I found the yum website (http://yum.baseurl.org/,) but their git repo (git://yum.baseurl.org/yum.git) times out and I am unable to clone it to find exactly how far back this goes in their commit history. I do know dnf reimplemented the same logic.

Alternatively, providing a configuration option to enable this behavior could be a viable solution for us.

oliverkurth:
> Alternatively, providing a configuration option to enable this behavior could be a viable solution for us.

Agreed, this is a good idea. We might need this for both output format and exit code. Maybe something like `dnf_check_update_compat`.

> If you intend for tdnf to be a drop in replacement, it should respect this existing behavior.

True, but `tdnf` is already used in Photon, and a lot of scripts expect a different behavior. This code has been there for several years.

mminkus:
> > If you intend for tdnf to be a drop in replacement, it should respect this existing behavior.
> 
> True, but `tdnf` is already used in Photon, and a lot of scripts expect a different behavior. This code has been there for several years.

I see. We're adopting Mariner (Microsoft's Linux distro) which uses tdnf by default, but its breaking various things for us (and segfaulting a lot.) We're fixing all the bugs we're coming across. Hopefully we can get it to a useable state and don't have to revert to dnf.

I'll work on adding in a config flag and call it `dnf_check_update_compat` as you suggested.

mminkus:
So adding a new config flag:

```
diff --git a/client/config.c b/client/config.c
index 48058b5..bd30d44 100644
--- a/client/config.c
+++ b/client/config.c
@@ -151,6 +151,13 @@ TDNFReadConfig(
                   &pConf->nOpenMax);
     BAIL_ON_TDNF_ERROR(dwError);

+    dwError = TDNFReadKeyValueBoolean(
+                  pSection,
+                  TDNF_CONF_KEY_CHECK_UPDATE_COMPAT,
+                  0,
+                  &pConf->nCheckUpdateCompat);
+
     dwError = TDNFConfigReadProxySettings(
                   pSection,
                   pConf);
diff --git a/client/defines.h b/client/defines.h
index 8db8b34..fbc218a 100644
--- a/client/defines.h
+++ b/client/defines.h
@@ -95,6 +95,7 @@ typedef enum
 #define TDNF_CONF_KEY_EXCLUDE             "excludepkgs"
 #define TDNF_CONF_KEY_MINVERSIONS         "minversions"
 #define TDNF_CONF_KEY_OPENMAX             "openmax"
+#define TDNF_CONF_KEY_CHECK_UPDATE_COMPAT "dnf_check_update_compat"

 //Repo file key names
 #define TDNF_REPO_KEY_BASEURL             "baseurl"
diff --git a/include/tdnftypes.h b/include/tdnftypes.h
index eb185e1..883741a 100644
--- a/include/tdnftypes.h
+++ b/include/tdnftypes.h
@@ -248,6 +248,7 @@ typedef struct _TDNF_CONF
     int nCleanRequirementsOnRemove;
     int nKeepCache;
     int nOpenMax;          //set max number of open files
+    int nCheckUpdateCompat;
     char* pszRepoDir;
     char* pszCacheDir;
     char* pszPersistDir;
```

I'm hitting a problem where I can't access that option from `TDNFCliCheckUpdateCommand`.

`pContext->hTdnf->pConf->nCheckUpdateCompat` gives an error because hTdnf is of type:
```typedef void * HTDNF;```

I can't find an example where any other config flag is used in a CLI Command.
And I also noticed `nInstallOnlyLimit` isn't even used at all.

**Comment by oliverkurth** [comment]
> I'm hitting a problem where I can't access that option from TDNFCliCheckUpdateCommand.

I see. It's a limitation caused by the separation between the CLI and the library backend.

I think the solution is to use a global, analogue to `isQuiet` and `isJson` . I know, globals are frowned upon, but I see no other way until we refactor this (it's on our road map).

So my idea for now is this: in `OpenHandle()`, in `client/api.c`, similar to `GlobalSetJson()` set a global variable, like `isDnfCheckUpdateCompat` which will then be used in `tools/cli/lib/api.c`. Note that this needs to be done *after* reading the config file (after `TDNFReadConfig()` in `OpenHandle()`.

> And I also noticed nInstallOnlyLimit isn't even used at all.

Yes, this needs to be fixed.

**Comment by mminkus** [comment]
Okay, updated but I had to include another header to get access to the PTDNF struct.

Including structs.h alone caused other issues as other headers were also needed.

I don't know if there is a better way to access config file options from a cli command.

```
[mminkus@ltx1-app5907 build]$ sudo bin/tdnf check-update
LNKD-coreutils.x86_64                         8.32-6.cm2 linkedin.stg.extra.clis
dbus.x86_64                                1.10.24-15.el7           linkedin-elr
openssl.x86_64                              1.0.2k-12.el7           linkedin-elr
openssl-libs.x86_64                         1.0.2k-12.el7           linkedin-elr
osmetrics.x86_64                            0.5.87-1.el7           linkedin-prod
[mminkus@ltx1-app5907 build]$ echo $?
0
[mminkus@ltx1-app5907 build]$ sudo vi /etc/tdnf/tdnf.conf
[mminkus@ltx1-app5907 build]$ sudo bin/tdnf check-update
LNKD-coreutils.x86_64                                                                                                          8.32-6.cm2                                                           linkedin.stg.extra.clis
dbus.x86_64                                                                                                                    1.10.24-15.el7                                                                  linkedin-elr
openssl.x86_64                                                                                                                 1.0.2k-12.el7                                                                   linkedin-elr
openssl-libs.x86_64                                                                                                            1.0.2k-12.el7                                                                   linkedin-elr
osmetrics.x86_64                                                                                                               0.5.87-1.el7                                                                   linkedin-prod
[mminkus@ltx1-app5907 build]$ echo $?
100
[mminkus@ltx1-app5907 build]$
```

**Comment by mminkus** [comment]
> I think the solution is to use a global, analogue to isQuiet and isJson . I know, globals are frowned upon, but I see no other way until we refactor this (it's on our road map).

Okay, I can redo my PR as including the library backend headers into the CLI isn't very elegant.

**Comment by oliverkurth** [comment]
> Okay, I can redo my PR as including the library backend headers into the CLI isn't very elegant.

Thank you. I saw what you have done, but I think I'd prefer what I suggested.

**Comment by mminkus** [comment]
> > Okay, I can redo my PR as including the library backend headers into the CLI isn't very elegant.
> 
> Thank you. I saw what you have done, but I think I'd prefer what I suggested.

Agreed, seems to be the lesser of two evils.

**Comment by mminkus** [comment]
So the globals seem defined in log.c and are only used by the logging code. All the cli code uses `pCmdArgs` to check for the presence of command line args, not the globals. I could add a `GlobalGetDnfCheckUpdateCompat` function to get the global variable though.

File: tools/cli/lib/api.c [hunk 1]
```diff
@@ -964,6 +1023,10 @@ TDNFCliCheckUpdateCommand(
     {
         TDNFFreePackageInfoArray(pPkgInfo, dwCount);
     }
+    /* yum and dnf return 100 if there are package updates available.
+       puppet depends on this behaviour, even with tdnf. */
+    if(nCheckUpdateCompat && dwCount > 0)
```
oliverkurth: Please check for `dwError == 0`. We don't want to overwrite an eventual error code with 100.

tools/cli/lib/api.c
```diff
@@ -948,13 +960,60 @@ TDNFCliCheckUpdateCommand(
     }
     else
     {
-        for(dwIndex = 0; dwIndex < dwCount; ++dwIndex)
+        // Format output like `tdnf list updates`, which is compatible with yum/dnf
+        if (nCheckUpdateCompat)
```
oliverkurth: This code is copied from `TDNFCliListCommand()`, correct?. Can you please put this into a function to avoid code duplication?

tools/cli/main.c
```diff
@@ -205,6 +205,11 @@ TDNFCliPrintError(
         return dwError;
     }
 
+    if (dwErrorCode == ERROR_TDNF_CLI_CHECK_UPDATES_AVAILABLE)
+    {
+        return;
```
oliverkurth: Function should return a value:
`return dwError`

oliverkurth (changes requested):
This looks good, just a few changes.

Adding tests for this would be great, but not required.

tools/cli/lib/api.c
mminkus: There's actually a lot of duplicate code between those two functions, more than just that output. I'll take a look at what I can do.

tools/cli/main.c
mminkus: Done

tools/cli/lib/api.c
mminkus: Done

tools/cli/lib/api.c
mminkus: Added `TDNFCliListPackagesPrint` function (open to better naming suggestions) that is called by both `TDNFCliListCommand` and `TDNFCliCheckUpdateCommand` to do json and pretty table formatted output. Uses the old logic if nJsonOutput or nCheckUpdateCompat are not true.

From running the commands - with and without json - the output looks the same to me. And with nCheckUpdateCompat set to true, `list updates` and `check-update` also has the same output, with the additional exit code.

mminkus:
> Adding tests for this would be great, but not required.

Looking at unit tests now.

oliverkurth (approved):
This looks good, thank you very much for the contribution!

oliverkurth:
> Looking at unit tests now.

I missed this comment - but you can open a new PR.

PR #406, opened 2023-02-23
Add unit tests to test return code of tdnf check-update when

mminkus:
dnf_check_update_compat is set true.

Unit tests run and pass.

### DIFF
--- a/pytests/tests/test_check_update.py
+++ b/pytests/tests/test_check_update.py
@@ -6,16 +6,30 @@
 # of the License are located in the COPYING file of this distribution.
 #
 
+import os
 import pytest
+import shutil
+
+WORKDIR = '/root/test_config'
+TEST_CONF_FILE = 'tdnf.conf'
+TEST_CONF_PATH = os.path.join(WORKDIR, TEST_CONF_FILE)
 
 
 @pytest.fixture(scope='module', autouse=True)
 def setup_test(utils):
+    tdnf_conf = os.path.join(utils.config['repo_path'], 'tdnf.conf')
+
+    utils.makedirs(WORKDIR)
+    shutil.copy(tdnf_conf, TEST_CONF_PATH)
+
+    utils.edit_config({'dnf_check_update_compat': '1'}, section='main', filename=TEST_CONF_PATH)
+
     yield
     teardown_test(utils)
 
 
 def teardown_test(utils):
+    shutil.rmtree(WORKDIR)
     spkg = utils.config["sglversion_pkgname"]
     mpkg = utils.config["mulversion_pkgname"]
     utils.run(['tdnf', 'erase', '-y', spkg, mpkg])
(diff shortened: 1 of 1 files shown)


pytests/tests/test_check_update.py
```diff
@@ -6,16 +6,30 @@
...
 
 
 @pytest.fixture(scope='module', autouse=True)
 def setup_test(utils):
+    tdnf_conf = os.path.join(utils.config['repo_path'], 'tdnf.conf')
+
+    utils.makedirs(WORKDIR)
+    shutil.copy(tdnf_conf, TEST_CONF_PATH)
+
+    utils.edit_config({'dnf_check_update_compat': '1'}, section='main', filename=TEST_CONF_PATH)
```
oliverkurth, line 25: I see this is always enabled in all tests. We need to test with the setting enabled and with the setting disabled.

You can modify `test_check_update_multi_version_package` to check that it still returns 0 when updates are available, and do the setting `dnf_check_update_compat=1` only for `test_check_update_dnf_compat()`. You can also add a test that with the setting it still returns 0 when no updates are available.

pytests/tests/test_check_update.py
```diff
@@ -40,6 +54,15 @@ def test_check_update_multi_version_package(utils):
     assert len(ret['stdout']) > 0
```
oliverkurth, line 59: `assert` here for the expected return code.

pytests/tests/test_check_update.py
```diff
@@ -40,6 +54,15 @@ def test_check_update_multi_version_package(utils):
...
 
 
+def test_check_update_dnf_compat(utils):
+    package = utils.config["mulversion_pkgname"] + '-' + utils.config["mulversion_lower"]
+    ret = utils.run(['tdnf', 'install', '-y', '--nogpgcheck', package])
+    assert utils.check_package(utils.config["mulversion_pkgname"])
+
+    ret = utils.run(['tdnf', '--config', TEST_CONF_PATH, 'check-update'])
+    assert ret['retval'] == 100
```
oliverkurth, line 70: Add an `assert` like `assert len(ret['stdout']) > 0`.

oliverkurth (changes requested):

pytests/tests/test_check_update.py
mminkus: So I based this on test_config.py, and I believe even though it appears it's enabled for all tests, it is only used when we use `--config`, `TEST_CONF_PATH` otherwise it uses the config file without the `dnf_check_update_compat=1` that was added.

pytests/tests/test_check_update.py
mminkus: done

pytests/tests/test_check_update.py
mminkus: done

pytests/tests/test_check_update.py
oliverkurth: Oh, yes, you are correct. I missed that.

oliverkurth (approved):

oliverkurth:
Thank you for the contribution!

PR #415, opened 2023-03-27
refactor yes/no question

oliverkurth:
Make the yes/no question behave better. This will terminate only on valid answers, and otherwise repeat the question. Also adding "yes" and "no" (case insensitive) to valid answers. An empty response will be interpreted the same as "n" or "no". This is the same behavior as Fedora's dnf.

Examples:
Repeat question on invalid answers, accept "no":
```
# tdnf install lsof  

Installing:
lsof                                   x86_64                       4.95.0-1.ph4                           photon-updates               207.70k                         121.54k

Total installed size: 207.70k
Total download size: 121.54k
Is this ok [y/N]: yeah
Is this ok [y/N]: maybe
Is this ok [y/N]: nope
Is this ok [y/N]: no
```
Accepts "yes":
```
# tdnf install lsof                                                                                                                                              

Installing:
lsof                                   x86_64                       4.95.0-1.ph4                           photon-updates               207.70k                         121.54k

Total installed size: 207.70k
Total download size: 121.54k
Is this ok [y/N]: yes
lsof                                    124456 100%
Testing transaction
Running transaction
Installing/Updating: lsof-4.95.0-1.ph4.x86_64
```
Accepts upper case "Y":
```
# tdnf remove lsof                                                                                                                                               

Removing:
lsof                                   x86_64                       4.95.0-1.ph4                           @System                      207.70k                           0.00b

Total installed size: 207.70k
Total download size:   0.00b
Is this ok [y/N]: Y
Testing transaction
Running transaction
Removing: lsof-4.95.0-1.ph4.x86_64
root [ /build/bld-ph4 ]#
```
Empty response is "no" (default):
```# tdnf install lsof                                                                                                                                              

Installing:
lsof                                   x86_64                       4.95.0-1.ph4                           photon-updates               207.70k                         121.54k

Total installed size: 207.70k
Total download size: 121.54k
Is this ok [y/N]: 
```

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2022 VMware, Inc. All Rights Reserved.
+ * Copyright (C) 2015-2023 VMware, Inc. All Rights Reserved.
  *
  * Licensed under the GNU Lesser General Public License v2.1 (the "License");
  * you may not use this file except in compliance with the License. The terms
@@ -428,13 +428,24 @@ TDNFYesOrNo(
 
     if(!pArgs->nAssumeYes && !pArgs->nAssumeNo)
     {
-        pr_crit("%s", pszQuestion);
-        while ((opt = getchar()) == '\n' || opt == '\r');
-        opt = tolower(opt);
-        if (opt != 'y' && opt != 'n')
-        {
-            dwError = ERROR_TDNF_INVALID_INPUT;
-            BAIL_ON_TDNF_ERROR(dwError);
+        while(1) {
+            pr_crit("%s", pszQuestion);
+            char buf[256] = {0};
+            char *ret;
+
+            ret = fgets(buf, sizeof(buf)-1, stdin);
+            if (ret != buf || buf[0] == 0) {
+                /* should not happen */
+                dwError = ERROR_TDNF_INVALID_INPUT;
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+            buf[strlen(buf)-1] = 0;
+            if (strcasecmp(buf, "yes") == 0 || strcasecmp(buf, "y") == 0 ||
+                    strcasecmp(buf, "n") == 0 || strcasecmp(buf, "no") == 0 ||
+                    buf[0] == 0) {
+                opt = tolower(buf[0]);
+                break;
+            }
         }
     }
 
--- a/common/log.c
+++ b/common/log.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2022 VMware, Inc. All Rights Reserved.
+ * Copyright (C) 2019-2023 VMware, Inc. All Rights Reserved.
  *
  * Licensed under the GNU Lesser General Public License v2.1 (the "License");
  * you may not use this file except in compliance with the License. The terms
@@ -83,7 +83,7 @@ void log_console(int32_t loglevel, const char *format, ...)
     }
 
     vfprintf(stream, format, args);
-
+    fflush(stream);
 end:
     va_end(args);
 }


common/utils.c
```diff
@@ -428,13 +428,24 @@ TDNFYesOrNo(
...
-        pr_crit("%s", pszQuestion);
-        while ((opt = getchar()) == '\n' || opt == '\r');
-        opt = tolower(opt);
-        if (opt != 'y' && opt != 'n')
-        {
-            dwError = ERROR_TDNF_INVALID_INPUT;
-            BAIL_ON_TDNF_ERROR(dwError);
+        while(1) {
+            pr_crit("%s", pszQuestion);
+            char buf[256];
```
sshedi: buf[256] = {}; instruction in line 442 may not be the case always

common/utils.c
```diff
@@ -428,13 +428,24 @@ TDNFYesOrNo(
...
     {
-        pr_crit("%s", pszQuestion);
-        while ((opt = getchar()) == '\n' || opt == '\r');
-        opt = tolower(opt);
-        if (opt != 'y' && opt != 'n')
-        {
-            dwError = ERROR_TDNF_INVALID_INPUT;
-            BAIL_ON_TDNF_ERROR(dwError);
+        while(1) {
+            pr_crit("%s", pszQuestion);
```
sshedi, line 432: do we need to do `fflush(stdout)` here?

common/utils.c
```diff
@@ -428,13 +428,24 @@ TDNFYesOrNo(
...
+            char buf[256];
+            char *ret;
+
```
sshedi, line 435: do __fpurge(stdin) here

common/utils.c
```diff
@@ -428,13 +428,24 @@ TDNFYesOrNo(
...
+            ret = fgets(buf, sizeof(buf)-1, stdin);
+            if (ret != buf || strlen(buf) == 0) {
```
sshedi: check buf for non null before calling strlen or you can do:
```
(... || buf && !buf[0])
```

common/utils.c
```diff
@@ -428,13 +428,24 @@ TDNFYesOrNo(
...
+                /* should not happen */
+                dwError = ERROR_TDNF_INVALID_INPUT;
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+            buf[strlen(buf)-1] = 0;
+            if (strcasecmp(buf, "yes") == 0 || strcasecmp(buf, "y") == 0 ||
```
sshedi, line 443: nit:
```
char strs[][4] = {"yes", "y", "no", "n"};
for (i = 0; i < 4; ++i) {
  if (strcasecmp(buf, strs[i]) break;
}

if (i >= 4) {
  handle answers other than strs
}
```

common/utils.c
oliverkurth: Good point - it may be needed because we don't send a new line, so it isn't flushed automatically. But I think it should be in `log_console()`. But I never saw any issue.

common/utils.c
oliverkurth: Nice idea if we had more possible answers, but with just 4 we don't save any lines.

common/utils.c
oliverkurth: Not needed. When `buf == NULL` then `ret != buf`.

common/utils.c
oliverkurth: I am not sure if that is needed. I experimented with it, but it does not seem to have an effect. I tried:
```
echo y | ./bin/tdnf install lsof
```
(which should always work).

Typing `y` and <return> while downloading cache data (to give me time), but it still accepts it. Which may not actually be a bad thing anyway.

It's a non-standard function, so it may not be a good idea.

common/utils.c
oliverkurth: I can do that, but I fail to see where this has an effect, unless `fgets()` is totally misbehaving.

common/utils.c
sshedi: NP then but still I would suggest avoid strlen function call and checking for non empty char in buf[0], faster and simple way.

common/utils.c
sshedi: Yes, it's gcc specific. But I don't see a problem with using it.

see this once: https://stackoverflow.com/questions/20633284/strange-behaviour-of-c-codefgets-function

oliverkurth: I think the issue in that discussion is the `scanf()` not eating the line feed. We don't have that in tdnf.

PR #420, opened 2023-04-07
various changes, build and run on Ubuntu

oliverkurth:
A few miscellaneous changes and fixes, some of them Ubuntu related, which will make `tdnf` build and run on Ubuntu and Debian (with an installroot) for testing and setting up an OS.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -706,7 +706,6 @@ TDNFOpenHandle(
 
     dwError = TDNFLoadRepoData(
                   pTdnf,
-                  REPOLISTFILTER_ALL,
                   &pTdnf->pRepos);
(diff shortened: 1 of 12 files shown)


client/api.c
```diff
@@ -281,6 +281,12 @@ TDNFCheckLocalPackages(
     pr_info("Checking all packages from: %s\n", pszLocalPath);
 
     pCmdLinePool = pool_create();
+
+    if (pool_setdisttype(pCmdLinePool, DISTTYPE_RPM) < 0) {
```
oliverkurth: I should probably put this with `pool_create()` into a separate function because it repeats 4 times in the code.

oliverkurth: Done.

PR #423, opened 2023-05-04
add --qf option for Repoquery

shivania2:
add `--qf` option for Repoquery
`tdnf repoquery --qf {format}`
Custom display format. `{format}` is the string to output for each matched package.
Every occurrence of `%{tag}` within format is replaced by the corresponding attribute of the package.


tools/cli/lib/api.c
```diff
@@ -671,6 +671,322 @@ TDNFCliRepoSyncCommand(
...
+}
+
+uint32_t find_dependencies(PTDNF_CLI_CONTEXT pContext, int nquery_arg, char **query_argv, void *data, const int tdwCount, const int pdwCount, PTDNF_CMD_ARGS* ppCmdArgs)
+{
+    uint32_t dwError = 0;
+    PTDNF_CMD_ARGS pCmdArgs = NULL;
+
+    if(pdwCount == 0)
+    {
+       dwError = TDNFCliParseArgs(nquery_arg, query_argv, &pCmdArgs);
```
oliverkurth: We shouldn't create our own command line and reparse arguments, then re-run `TDNFRepoQurey()`. There needs to be another way.

tools/cli/lib/api.c
```diff
@@ -671,6 +671,322 @@ TDNFCliRepoSyncCommand(
...
+        pszVal = pPkgInfo->pszURL;
+    } else if (strcmp(tag, "requires") == 0) {
+        nquery_arg = 8;
+        char *query_argv[] = {"", "", "", "", "", "repoquery", "--requires", pPkgInfo->pszName};
+        dwError = find_dependencies(pContext, nquery_arg, query_argv, data, tdwCount, pdwCount, ppCmdArgs);
+        BAIL_ON_CLI_ERROR(dwError);
+        pszVal = pPkgInfo->ppszDependencies[0];
+    } else if (strcmp(tag, "provides") == 0) {
+        nquery_arg = 8;
+        char *query_argv[] = {"", "", "", "", "", "repoquery", "--provides", pPkgInfo->pszName};
```
oliverkurth: Why do we need these empty args?

Anyway, this is not how we should do this.

tools/cli/lib/api.c
```diff
@@ -671,6 +671,322 @@ TDNFCliRepoSyncCommand(
...
+              } else if (strcmp(ppszLines[j], ppszLines[j-1]))
+              {
+                   strcpy(&pszdependencies[p], "\n");
+                   p += nSepSize;
+                   strcpy(&pszdependencies[p], ppszLines[j]);
+                   p += strlen(ppszLines[j]);
+              }
+          }
+
+          dwError = TDNFAllocateString(pszdependencies, &ppPkgInfo->ppszDependencies[0]);
```
oliverkurth: So basically, you are concatenating all string elements into one. We shouldn't do that, it's unmaintainable.

tools/cli/lib/api.c
```diff
@@ -671,6 +671,322 @@ TDNFCliRepoSyncCommand(
...
+    } else if (strcmp(tag, "license") == 0) {
+        pszVal = pPkgInfo->pszLicense;
+    } else if (strcmp(tag, "url") == 0) {
+        pszVal = pPkgInfo->pszURL;
+    } else if (strcmp(tag, "requires") == 0) {
+        nquery_arg = 8;
+        char *query_argv[] = {"", "", "", "", "", "repoquery", "--requires", pPkgInfo->pszName};
+        dwError = find_dependencies(pContext, nquery_arg, query_argv, data, tdwCount, pdwCount, ppCmdArgs);
+        BAIL_ON_CLI_ERROR(dwError);
+        pszVal = pPkgInfo->ppszDependencies[0];
```
oliverkurth: Instead of concatenating the dependencies in `TDNFQueryFormatDepCommand()`, we can do that here.

oliverkurth (changes requested):
I see multiple issues here.

First is the build error:```
x86_64-unknown-linux-gnu-gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -O2 -g -O2 -g -fPIC -I/usr/include/python3.10 -c /build/python/tdnfbase.c -o build/temp.linux-x86_64-3.10/build/python/tdnfbase.o -I/build/include
In file included from /build/include/tdnf.h:16,
                 from /build/python/includes.h:13,
                 from /build/python/tdnfbase.c:9:
/build/include/tdnftypes.h:245:5: error: unknown type name 'bool'
  245 |     bool pszQueryFormat_dep;
      |     ^~~~
error: command '/usr/bin/x86_64-unknown-linux-gnu-gcc' failed with exit code 1
make[3]: *** [python/CMakeFiles/python.dir/build.make:77: python/CMakeFiles/python] Error 1```

After fixing this, the `repoquery` command segfaults, even without using the `--qf` flag. This needs better testing.

This also seems to break when querying dependencies multiple times:
```
# ./bin/tdnf repoquery --qf "%{requires} %{conflicts}" tdnf                                                                                                            
...
rpm-libs
tdnf-cli-libs
zlib (null)
Segmentation fault (core dumped)
```

I also see major problems in the design. We should not create function calls with passing command line arguments that are then parsed. Also concatenating the items in `ppzsDependencies` into one is not a good solution.

I see why the first is done - the code uses a client/server architecture and we don't have an API to get just dependencies. Moreover, the `TDNF_PKG_INFO` structure has only one field for dependencies (ppszDependencies). I also see that it may be inefficient to look up all possible dependencies (requires, conflicts,...) even if we don't need all or any of them.

I see two possible solutions:

1) create API calls that collect one kind of dependency, and call this when required via `tag_handler()` (directly or indirectly)
2) split `ppszDependencies` into multiple arrays, one for each dependency (listed in `REPOQUERY_DEP_KEY`), and make `depKey` in `TDNF_REPOQUERY_ARGS` a set of flags instead of an enum. We would need to scan the format string for what keys we need before calling `TDNFRepoQuery()` .

I think I would prefer the second solution because it aligns better with the current design. It will require changes for the current implementation for `repoquery`, but that is fine.

It's not all bad - the function `TDNFPopulatePkgInfoQueryFormat()` and the handling of the other single string fields is good, and I would probably have implemented that in a similar way.

tools/cli/lib/api.c
shivania2: Removed custom command line parser and implemented the approach two for queryformat implementation.

tools/cli/lib/api.c
shivania2: Removed the code for reparsing the arguments

tools/cli/lib/api.c
shivania2: Done

tools/cli/lib/api.c
```diff
@@ -671,6 +671,211 @@ TDNFCliRepoSyncCommand(
...
+    }
+
+    if (pszVal) {
+        *ppszVal = strdup(pszVal);
+    } else {
+        *ppszVal = strdup("(null)");
+    }
+
+cleanup:
+    TDNF_CLI_SAFE_FREE_MEMORY(*ppszLines);
```
oliverkurth: This segfaults. Should be `TDNF_CLI_SAFE_FREE_MEMORY(ppszLines);`.

oliverkurth (commented):
When using an empty list it prints `(null)`:
```
# ./bin/tdnf repoquery --qf "%{conflicts}" tdnf                                                                                                                       
(null)
(null)
root [ /build/bld-shiv ]# ./bin/tdnf repoquery --conflicts tdnf
```

tools/cli/lib/api.c
```diff
@@ -860,11 +1130,11 @@ TDNFCliRepoQueryCommand(
             {
                 pPkgInfo = &pPkgInfos[i];
 
-                if (pPkgInfo->ppszDependencies)
+                if (pPkgInfo->ppszDependencies[depKey])
```
oliverkurth: There is another segfault here:
```
(gdb) set args repoquery --list tdnf  
(gdb) run
Starting program: /build/bld-shiv/bin/tdnf repoquery --list tdnf  
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7e6c02d in TDNFCliRepoQueryCommand (pContext=0x7fffffffeb30, pCmdArgs=0x408010) at /build/tools/cli/lib/api.c:1133
1133	                if (pPkgInfo->ppszDependencies[depKey])
(gdb) print depKey
$1 = 10
```

client/api.c
```diff
@@ -1524,15 +1526,23 @@ TDNFRepoQuery(
     dwError = SolvGetQueryResult(pQuery, &pPkgList);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    /* handle query options */
+    if(pRepoqueryArgs->pszQueryFormat)
```
oliverkurth, line 1529: We need to address the case someone tries to do a mixed query, like: `tdnf repoquery --qf "%{provides}" --requires tdnf `. Currently, it just ignores any other options when `--qf` is given.

It maybe okay to throw an error, but look what `dnf `does.

include/tdnftypes.h
```diff
@@ -166,7 +166,7 @@ typedef struct _TDNF_PKG_INFO
     char* pszFormattedDownloadSize;
     char* pszRelease;
     char* pszLocation;
-    char **ppszDependencies;
+    char ***ppszDependencies;
```
oliverkurth: Should be renamed to `pppszDependencies` (three `p`s)

include/tdnftypes.h
```diff
@@ -406,8 +405,9 @@ typedef struct _TDNF_REPOQUERY_ARGS
 
     /* query options */
     int nChangeLogs;
-    REPOQUERY_DEP_KEY depKey; /* list dependencies of this type, 0 => unset */
+    int *depKey; /* list dependencies of this type, 0 => unset */
```
oliverkurth: Instead of an array of integers, use a bit field.

tools/cli/lib/api.c
```diff
@@ -671,6 +671,211 @@ TDNFCliRepoSyncCommand(
...
+                    if (q < *out + *len-1) {
+                        q++; *q = 0;
+                    }
+                }
+                *q = 0;
+                free(value);
+            }
+            p += tag_len + 3;
+
+        } else {
```
oliverkurth, line 863: We should handle C format like special characters like `\n` and so. Maybe you can copy code from `jsondump/jsondump.c `, in `jsonify_string()`.

tools/cli/lib/api.c
```diff
@@ -698,9 +907,53 @@ TDNFCliRepoQueryCommand(
...
+            while(*t && *t != '}') t++;
+
+            if (!*t)
+                return -1;
+            tag_len = t - &pszQueryFormat[2];
+            tag = (char *)malloc(tag_len+1);
+            strncpy(tag, &pszQueryFormat[2], tag_len);
+            tag[tag_len] = 0;
+
+            if (strcmp(tag, "provides") == 0) {
```
oliverkurth: We have more than these 5 keys ("provides", ...) . See `depKeys` in `tools/cli/lib/parserepoqueryargs.c`.

oliverkurth (changes requested):

tools/cli/lib/api.c
```diff
@@ -671,6 +671,211 @@ TDNFCliRepoSyncCommand(
...
+ 
+            dwError = tag_handler(tag, data, &value);
+            if (dwError == ERROR_TDNF_INVALID_PARAMETER) {
+                pr_err("Package doesn't have attribute %s\n", tag);
+            }
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            if ((*len <= strlen(value))  || ((*len >= strlen(*out)) && (strlen(value) > (*len - strlen(*out))))) {
+                nLength = strlen(*out) + strlen(value) + buffer_bytes;
+                dwError = TDNFReAllocateMemory(nLength, (void **)out);
```
oliverkurth: Nit: this works, but usually for this pattern the space for reallocation is doubled. That may safe a few repeated reallocations. If something needs n+1 bytes, it's most likely okay to allocate 2*n bytes.

tools/cli/lib/api.c
shivania2: Done

tools/cli/lib/api.c
shivania2: Resolved the seg fault.

client/api.c
shivania2: In fedora when we issue mixed query than the result of last query is printed. 
In photon if we issue mixed query than the result of first query is printed.
Should we add an error for mixed query?

include/tdnftypes.h
shivania2: Done

include/tdnftypes.h
shivania2: Changed to bitfield

tools/cli/lib/api.c
shivania2: Added code to handle special characters. But in output special characters are printed as it is.

tools/cli/lib/api.c
shivania2: Added other depkeys also

tools/cli/lib/api.c
shivania2: Done

client/api.c
oliverkurth: Yes, I think we should throw an error. Looks like `dnf` isn't properly handling it either.

tools/cli/lib/api.c
oliverkurth: Instead of having so many `else if`'s, can we loop through them?

tools/cli/lib/api.c
oliverkurth: Sorry for pointing at `jsondump/jsondump.c` -it's the other way around.

If the format string contains "\n", it should be converted to a line feed. Similar for the others. So you need to scan the format string for backwards slashes (`if (*p=='\\')`), then see if the following char is a literal 'n'. If so, convert to a LF, like `'\n'`.

`dnf` (Fedora) does it correctly:
```
# dnf repoquery --qf "start\n%{url}\nend" dnf
Last metadata expiration check: 0:05:58 ago on Mon May 15 18:39:33 2023.
start
https://github.com/rpm-software-management/dnf
end
```

You implementation does not:
```
# ./bin/tdnf repoquery --qf "start\n%{url}\nend" tdnf
start\nhttps://github.com/vmware/tdnd
```
And it also just drops the "end". Something else is not right there.

oliverkurth (changes requested):
This pints garbage at the end:
```
# ./bin/tdnf repoquery --qf "Provides\n%{provides}end" tdnf
Provides\nconfig(tdnf)
libtdnf.so.3()(64bit)
tdnf
tdnf(x86-64)
yud?hU
```

This does not work:
```
# ./bin/tdnf repoquery --qf "Requires\n%{requires}\nProvides%{provides}end" tdnf
Requires\n/usr/bin/env
curl-libs
libc.so.6(GLIBC_2.34)(64bit)
libcurl.so.4()(64bit)
libpopt.so.0()(64bit)
librpm.so.9()(64bit)
librpmio.so.9()(64bit)
libsolv
libsolv.so.1()(64bit)
libsolv.so.1(SOLV_1.0)(64bit)
libsolv.so.1(SOLV_1.1)(64bit)
libsolvext.so.1()(64bit)
libsolvext.so.1(SOLV_1.0)(64bit)
libsqlite3.so.0()(64bit)
libssl.so.3()(64bit)
libtdnfcli.so.3()(64bit)
openssl-libs
rpm-libs
tdnf-cli-libs
zlis
```

tools/cli/lib/api.c
shivania2: Done

tools/cli/lib/api.c
shivania2: Changed the code and Verified special characters are handled properly in the output result.

client/api.c
shivania2: Added code to throw error for mixed query

oliverkurth: Thank you. But can we have a more meaningful output?
```
# ./bin/tdnf repoquery --qf "Requires\n%{requires}\n" --requires tdnf
```

tools/cli/lib/api.c
```diff
@@ -671,6 +684,272 @@ TDNFCliRepoSyncCommand(
     goto cleanup;
 }
 
+uint32_t parse_depResult(char **ppszDependencies, char **ppszVal)
```
oliverkurth: You can use `TDNFJoinArrayToString()` from `common/strings.c` and make this function much simpler.

Also, use another name for it - it's not parsing anything. It also has nothing special about dependencies, it's just sorting an array of strings and joining them. Maybe `TDNFJoinArrayToStringSorted()`, and move it to `common/utils.c`.

tools/cli/lib/api.c
```diff
@@ -671,6 +684,272 @@ TDNFCliRepoSyncCommand(
...
+        pszVal = pPkgInfo->pszSourcePkg;
+    } else if (strcmp(tag, "description") == 0) {
+        pszVal = pPkgInfo->pszDescription;
+    } else if (strcmp(tag, "summary") == 0) {
+        pszVal = pPkgInfo->pszSummary;
+    } else if (strcmp(tag, "license") == 0) {
+        pszVal = pPkgInfo->pszLicense;
+    } else if (strcmp(tag, "url") == 0) {
+        pszVal = pPkgInfo->pszURL;
+    } else if (strcmp(tag, "provides") == 0) {
```
oliverkurth: Also here we can use a loop.

tools/cli/lib/api.c
oliverkurth: Thank you!

oliverkurth (changes requested):
This looks much better now.

Can you please also add tests in the `pytests/tests` directory?

include/tdnftypes.h
oliverkurth: You declared it a pointer to `int`, but it's nowhere used as a pointer. For flags, use `unsigned int`.

oliverkurth: Also, use another name that reflects that it's a set of flags. `depKeys` (plural) is fine, or `depKeySet`.

shivania2: Done

tools/cli/lib/api.c
shivania2: Done

tools/cli/lib/api.c
shivania2: Done

client/api.c
shivania2: Added a new error code: "ERROR_TDNF_CLI_INVALID_MIXED_QUERY_QUERYFORMAT"

oliverkurth:
The test is failing:
```
=================================== FAILURES ===================================
_______________________________ test_queryformat _______________________________

utils = <pytests.conftest.TestUtils object at 0x7f9357b81b10>

    def test_queryformat(utils):
        querytags = ['name', 'arch', 'version', 'reponame', 'release',
                     'evr', 'sourcename', 'size', 'downloadsize', 'installsize',
                     'sourcerpm', 'description', 'summary', 'license', 'url',
                     'conflicts', 'enhances', 'obsoletes', 'provides',
                     'recommends', 'requires', 'suggests', 'supplements']
    
        for tag in querytags:
            ret = utils.run(['tdnf',
                             'repoquery',
                             '--qf',
                             '%{{{}}}'.format(tag),
                             BASE_PKG
                             ])
            assert ret['retval'] == 0
    
        ret = utils.run(['tdnf',
                         'repoquery',
                         '--qf',
                         '%{name} : %{requires}',
                         BASE_PKG
                         ])
        assert ret['retval'] == 0
        assert 'tdnf-repoquery-base' in '\n'.join(ret['stdout'])
>       assert 'tdnf-test-cleanreq-required' in '\n'.join(ret['stdout'])
E       AssertionError: assert 'tdnf-test-cleanreq-required' in 'tdnf-repoquery-base :'
E        +  where 'tdnf-repoquery-base :' = <built-in method join of str object at 0x7f93599e3998>(['tdnf-repoquery-base :'])
E        +    where <built-in method join of str object at 0x7f93599e3998> = '\n'.join

tests/test_repoquery.py:270: AssertionError
```

pytests/tests/test_repoquery.py
```diff
@@ -241,3 +241,50 @@ def test_arch(utils):
...
+
+    ret = utils.run(['tdnf',
+                     'repoquery',
+                     '--qf',
+                     '%{obsoletes}\n%{arch}',
+                     BASE_PKG
+                     ])
+    assert ret['retval'] == 0
+    assert 'tdnf-test-dummy-obsoleted' in '\n'.join(ret['stdout'])
+    assert 'x86_64' in '\n'.join(ret['stdout'])
```
oliverkurth: This needs to work with `aarch64` too, use `ARCH`. An example is above in line 243.

pytests/tests/test_repoquery.py
```diff
@@ -241,3 +241,50 @@ def test_arch(utils):
     ret = utils.run(['tdnf', 'repoquery', '--arch', 'noarch'])
     assert 'noarch' in "\n".join(ret['stdout'])
     assert ARCH not in "\n".join(ret['stdout'])
+
+# Repoquery should list output for all the querys in the correct format
+def test_queryformat(utils):
```
oliverkurth, line 247: I like this, but it's not enough - we need to test that the formating is correct too for example.

Test for example that `Requires:\n%{requires}\nConficts:\n%{conflicts}\n` works as expected.

pytests/tests/test_repoquery.py
```diff
@@ -241,3 +241,50 @@ def test_arch(utils):
...
+                 'evr', 'sourcename', 'size', 'downloadsize', 'installsize',
+                 'sourcerpm', 'description', 'summary', 'license', 'url',
+                 'conflicts', 'enhances', 'obsoletes', 'provides',
+                 'recommends', 'requires', 'suggests', 'supplements']
+
+    for tag in querytags:
+        ret = utils.run(['tdnf',
+                         'repoquery',
+                         '--qf',
+                         '%{{{}}}'.format(tag),
```
oliverkurth: Just a nit: you can use `f"%{{{tag}}}"`.

A lot of places still use `format()`, but for new code we should use it like `f"foo: {foo}"`.

client/api.c
oliverkurth: Thank you!

oliverkurth (changes requested):

pytests/tests/test_repoquery.py
shivania2: Done

pytests/tests/test_repoquery.py
shivania2: Done

pytests/tests/test_repoquery.py
shivania2: ARCH represent the platform architecture. --qf option requires is not accepting ARCH argument. Added test case to identify noarch architecture

pytests/repo/tdnf-repoquery-queryformat.spec
```diff
@@ -0,0 +1,31 @@
...
+Vendor:     VMware, Inc.
+Distribution:   Photon
+License:    VMware
+Url:        http://www.vmware.com
+Group:      Applications/tdnftest
+Requires:   tdnf-test-cleanreq-required
+Conflicts:  tdnf-test-dummy-conflicts-0
+Provides:   tdnf-repoquery-queryformat
+Obsoletes:  tdnf-test-dummy-obsoleted
+
```
oliverkurth: This breaking other tests. Please find another way for your tests.

The `Obsoletes: tdnf-test-dummy-obsoleted` seems to break the tests `test_install_obsoleting` and `test_install_obsoleted_installed` in `test_install.py`.

The `Requires: tdnf-test-cleanreq-required` seems to break the `tdnf_py_distro_sync` test in `test_tdnf_python.py`:
```
tests/test_tdnf_python.py:231: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

quiet = True, refresh = False, cfg = None

    def tdnf_py_distro_sync(quiet=False, refresh=False, cfg=None):
        conf = cfg if cfg else config
>       tdnf.distro_sync(quiet=quiet, refresh=refresh, config=conf)
E       SystemError: <built-in function distro_sync> returned a result with an exception set

tests/test_tdnf_python.py:84: SystemError
------------------------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------------------------
Found 1 problems
tdnf-test-cleanreq-required is needed by tdnf-repoquery-queryformat-1.0.1-2.x86_64
```

There is also a failure in `test_tdnf_automatic.py` that I haven't figured out yet, but is probably somehow related too.

pytests/tests/test_repoquery.py
oliverkurth: Yes, but when we run this test on `aarch64` we will see `aarch64` and not `x86_64`.

Using `noarch` works.

oliverkurth (changes requested):
We are getting closer.

Please fix the tests. You can run the tests locally before pushing your changes.

Please also add a description for the PR.

pytests/repo/tdnf-repoquery-queryformat.spec
shivania2: Added changes to tdnf-repoquery-queryformat.spec and test_repoquery.py. Content of tdnf-repoquery-queryformat.spec is not impacting other tests

shivania2: Locally test_tdnf_python.py and test_install.py passed. test_tdnf_automatic.py is failing with and without my changes

oliverkurth:
The `pytest` tests are passing now, good. But there are still errors. Please fix. I am not going to merge any PR that shows errors, and I expect that submitters run the tests locally.
```
============ 328 passed, 1 skipped, 1 warning in 197.75s (0:03:17) =============
[100%] Built target check
../pytests/tests/test_repoquery.py:246:1: E302 expected 2 blank lines, found 1
../pytests/tests/test_repoquery.py:260:14: E211 whitespace before '('
flake8 tests failed
Error: Process completed with exit code 1.
```

oliverkurth (changes requested):
This looks good now! Just fix the `flake8` errors and I will merge this.

shivania2:
> [ ](/shivania2)

On my system there are couple of false error showing up locally(I am trying to fix it). Therefore I was not able to catch flake8 error. 
Thank You I will fix the flake8 error

oliverkurth (approved):
Thank you very much for the hard work!

PR #428, opened 2023-06-02
fix reinstall conflicts

oliverkurth:
This fixes file conflict errors on reinstall with a package that has binary differences, but same NEVRA. Change is for `stable-3.1`.

This may also be a better way to do reinstalls, and maybe it should be done the same way in the other active branches. `rpm` makes a difference between `reinstall` and `replacepkgs`. This change makes `tdnf` behave as if calling `rpm` with `--reinstall`. Before it was more like `--replacepkgs`.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -164,6 +164,8 @@ TDNFAlterCommand(
     PTDNF_SOLVED_PKG_INFO pSolvedInfo
     )
 {
+    UNUSED(nAlterType);
+
     uint32_t dwError = 0;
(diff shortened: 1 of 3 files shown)


sshedi:
Tested the changes, `reinstall` is working as expected.
Please add a test for the same, otherwise LGTM.

oliverkurth:
I will add a test later, it needs some work. Also, currently on github tests aren't run for the `stable-3.1`, I'd probably should change that too.

PR #437, opened 2023-07-11
Compare checksum for repodata against the downloaded rpm file

shivania2:
tdnf when downloading packages (or using from a repo accessible as a file), does not check for the checksum which is provided in the metadata. Recently there is a issue caught https://bugzilla.eng.vmware.com/show_bug.cgi?id=3237405(Fixed by checking the size), a proper check would have caught this issue earlier.

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -965,3 +965,27 @@ error:
     TDNF_SAFE_FREE_MEMORY(pszDirName);
     goto cleanup;
 }
+
+int32_t strtoi(const char *ptr)
+{
+    char *p = NULL;
+    long int tmp = 0;
+
+    tmp = strtol(ptr, &p, 10);
+
+    if (*p || tmp > INT_MAX || tmp < INT_MIN)
+    {
+        pr_crit("WARNING: invalid arg to %s: '%s'\n", __func__, ptr);
+        return 0;
+    }
+
+    return (int32_t) tmp;
+}
+
+int isTrue(const char *str)
+{
+    if (!strcasecmp(str, "false"))
+        return 0;
+
+    return !strcasecmp(str, "true") || strtoi(str);
+}
--- a/client/config.c
+++ b/client/config.c
@@ -26,12 +26,6 @@ TDNFConfGetRpmVerbosity(
     return nLogLevel;
 }
 
-static
-int isTrue(const char *str)
-{
-    return strcasecmp(str, "true") == 0 || atoi(str) != 0;
-}
-
 uint32_t
 TDNFReadConfig(
     PTDNF pTdnf,
@@ -108,7 +102,7 @@ TDNFReadConfig(
 
         if (strcmp(cn->name, TDNF_CONF_KEY_INSTALLONLY_LIMIT) == 0)
         {
-            pConf->nInstallOnlyLimit = atoi(cn->value);
+            pConf->nInstallOnlyLimit = strtoi(cn->value);
         }
         else if (strcmp(cn->name, TDNF_CONF_KEY_CLEAN_REQ_ON_REMOVE) == 0)
         {
@@ -153,7 +147,7 @@ TDNFReadConfig(
         }
         else if (strcmp(cn->name, TDNF_CONF_KEY_OPENMAX) == 0)
         {
-            pConf->nOpenMax = atoi(cn->value);
+            pConf->nOpenMax = strtoi(cn->value);
         }
         else if (strcmp(cn->name, TDNF_CONF_KEY_CHECK_UPDATE_COMPAT) == 0)
         {
--- a/client/plugins.c
+++ b/client/plugins.c
@@ -223,12 +223,6 @@ _TDNFFreePlugin(
     }
 }
 
-static
-int isTrue(const char *str)
-{
-    return strcasecmp(str, "true") == 0 || atoi(str) != 0;
-}
-
 /* read config file */
 static
 uint32_t
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -421,12 +421,6 @@ error:
     goto cleanup;
 }
 
-static
-int isTrue(const char *str)
-{
-    return strcasecmp(str, "true") == 0 || atoi(str) != 0;
-}
-
 uint32_t
 TDNFLoadReposFromFile(
     PTDNF pTdnf,
@@ -525,23 +519,23 @@ TDNFLoadReposFromFile(
             }
             else if (strcmp(cn->name, TDNF_REPO_KEY_PRIORITY) == 0)
             {
-                pRepo->nPriority = atoi(cn->value);
+                pRepo->nPriority = strtoi(cn->value);
             }
             else if (strcmp(cn->name, TDNF_REPO_KEY_TIMEOUT) == 0)
             {
-                pRepo->nTimeout = atoi(cn->value);
+                pRepo->nTimeout = strtoi(cn->value);
             }
             else if (strcmp(cn->name, TDNF_REPO_KEY_RETRIES) == 0)
             {
-                pRepo->nRetries = atoi(cn->value);
+                pRepo->nRetries = strtoi(cn->value);
             }
             else if (strcmp(cn->name, TDNF_REPO_KEY_MINRATE) == 0)
             {
-                pRepo->nMinrate = atoi(cn->value);
+                pRepo->nMinrate = strtoi(cn->value);
             }
             else if (strcmp(cn->name, TDNF_REPO_KEY_THROTTLE) == 0)
             {
-                pRepo->nThrottle = atoi(cn->value);
+                pRepo->nThrottle = strtoi(cn->value);
             }
             else if (strcmp(cn->name, TDNF_REPO_KEY_SSL_VERIFY) == 0)
             {
--- a/common/includes.h
+++ b/common/includes.h
@@ -21,6 +21,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 #include <libgen.h>
+#include <limits.h>
 
 #include <tdnftypes.h>
 #include <tdnferror.h>
--- a/common/prototypes.h
+++ b/common/prototypes.h
@@ -319,4 +319,8 @@ tdnflockNewAcquire(
     const char *descr
     );
 
+int32_t strtoi(const char *ptr);
+
+int isTrue(const char *str);
+
 #endif /* __COMMON_PROTOTYPES_H__ */
--- a/llconf/entry.h
+++ b/llconf/entry.h
@@ -19,6 +19,8 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#pragma once
+
 /**
  * A structure to hold the results of #cnf_find_entry.
  */
--- a/llconf/ini.h
+++ b/llconf/ini.h
@@ -19,6 +19,8 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#pragma once
+
 struct cnfnode *parse_ini(struct cnfmodule *cm, FILE *fptr);
 int unparse_ini(struct cnfmodule *cm, FILE *fptr, struct cnfnode *cn_root);
 void register_ini(struct cnfnode *opt_root);
--- a/llconf/lines.h
+++ b/llconf/lines.h
@@ -19,6 +19,8 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#pragma once
+
 /** \file
  * Handling for collections of lines of text.
  *
--- a/llconf/modules.h
+++ b/llconf/modules.h
@@ -19,6 +19,8 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#pragma once
+
 extern struct cnfmodule *cnfmodules;
 
 /** Information for registering a parser module. */
--- a/llconf/strutils.h
+++ b/llconf/strutils.h
@@ -19,6 +19,8 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#pragma once
+
 #define cp_buf_until(dest, p, expr) \
   { \
     char *q = buf; \
--- a/plugins/repogpgcheck/repogpgcheck.c
+++ b/plugins/repogpgcheck/repogpgcheck.c
@@ -312,12 +312,6 @@ error:
     goto cleanup;
 }
 
-static
-int isTrue(const char *str)
-{
-    return strcasecmp(str, "true") == 0 || atoi(str) != 0;
-}
-
 uint32_t
 TDNFRepoGPGCheckReadConfig(
     PTDNF_PLUGIN_HANDLE pHandle,
--- a/tools/cli/lib/parsehistoryargs.c
+++ b/tools/cli/lib/parsehistoryargs.c
@@ -72,10 +72,10 @@ TDNFCliParseHistoryArgs(
         dwError = TDNFSplitStringToArray(pArgs->ppszCmds[2], "-", &ppszRange);
         BAIL_ON_CLI_ERROR(dwError);
 
-        pHistoryArgs->nFrom = atoi(ppszRange[0]);
+        pHistoryArgs->nFrom = strtoi(ppszRange[0]);
         if (ppszRange[1])
         {
-            pHistoryArgs->nTo = atoi(ppszRange[1]);
+            pHistoryArgs->nTo = strtoi(ppszRange[1]);
         }
     }
 
@@ -93,11 +93,11 @@ TDNFCliParseHistoryArgs(
         }
         else if (strcasecmp(pSetOpt->pszOptName, "from") == 0)
         {
-            pHistoryArgs->nFrom = atoi(pSetOpt->pszOptValue);
+            pHistoryArgs->nFrom = strtoi(pSetOpt->pszOptValue);
         }
         else if (strcasecmp(pSetOpt->pszOptName, "to") == 0)
         {
-            pHistoryArgs->nTo = atoi(pSetOpt->pszOptValue);
+            pHistoryArgs->nTo = strtoi(pSetOpt->pszOptValue);
         }
     }
 


client/rpmtrans.c
```diff
@@ -898,6 +899,16 @@ TDNFTransAddInstallPkg(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    dwError = TDNFGetSHA256DigestForFile(pszFilePath, digest_from_file);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (memcmp(digest_from_file, pInfo->pszChecksum, SHA256_DIGEST_LENGTH))
```
oliverkurth: The reason this fails is that you are comparing a string in hexadecimal format (`pInfo->pszChecksum`) with a binary checksum (`digest_from_file`). You need to convert the hex string to binary and then compare. See how it's done in `plugins/metalink/utils.c`:
```
            TDNFCheckHexDigest(currHashInfo->value, hash_ops[currHashType].length)) {
            dwError = TDNFChecksumFromHexDigest(currHashInfo->value, digest);
            BAIL_ON_TDNF_ERROR(dwError);

            dwError = TDNFCheckHash(pszFile, digest, hash_Type);
```
Here the function `TDNFChecksumFromHexDigest` converts from hex string to binary.

client/utils.c
```diff
@@ -330,6 +330,104 @@ TDNFGetFileSize(
     goto cleanup;
 }
 
+uint32_t
```
oliverkurth: I see you have copied this from `plugins/metalink/utils.c` and slightly modified.

Instead please just move those functions (copy here and remove in `plugins/metalink/utils.c`) , and make the metalink plugin use those functions from here (you may have to move them to `common/utils.c` instead of `client/utils.c`, not sure).

That won't work with these modifications because you degeneralized this to just use sha256.

oliverkurth (changes requested):

solv/tdnfpackage.c
```diff
@@ -820,6 +820,58 @@ SolvGetPackageId(
...
+    }
+
+    pSolv = pool_id2solvable(pSack->pPool, dwPkgId);
+    if(!pSolv)
+    {
+        dwError = ERROR_TDNF_NO_DATA;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pszTemp = solvable_lookup_checksum(pSolv, SOLVABLE_CHECKSUM, &checksumtype);
```
oliverkurth: Looks like `libsolv` has a function to get the checksum in binary: https://github.com/openSUSE/libsolv/blob/2fc0cb31a2010d70f3f9197189730400edbd6ef4/src/solvable.h#L76

solv/tdnfpackage.c
```diff
@@ -820,6 +820,58 @@ SolvGetPackageId(
...
+    PSolvSack pSack,
+    uint32_t dwPkgId,
+    char** ppszChecksum)
+{
+
+    uint32_t dwError = 0;
+    const char* pszTemp = NULL;
+    char* pszChecksum = NULL;
+    Solvable *pSolv = NULL;
+    int checksumtype= REPOKEY_TYPE_SHA256;
```
oliverkurth: Can you find out if repo data can use other types than sha256?

oliverkurth (changes requested):
I haven't found out why, but somehow this breaks removing packages:
```
# ./bin/tdnf install lsof

Installing:
libtirpc                               x86_64                       1.3.3-2.ph5                            photon-updates               194.10k                          92.17k
lsof                                   x86_64                       4.96.4-1.ph5                           photon-updates               207.83k                         120.93k

Total installed size: 401.92k
Total download size: 213.10k
Is this ok [y/N]: y
libtirpc                                 94380 100%
lsof                                    123832 100%
Testing transaction
Running transaction
Installing/Updating: libtirpc-1.3.3-2.ph5.x86_64
Installing/Updating: lsof-4.96.4-1.ph5.x86_64

# ./bin/tdnf remove lsof
Nothing to do.
```

solv/tdnfpackage.c
oliverkurth: Look at https://github.com/rpm-software-management/createrepo_c , it looks like it supports different hash types:

checksum | Checksum type to use | md5, sha, sha1, sha224, sha256, sha384, sha512 | sha256

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+        dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+cleanup:
+    return dwError;
+error:
+    if (!IsNullOrEmptyString(filename))
+    {
+        pr_err("Error: Validating metalink (%s) FAILED (digest mismatch)\n", filename);
```
oliverkurth: You need to remove references to "metalink" since the code isn't specific to it any more.

client/packageutils.c
```diff
@@ -1148,6 +1148,12 @@ TDNFPopulatePkgInfos(
                       &pPkgInfo->pszLocation);
         BAIL_ON_TDNF_ERROR(dwError);
 
+        dwError = SolvGetPkgChecksumFromId(
```
oliverkurth: This shouldn't fail when there is no checksum available. This is why removing packages fails - it tries to get the checksum from an installed package, which isn't there.

We should check the error for `ERROR_TDNF_NO_DATA`, and in that case ignore the error. There may also be repositories without any checksum data (?), and we shouldn't fail with those (or maybe make it conditional with an option).

Later, when we want to check the checksum we can check for `pPkgInfo->pszChecksum == NULL`.

client/rpmtrans.c
```diff
@@ -898,6 +900,20 @@ TDNFTransAddInstallPkg(
...
 
+    if(pInfo->pszChecksum != NULL) {
+        hash = hash_ops + TDNF_HASH_SHA256;
+
+        dwError = TDNFGetDigestForFile(pszFilePath, hash, digest_from_file);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        if (memcmp(digest_from_file, pInfo->pszChecksum, hash->length))
+        {
+            pr_err("rpm file (%s) checksum (%s) does not match repodata checksum (%s)\n", pszFilePath, digest_from_file, pInfo->pszChecksum);
```
oliverkurth: You are printing binary data here in the error message. That's why we get failures in the github test run like this:
```
        stdout = out.decode().strip()
>       stderr = err.decode().strip()
E       UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcf in position 137: invalid continuation byte
```

But I don't understand why this error message appears in the first place - the checksum should match. And it works when I test this locally.

solv/tdnfpackage.c
```diff
@@ -820,6 +820,58 @@ SolvGetPackageId(
...
+
+    pszTemp = solvable_lookup_bin_checksum(pSolv, SOLVABLE_CHECKSUM, &checksumtype);
+
+    if(!pszTemp)
+    {
+        dwError = ERROR_TDNF_NO_DATA;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateString((char *)pszTemp, (char **)&pszChecksum);
```
oliverkurth: Ah, this is wrong: `solvable_lookup_bin_checksum()` is returning binary data, not a string. You need to figure out the length of the data and use `TDNDAllocateMemory()`, then copy the data.

solv/tdnfpackage.c
```diff
@@ -820,6 +820,61 @@ SolvGetPackageId(
...
+
+    pszTemp = solvable_lookup_bin_checksum(pSolv, SOLVABLE_CHECKSUM, &checksumtype);
+
+    if(!pszTemp)
+    {
+        dwError = ERROR_TDNF_NO_DATA;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    pszTemp_size = strlen((const char*)pszTemp) + 1;
```
oliverkurth: Again, this is binary data, not a `nul` terminated string. You cannot use `strlen()` for it.

oliverkurth (changes requested):

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+    if (IsNullOrEmptyString(filename) || !hash || !digest)
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    fd = open(filename, O_RDONLY);
+    if (fd < 0)
+    {
+        pr_err("Metalink: validating (%s) FAILED\n", filename);
```
ntsbtz: Is it only related to metalink if not this error message looks confusing.

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+    }
+
+    dwError = EVP_DigestInit_ex(ctx, digest_type, NULL);
+    if (!dwError)
+    {
+        pr_err("Digest Init Failed\n");
+        dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
+        /* MD5 is not approved in FIPS mode. So, overrriding
+           the dwError to show the right error to the user */
+#if defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
```
ntsbtz: Looks quite open can put a macro function and call it.

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+#else
+        if (FIPS_mode() && !strcasecmp(hash->hash_type, "md5"))
+#endif
+        {
+            dwError = ERROR_TDNF_FIPS_MODE_FORBIDDEN;
+        }
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    while ((length = read(fd, buf, (sizeof(buf)-1))) > 0)
```
ntsbtz: instead of sizeof call you can do BUFSIZ - 1 as we know the size of buf.

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+            pr_err("Digest Update Failed\n");
+            dwError = ERROR_TDNF_CHECKSUM_VALIDATION_FAILED;
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+        memset(buf, 0, BUFSIZ);
+    }
+
+    if (length == -1)
+    {
+        pr_err("Metalink: validating (%s) FAILED\n", filename);
```
ntsbtz: See the above comment is it only related to metalink or not.

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+    const char *hex_digest,
+    int digest_length
+    )
+{
+    int i = 0;
+    if(IsNullOrEmptyString(hex_digest) ||
+       (digest_length <= 0))
+    {
+        return 0;
+    }
```
ntsbtz: \n after this for better view.

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+
+    buf[0] = hex_digest[0];
+    buf[1] = hex_digest[1];
+
+    errno = 0;
+    val = strtoul(buf, NULL, 16);
+    if(errno)
+    {
+        pr_err("Error: strtoul call failed\n");
```
ntsbtz: This is correct way to log error msg. Wherever we used pr_err i was expecting like that Error: <error msg> that will help while debuging failures.

common/utils.c
```diff
@@ -965,3 +984,257 @@ TDNFDirName(
...
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+    *uintValue = (unsigned char)(val&0xff);
```
ntsbtz: Readablity purpose 
*uintValue = (unsigned char)(val&0xff) --> *uintValue = (unsigned char)(val & 0xff);

solv/tdnfpackage.c
```diff
@@ -820,6 +820,60 @@ SolvGetPackageId(
     goto cleanup;
 }
 
+uint32_t
+SolvGetPkgChecksumFromId(
+    PSolvSack pSack,
+    uint32_t dwPkgId,
+    int checksumtype,
```
ntsbtz: Maintain the variable naming generic through out the code
int checksumtype --> int checksumType
int checksumlen --> int checksumLen

solv/tdnfpackage.c
```diff
@@ -820,6 +820,60 @@ SolvGetPackageId(
...
+    {
+        dwError = ERROR_TDNF_NO_DATA;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(checksumlen, sizeof(unsigned char), (void **)&pszChecksum);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    memcpy(pszChecksum, pszTemp, checksumlen);
+    *ppszChecksum = pszChecksum;
```
ntsbtz: Outside this function there is a memory leak not freeing the ppszChecksum memory .
once you return from this function while cleanup TDNFFreePackageInfo is getting called need to add free for ppszChecksum.

client/packageutils.c
```diff
@@ -1148,6 +1148,19 @@ TDNFPopulatePkgInfos(
...
+                      pSack,
+                      dwPkgId,
+                      REPOKEY_TYPE_SHA256,
+                      SIZEOF_SHA256,
```
ntsbtz: This function only support for SHA256 ?

ntsbtz (changes requested):
There is some review comments please address them before merging.

client/packageutils.c
ntsbtz: > This function only support for SHA256 ?

If this is not specific for SHA256  then we can consider to check other checksum as well.
But if the expectation is only for SHA256 then its fine.

ntsbtz (approved):
Thanks for taking care of review comments.
LGTM !!! apart from SHA256 check is little confusing for me.

solv/tdnfpackage.c
```diff
@@ -820,6 +820,60 @@ SolvGetPackageId(
...
+
+    pszTemp = solvable_lookup_bin_checksum(pSolv, SOLVABLE_CHECKSUM, &checksumType);
+
+    if(!pszTemp)
+    {
+        dwError = ERROR_TDNF_NO_DATA;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(checksumLen, sizeof(unsigned char), (void **)&pszChecksum);
```
oliverkurth: Okay good, this is better now.

Please change the variable name `pszChecksum` to something like `pChecksum` or `pbChecksum`. The `psz` prefix means "pointer to a zero terminated string", which this is not. You can google for "hungarian notation".

Also for `ppszChecksum`.

solv/tdnfpackage.c
oliverkurth: Agree with Nitesh, you need to clean in `TDNFFreePackageInfoContents()` in `common/utils.c`.

client/packageutils.c
oliverkurth: I commented before that repodata can have other checksum types.

oliverkurth (changes requested):
Please use valgrind to check for memory leaks.

Also, run this with coverity once.

client/packageutils.c
```diff
@@ -1094,6 +1094,7 @@ TDNFPopulatePkgInfos(
     Id dwPkgId = 0;
     PTDNF_PKG_INFO pPkgInfos = NULL;
     PTDNF_PKG_INFO pPkgInfo = NULL;
+    int ChecksumType = 0;
```
oliverkurth: use `nChecksumType`.

include/tdnftypes.h
```diff
@@ -153,6 +153,7 @@ typedef struct _TDNF_PKG_INFO
     uint32_t dwEpoch;
     uint32_t dwInstallSizeBytes;
     uint32_t dwDownloadSizeBytes;
+    int ichecksumType;
```
oliverkurth: use `nChecksumType`

oliverkurth (changes requested):
This is looking really good now!

Just a few nits, and please write a test.

I tested this also:
* I created a repo with two packages and changed the checksum in the metadata for one of them, it failed to install as expected with the 1528 error message
* I tested the repo with sha512 (`-s sha512` option to `createrepo`), and that worked too
* I also tested with `valgrind`, it found no memory leaks.

Please write a test. You can manipulate the checksum in the metadata data this way:
* create a repo with a few packages with `createrepo`
* the metadata with the checksums are in the directory `repodata` in files ending with  `.primary.xml.gz`
* you need to unzip the file with `gunzip`, then change the checksum for a package
* compress the file again with the same name
* try to install, it should fail.

If the repo is for example in `/repo`, you can test an install with:
`tdnf --repofrompath=test,/repo --repoid=test install lsof`

See for example `pytests/tests/test_repofrompath.py`

shivania2:
> This is looking really good now!
> 
> Just a few nits, and please write a test.
> 
> I tested this also:
> 
> * I created a repo with two packages and changed the checksum in the metadata for one of them, it failed to install as expected with the 1528 error message
> * I tested the repo with sha512 (`-s sha512` option to `createrepo`), and that worked too
> * I also tested with `valgrind`, it found no memory leaks.
> 
> Please write a test. You can manipulate the checksum in the metadata data this way:
> 
> * create a repo with a few packages with `createrepo`
> * the metadata with the checksums are in the directory `repodata` in files ending with  `.primary.xml.gz`
> * you need to unzip the file with `gunzip`, then change the checksum for a package
> * compress the file again with the same name
> * try to install, it should fail.
> 
> If the repo is for example in `/repo`, you can test an install with: `tdnf --repofrompath=test,/repo --repoid=test install lsof`
> 
> See for example `pytests/tests/test_repofrompath.py`

Able to create sha512 repo, modified the checksum and repacked the files ending with  `.primary.xml.gz`. But package install is failed with command "tdnf --repofrompath=test,/repo --repoid=test install lsof"

pytests/tests/test_install.py
```diff
@@ -22,7 +26,46 @@ def setup_test(utils):
 def teardown_test(utils):
     pkgname = utils.config["mulversion_pkgname"]
     utils.run(['tdnf', 'erase', '-y', pkgname])
-
+    if os.path.isdir(WORKDIR):
+        shutil.rmtree(WORKDIR)
+
+def create_repo(utils):
```
oliverkurth: Please move all the checksum tests into a new `test_checksum.py` (or similar) file.

pytests/tests/test_install.py
```diff
@@ -22,7 +26,46 @@ def setup_test(utils):
...
+    ret = utils.run(['tdnf', '--repo={}'.format(reponame),
+                     '--download-metadata',
+                     'reposync'],
+                    cwd=workdir)
+    assert ret['retval'] == 0
+    synced_dir = os.path.join(workdir, reponame)
+    assert os.path.isdir(synced_dir)
+    assert os.path.isdir(os.path.join(synced_dir, 'repodata'))
+    assert os.path.isfile(os.path.join(synced_dir, 'repodata', 'repomd.xml'))
+
```
oliverkurth: Nit: needs two empty lines between functions (`flake8` will call this out)

pytests/tests/test_install.py
```diff
@@ -22,7 +26,46 @@ def setup_test(utils):
...
+            # Find the package element with the specified package_name
+            for package in root.findall(".//ns:package", ns):
+                name = package.find("ns:name", ns).text
+                if name == package_name:
+                    # Update the checksum value
+                    checksum_element = package.find("ns:checksum", ns)
+                    checksum_element.text = new_checksum
+                    # Compress the updated XML data again with the same name
+                    with gzip.open(file_path, 'wt', encoding='utf-8') as f:
+                        f.write(ET.tostring(root, encoding='utf-8').decode())
```
oliverkurth: Also here and other places.

oliverkurth (changes requested):
Thanks for adding tests - now you need to find out why they are failing ;-)

pytests/tests/test_checksum.py
```diff
@@ -0,0 +1,132 @@
...
+            suffix = find_suffix(filename, orig_pkg)
+
+            if suffix is not None:
+                orig_filename = orig_pkg + suffix
+                orig_path = os.path.join(workdir, filename)
+
+        elif filename.startswith(new_pkg):
+            suffix = find_suffix(filename, new_pkg)
+
+            if not os.path.exists(orig_path):
```
oliverkurth: This won't work if you find `new_pkg` first - `orig_path` will not be set (it will still be '')

pytests/tests/test_checksum.py
```diff
@@ -0,0 +1,132 @@
...
+                orig_filename = orig_pkg + '-1.0.1-2.x86_64.rpm'
+                orig_path = os.path.join(workdir, orig_filename)
+                file = open(orig_path, 'w')
+                file.close()
```
oliverkurth: What is the purpose of opening then closing the file?

pytests/tests/test_checksum.py
```diff
@@ -0,0 +1,132 @@
...
+                    shutil.copy2(new_path, orig_path)
+
+
+# install package with SHA512
+def test_install_package_with_sha512_checksum(utils):
+    workdir = WORKDIR
+    reponame = 'photon-test-sha512'
+    synced_dir = os.path.join(workdir, reponame)
+
+    enable_and_create_repo(utils)
```
oliverkurth: Since you are not modifying the repo for this test, you can just use `photon-test-sha512` directly.

pytests/tests/test_checksum.py
```diff
@@ -0,0 +1,132 @@
...
+    pattern = rf'^{escaped_prefix}(.+)$'
+
+    match = re.search(pattern, filename)
+    if match:
+        return match.group(1)
+    else:
+        return None
+
+
+def modify_rpm(workdir, orig_pkg, new_pkg):
```
oliverkurth: I am not sure what this function is supposed to do.

If it's to copy `orig_pkg` to `new_pkg`, this can be done much simpler.

oliverkurth (changes requested):

pytests/tests/test_checksum.py
```diff
@@ -0,0 +1,117 @@
...
+    orig_filename = ''
+    orig_path = ''
+
+    for filename in os.listdir(workdir):
+        if filename.startswith(orig_pkg):
```
oliverkurth: If we find `new_pkg` before `orig_pkg`, this won't work because it wont't set `orig_path`, right?

This function is very confusing. Please rewrite it with a simpler version.

oliverkurth (changes requested):
The functional changes look good to me, but you need to make changes in the test code. I don't think it works as intended.

PR #440, opened 2023-07-19
Misc fixes

sshedi:
When quiet option is passed with tdnf commands, either assumeyes or assumeno must be passed, if not don't suppress the info messages from tdnf.

```
root@phdev:~ # tdnf install vsftpd -q
Is this ok [y/N]:
```

Above is a sample output of the issue I'm talking about. Users really don't know what we are asking y/n for here and what is happening behind the screens.

So, if assumeyes or no is not passed, just output all the info.

```
root@phdev[±|verbose-fix ?:6 ✗]:~/tdnf.git/build # ./bin/tdnf install vsftpd -q

Installing:
vsftpd                  x86_64            3.0.5-4.ph5             local             182.91k          93.16k

Total installed size: 182.91k
Total download size:  93.16k
Is this ok [y/N]:
```

Sample output with the proposed fix. I know this is the same in Fedora at the moment but we don't have to be in alignment with dnf for everything.

cc: @tapakund @dweepadvani

### DIFF
--- a/tools/cli/main.c
+++ b/tools/cli/main.c
@@ -113,6 +113,7 @@ int main(int argc, char **argv)
             if (strcmp(pszCmd, arCmdMap[i].pszCmdName) == 0)
             {
                 pCmd = &arCmdMap[i];
+                break;
             }
         }
 
--- a/tools/cli/main.c
+++ b/tools/cli/main.c
@@ -6,19 +6,6 @@
  * of the License are located in the COPYING file of this distribution.
  */
 
-/*
- * Module   : main.c
- *
- * Abstract :
- *
- *            tdnf
- *
- *            command line tool
- *
- * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
- *
- */
-
 #include "includes.h"
 
 static TDNF_CLI_CMD_MAP arCmdMap[] =
--- a/client/api.c
+++ b/client/api.c
@@ -6,18 +6,6 @@
  * of the License are located in the COPYING file of this distribution.
  */
 
-/*
- * Module   : api.c
- *
- * Abstract :
- *
- *            tdnfclientlib
- *
- *            client library
- *
- * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
- */
-
 #include "includes.h"
 
 uid_t gEuid;
@@ -439,7 +427,7 @@ TDNFClean(
             dwError = 0;
         }
         BAIL_ON_TDNF_ERROR(dwError);
-        
+
         pr_info("\n");
     }
 cleanup:
--- a/client/client.c
+++ b/client/client.c
@@ -6,18 +6,6 @@
  * of the License are located in the COPYING file of this distribution.
  */
 
-/*
- * Module   : client.c
- *
- * Abstract :
- *
- *            tdnfclientlib
- *
- *            client library
- *
- * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
- */
-
 #include "includes.h"
 
 uint32_t
--- a/client/config.c
+++ b/client/config.c
@@ -6,18 +6,6 @@
  * of the License are located in the COPYING file of this distribution.
  */
 
-/*
- * Module   : config.c
- *
- * Abstract :
- *
- *            tdnfclientlib
- *
- *            client library
- *
- * Authors  : Priyesh Padmavilasom (ppadmavilasom@vmware.com)
- */
-
 #include "includes.h"
 
 #include "../llconf/nodes.h"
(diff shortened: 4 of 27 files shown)


tools/cli/main.c
```diff
@@ -113,15 +100,33 @@ int main(int argc, char **argv)
...
             {
                 pCmd = &arCmdMap[i];
+                break;
             }
         }
 
         if (pCmd)
         {
-            if (pCmd->ReqRoot && geteuid())
+            if (pCmd->ReqRoot)
```
sshedi: The actual change is here and rest is just cleanup.

oliverkurth:
I don't agree with this. If the user wants info they should not give the `-q` option. We shouldn't second guess what the user wants, this just leads to bugs that are hard to debug. Or: "If you don't want to shoot yourself into the foot, don't point your gun at it and pull the trigger". It's that simple, just less dramatic.

Maybe it's part of a script, and they display their own message, and just want the "Is this ok [y/N]:" displayed. (Maybe there is bear between your gun and your foot).

That Fedora behaves the same way is another good reason:
```
[root@6686562c19ca ~]# dnf -q install lsof
Is this ok [y/N]: N
Operation aborted.
[root@6686562c19ca ~]# 
```

I like the other changes though.

sshedi:
> We shouldn't second guess what the user wants, this just leads to bugs that are hard to debug.

Agree with the statement but not in this context. This fix is completely harmless.

Don't we refresh cache from many places on users behalf? If you apply the same analogy for it, we should not be doing refreshing the cache as well. The whole intent of this PR is something similar.
```
root@phdev:~ # tdnf install vsftpd -q
Is this ok [y/N]:
```
How end user is supposed to understand what the question here is and answer yes or no.

> Maybe it's part of a script, and they display their own message, and just want the "Is this ok [y/N]:" displayed. (Maybe there is bear between your gun and your foot).

There are different ways to suppress texts for this case and if someone is using tdnf from a script and imagine seeing a prompt like this in the middle of something and being clue less on what's happening and what is this prompt for.
Either we need this fix of we should not allow -q option for install, remove options when there is no assumeyes or assumeno and just throw an error and abort. That way we will not allow users to shoot themselves in the foot. Prevention is better than cure.

oliverkurth:
I want to be able to shoot into the direction of my foot if there is a bear in between. It saves lives.

I could see a point (one you are not making) to suppress other output. So maybe we can have an different option to suppress the output from downloading or other messages, and just show what it is going to do. Create a PR with that change if that is what you want.

> Don't we refresh cache from many places on users behalf?

That's not related to this at all.

And we have `--cacheonly` option to suppress that. And with that option we don't refresh the cache anyway for any reason - just like we don't print output when `--quiet` is given as option. This argument works _against_ your case, not for it.

> There are different ways to suppress texts for this case and if someone is using tdnf from a script and imagine seeing a prompt like this in the middle of something and being clue less on what's happening and what is this prompt for.

It's a bug in the script then, if it doesn't account for that.

sshedi:
> That's not related to this at all.
How? Let's say if I do this:
```
tdnf clean all
tdnf install <pkg>
```
Tdnf refreshes the cache without user's consent right?

oliverkurth:
> tdnf clean all
tdnf install <pkg>
Tdnf refreshes the cache without user's consent right?

That's absurd. No, it doesn't.

The cache might be outdated before it expired, so I probably *want* it to be refreshed. I have done that countless times, and recommended it to others.

If I do `tdnf --cacheonly` and _then_ it refreshes the cache, for any reason, it would be without consent.

dweepadvani:
@oliverkurth, @sshedi

What I get from discussions here that they are based upon 2 assumptions -

1) Assumption of @sshedi pertaining to user wanting information about actions being taken by tdnf with respect to changes being  made by tdnf to the system

2) Assumption of @oliverkurth about user wanting no information whatsoever as the user had provided -q argument to tdnf

For (2), as @sshedi said above and as per earlier discussions too, invoking TDNFYesOrNo() (showing interactive prompt to user) does not make sense when user provided -q command line option. I understand that dnf has the very same behaviour which is being addressed by this code change (as demonstrated above) and taking dnf here as the source of truth may not be right because that behaviour is lacking on user experience and expectations.

I too believe that (1) above is the step in right direction for providing better user experience with respect to tdnf usage. As @sshedi stated above, having prerequisite of providing --assumeyes or --assumeno for -q option makes more sense.

The debugging aspects in the context of output suppression by tdnf needs a separate discussion.

oliverkurth:
@sshedi can you take out the y/n change? - I think we discussed that to the end. The rest looks good to me.

oliverkurth:
Test failure is fixed in 05b8e4e758b2a80df0a3d821a8df32444cbe4ee8 .

I don't know if there is a way for me to rebase this PR and re-test. But this should pass when merged.

PR #444, opened 2023-08-03
Quiet option fix

sshedi:


pytests/tests/test_quiet.py
```diff
@@ -0,0 +1,83 @@
...
+    cmd = f"tdnf erase -qy {pkg}".split()
+    utils.run(cmd)
+
+
+def ensure_empty_ret(ret):
+    assert ret["retval"] == 0
+    assert not "\n".join(ret["stdout"])
+    stderr_l = []
+    for ele in ret["stderr"]:
+        if "NOKEY" not in ele:
```
sshedi, line 29: @oliverkurth this is for ignoring gpg key warning, somehow --nogpgcheck didn't work for me.

tools/cli/main.c
```diff
@@ -106,10 +106,29 @@ int main(int argc, char **argv)
...
+                    BAIL_ON_CLI_ERROR(dwError);
+                }
+
+                /*
+                 * For operations other than makecache & clean
+                 * suppressing info and prompting y/n doesn't make sense
+                 * Enable info if assumeyes or assumeno is not provided
+                 * along with quiet.
+                 */
+                if ((strcmp(pszCmd, "makecache") &&
```
oliverkurth: Why exceptions only for `clean` and `makecache`? There are a lot more commands that don't require input. It may be better to have a list of commands that may expect input, that is only commands that install, remove or update/downgrade packages.

tools/cli/main.c
```diff
@@ -106,10 +106,29 @@ int main(int argc, char **argv)
...
+                     strcmp(pszCmd, "clean")) &&
+                     pCmdArgs->nQuiet &&
+                     !pCmdArgs->nAssumeYes && !pCmdArgs->nAssumeNo)
+                {
+                    pr_err("--assumeyes|-y or --assumeno is must when quiet is set\n");
```
oliverkurth: Better english: "--assumeyes|-y or --assumeno is required when quiet is set\n"

oliverkurth (changes requested):

tools/cli/main.c
```diff
@@ -106,10 +106,29 @@ int main(int argc, char **argv)
...
+                    dwError = ERROR_TDNF_CLI_NOT_ENOUGH_ARGS;
```
oliverkurth: I think it's better to add a new error code. This would print "The command line parser could not continue. Expected at least one argument.", which is confusing.

In that case, you also don't need the `pr_err()`, because you can set the error code to "--assumeyes|-y or --assumeno is required when quiet is set".

pytests/tests/test_quiet.py
```diff
@@ -0,0 +1,83 @@
...
+            stderr_l.append(ele)
+
+    assert not stderr_l
+
+
+err_str = "--assumeyes|-y or --assumeno is must when quiet is set"
+
+def ensure_non_empty_ret(ret):
+    assert ret["retval"] == 909
```
oliverkurth: This fails, I guess you meant `904`?

But also see my other comment - add a new code.

oliverkurth (changes requested):

pytests/tests/test_quiet.py
oliverkurth: Can you show the output?

sshedi:
```
['warning: /root/tdnf.git/build/pytests/repo/cache/tdnf/photon-test/rpms/RPMS/x86_64/tdnf-test-one-1.0.1-2.x86_64.rpm: Header V4 RSA/SHA256 Signature, key ID 188e672b: NOKEY']
```

I added nogpgcheck option, but still tdnf throws this warning. Am I missing something?

tools/cli/main.c
sshedi: Good input. Ack.

tools/cli/main.c
sshedi: Agree

tools/cli/main.c
sshedi: My idea here is:

If a command needs to be run as root and if quiet is set for options other than makecache & clean, then abort.
So, I think this is enough and simple. Can you give me an example to the problem you see here?

sshedi:
We can also fix this by doing something like:
```
--- a/common/utils.c
+++ b/common/utils.c
@@ -443,6 +443,9 @@ TDNFYesOrNo(

     if(!pArgs->nAssumeYes && !pArgs->nAssumeNo)
     {
+        if (pArgs->nQuiet) {
+            error out;
+        }
         while(1) {
```

But this will be a costly operation. Let's fail early and quick. Imagine downloading giga byte size package on a slow network and seeing this error post download.

tools/cli/main.c
oliverkurth: There are a lot of commands that don't require input, like `check`, `check-local`, etc. So if you do `tdnf check -q` your change will throw an error.

I haven't counted, but I think there are more commands that do *not* require input than there are that do.

oliverkurth:
> But this will be a costly operation. Let's fail early and quick. Imagine downloading giga byte size package on a slow network and seeing this error post download.

The question is *before* downloading packages though. Which makes sense because you want to say be able to say "no" before downloading gigabytes on your modem link.

sshedi:
> > But this will be a costly operation. Let's fail early and quick. Imagine downloading giga byte size package on a slow network and seeing this error post download.
> 
> The question is _before_ downloading packages though. Which makes sense because you want to say be able to say "no" before downloading gigabytes on your modem link.

Ahh, I missed it. Shall we switch to this solution? Simple and easy to maintain.

tools/cli/main.c
sshedi: No; `check` won't throw error because it doesn't need root permissions. Only exception I found is for `tdnf history` rollback, undo, redo. But I think we should switch to simpler solution like I explained below.

oliverkurth:
> > > But this will be a costly operation. Let's fail early and quick. Imagine downloading giga byte size package on a slow network and seeing this error post download.
> > 
> > 
> > The question is _before_ downloading packages though. Which makes sense because you want to say be able to say "no" before downloading gigabytes on your modem link.
> 
> Ahh, I missed it. Shall we switch to this solution? Simple and easy to maintain.

Yes, agreed.

oliverkurth:
I think this looks good. But as discussed, this will be merged some time in the future for Photon > 5.0 (I created a label for that purpose).

oliverkurth (approved):

sshedi:
> I think this looks good. But as discussed, this will be merged some time in the future for Photon > 5.0 (I created a label for that purpose).

No problem. Thanks.

PR #446, opened 2023-08-04
Remove atoi usages from tdnf

sshedi:
- Replace atoi with strotol
- Move isTrue function to a common place
- Remove duplicate pr_err macro definitions
- And adjustments to make the build normal after the fixes

### DIFF
--- a/client/config.c
+++ b/client/config.c
@@ -29,7 +29,9 @@ TDNFConfGetRpmVerbosity(
 static
 int isTrue(const char *str)
 {
-    return strcasecmp(str, "true") == 0 || atoi(str) != 0;
+    int val = 0;
+
+    return strcasecmp(str, "true") == 0 || strtoi(str) != 0;
 }
 
 uint32_t
@@ -108,7 +110,7 @@ TDNFReadConfig(
 
         if (strcmp(cn->name, TDNF_CONF_KEY_INSTALLONLY_LIMIT) == 0)
         {
-            pConf->nInstallOnlyLimit = atoi(cn->value);
+            pConf->nInstallOnlyLimit = strtoi(cn->value);
         }
         else if (strcmp(cn->name, TDNF_CONF_KEY_CLEAN_REQ_ON_REMOVE) == 0)
         {
@@ -153,7 +155,7 @@ TDNFReadConfig(
         }
         else if (strcmp(cn->name, TDNF_CONF_KEY_OPENMAX) == 0)
         {
-            pConf->nOpenMax = atoi(cn->value);
+            pConf->nOpenMax = strtoi(cn->value);
         }
         else if (strcmp(cn->name, TDNF_CONF_KEY_CHECK_UPDATE_COMPAT) == 0)
         {
--- a/client/config.c
+++ b/client/config.c
@@ -26,14 +26,6 @@ TDNFConfGetRpmVerbosity(
     return nLogLevel;
 }
 
-static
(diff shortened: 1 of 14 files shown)


sshedi:
@oliverkurth need your help on this one. I'm not able to reproduce this in my build env. Strange. PTAL.

client/config.c
```diff
@@ -29,7 +29,9 @@ TDNFConfGetRpmVerbosity(
 static
 int isTrue(const char *str)
 {
-    return strcasecmp(str, "true") == 0 || atoi(str) != 0;
+    int val = 0;
```
oliverkurth: What's `val` for? It's unused, what am I missing?

common/utils.c
```diff
@@ -965,3 +965,19 @@ TDNFDirName(
     TDNF_SAFE_FREE_MEMORY(pszDirName);
     goto cleanup;
 }
+
+int32_t strtoi(const char *ptr)
```
oliverkurth, line 969: Is there a particular reason for this? If it's a sanity check, you'd probably should check if it's a number too - it could be something non-numeric.

llconf/entry.h
```diff
@@ -19,6 +19,16 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+
+#include <tdnf.h>
+#include <tdnf-common-defines.h>
+
+#include "../common/config.h"
+#include "../common/structs.h"
+#include "../common/prototypes.h"
+
```
oliverkurth: Did you see any error? I'd rather not have any outside dependencies in the `llconf` directory, it should be possible to use it in another project.

history/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ target_link_libraries(${TDNF_HISTORY_UTIL_BIN}
     ${RPM_LIBRARIES}
     ${SQLITE3_LIBRARIES}
     ${LIB_TDNF_HISTORY}
+    ${LIB_TDNF_COMMON}
```
oliverkurth: I'd rather have this as a standalone part.

sshedi:
> @oliverkurth need your help on this one. I'm not able to reproduce this in my build env. Strange. PTAL.

I got the issue, see my changes and commit message for `test_repogpgcheck.py`

oliverkurth:
> I got the issue, see my changes and commit message for `test_repogpgcheck.py`

That's not a fix. The problem is that your usage of `strtoi()` in `isTrue()`, which makes it print a warning. `False` is an acceptable value for `skip_if_unavailable`.

common/utils.c
oliverkurth: Nm, it does print a warning (which also caused your problem, see my other comment).

llconf/entry.c
```diff
@@ -137,7 +137,7 @@ void _cnf_find_entry(struct cnfresult **pcr, struct cnfnode *cn_parent,
 		if(*p != ']')
 			return; // FIXME: error condition
 
-		index = atoi(tmp);
+		index = strtoi(tmp);
```
oliverkurth: This only works in `tdnf` because it has `strtoi()` in its libraries. If this code were copied to another project without `strtoi()` it will fail.

I don't think we need this here. `atoi()` is just fine.

oliverkurth (changes requested):

oliverkurth:
LGTM, thank you!

PR #448, opened 2023-08-07
Add helper script to build tdnf using llvm

sshedi:
I was just playing around with this and it took a while to get it working. So added a helper script to do it.

Also, llvm build showed some good warnings, fixed them all.

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -468,7 +468,7 @@ TDNFLoadReposFromFile(
     /* coverity[var_deref_op] */
     for(cn_section = cn_conf->first_child; cn_section; cn_section = cn_section->next)
     {
-        if ((cn_section->name[0] == '.'))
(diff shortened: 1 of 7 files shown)


oliverkurth:
This is completely optional, correct? Is there any advantage in using `llvm`?

oliverkurth:
Also, I think it would be good to run all tests when built this way.

In any case, the warning fixes are good.

sshedi:
> This is completely optional, correct? Is there any advantage in using `llvm`?

Completely optional. Just dog fooding. I don't even want it to add to our workflow, llvm, clang together add ~2GB download size. It makes test runs really slow.

sshedi:
> any advantage in using llvm?

Warnings, maybe sometimes build might not work. It would be nice to keep our code compatible with gcc, clang both.

PR #449, opened 2023-08-15
Compare checksum for repodata against the downloaded rpm file

shivania2:
tdnf when downloading packages (or using from a repo accessible as a file), does not check for the checksum which is provided in the metadata. Recently there is a issue caught https://bugzilla.eng.vmware.com/show_bug.cgi?id=3237405(Fixed by checking the size), a proper check would have caught this issue earlier.


oliverkurth:
Thank you Shivani! This looks good.

PR #452, opened 2023-09-26
tdnf: Add check to identify duplicate repo id

shivania2:
When a repository(Repo id) is set multiple times, tdnf should throw an error to prevent issues.

Recently a customer was disabling all photon* repos in the files in /etc/yum.repos.d, and replaced them with another repo config, which used the same repo ids ('photon', 'photon-updates', etc.). The URLs in the original repo configs  (pointing to packages.vmware.com) were not reachable. The URLs in the replaced repo configs were reachable. However, tdnf used the URLs from the original repo configs, which lead to timeouts. Disabling the repositories is not enough - tdnf still uses the settings, and the setting that is used is arbitrary. This issue could have been prevented if tdnf threw an error.

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -28,6 +28,8 @@ TDNFLoadRepoData(
     DIR *pDir = NULL;
     struct dirent *pEnt = NULL;
     char **ppszUrlIdTuple = NULL;
+    PTDNF_REPO_DATA pRepoParsePre = NULL;
+    PTDNF_REPO_DATA pRepoParseNext = NULL;
 
     if(!pTdnf || !pTdnf->pConf || !pTdnf->pArgs || !ppReposAll)
     {
@@ -115,10 +117,20 @@ TDNFLoadRepoData(
 
         TDNF_SAFE_FREE_MEMORY(pszRepoFilePath);
         pszRepoFilePath = NULL;
-
         /* may have added multiple repos, go to last one */
         while (*ppRepoNext)
             ppRepoNext = &((*ppRepoNext)->pNext);
+    }
+
+    for (pRepoParsePre = pReposAll; pRepoParsePre; pRepoParsePre = pRepoParsePre->pNext) {
+
+        for (pRepoParseNext = pRepoParsePre->pNext; pRepoParseNext; pRepoParseNext = pRepoParseNext->pNext) {
+            if (!strcmp(pRepoParsePre->pszId, pRepoParseNext->pszId)) {
+                pr_err("ERROR: duplicate repo id: %s\n", pRepoParsePre->pszId);
+                dwError = ERROR_TDNF_DUPLICATE_REPO_ID;
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+        }
     }
 
     *ppReposAll = pReposAll;
--- a/pytests/tests/test_baseurls.py
+++ b/pytests/tests/test_baseurls.py
@@ -12,7 +12,6 @@ import pytest
 
 WORKDIR = '/root/baseurls/workdir'
 REPOFILENAME = 'baseurls.repo'
-TESTREPO = 'photon-test'
 REPONAME = 'baseurls-repo'
 
 
@@ -33,7 +32,7 @@ def teardown_test(utils):
 
 
 def test_multiple_baseurls(utils):
-    reponame = TESTREPO
+    reponame = REPONAME
     workdir = WORKDIR
     utils.makedirs(workdir)
 
--- a/pytests/tests/test_repolist.py
+++ b/pytests/tests/test_repolist.py
@@ -50,6 +50,8 @@ def setup_test(utils):
 def teardown_test(utils):
     os.remove(os.path.join(utils.config['repo_path'], 'yum.repos.d', 'foo.repo'))
     os.remove(os.path.join(utils.config['repo_path'], 'yum.repos.d', 'bar.repo'))
+    os.remove(os.path.join(utils.config['repo_path'], "yum.repos.d", 'test.repo'))
+    os.remove(os.path.join(utils.config['repo_path'], "yum.repos.d", 'test1.repo'))
 
 
 def find_repo(repolist, id):
@@ -126,3 +128,35 @@ def test_repolist_invalid(utils):
 def test_repolist_memcheck(utils):
     ret = utils.run_memcheck(['tdnf', 'repolist'])
     assert ret['retval'] == 0
+
+
+# multiple repoid
+def test_multiple_repoid(utils):
+    reponame = 'test.repo'
+    repofile_test = os.path.join(utils.config['repo_path'], 'yum.repos.d', reponame)
+    utils.edit_config(
+        {
+            'name': 'Test Repo',
+            'enabled': '1',
+            'baseurl': 'http://pkgs.test.org/test'
+        },
+        section='test',
+        filename=repofile_test
+    )
+
+    reponame = 'test1.repo'
+    repofile_test1 = os.path.join(utils.config['repo_path'], 'yum.repos.d', reponame)
+    utils.edit_config(
+        {
+            'name': 'Test Repo',
+            'enabled': '1',
+            'baseurl': 'http://pkgs.test1.org/test1'
+        },
+        section='test',
+        filename=repofile_test1
+    )
+
+    ret = utils.run(['tdnf',
+                     '--disablerepo=*', '--enablerepo={}'.format(reponame),
+                     'makecache'])
+    assert ret['retval'] == 1037
--- a/client/defines.h
+++ b/client/defines.h
@@ -232,6 +232,7 @@ typedef enum
     {ERROR_TDNF_INVALID_INPUT, "ERROR_TDNF_INVALID_INPUT", "Invalid input."},\
     {ERROR_TDNF_CACHE_DISABLED, "ERROR_TDNF_CACHE_DISABLED", "cache only is set, but no repo data found"},\
     {ERROR_TDNF_CACHE_DIR_OUT_OF_DISK_SPACE, "ERROR_TDNF_CACHE_DIR_OUT_OF_DISK_SPACE", "Insufficient disk space at cache directory /var/cache/tdnf (unless specified differently in config). Try freeing space first."},\
+    {ERROR_TDNF_DUPLICATE_REPO_ID,         "ERROR_TDNF_DUPLICATE_REPO_ID",         "Duplicate repo id"}, \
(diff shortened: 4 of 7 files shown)


client/repoutils.c
```diff
@@ -541,7 +541,9 @@ TDNFFindRepoById(
     )
 {
     uint32_t dwError = 0;
+    PTDNF_REPO_DATA parse_pRepos = NULL;
```
oliverkurth: Nit: please adhere to naming standards - name it `pRepoParse` or something. Use a suffix, not a prefix.

client/repoutils.c
```diff
@@ -553,18 +555,23 @@ TDNFFindRepoById(
...
+            nFound = 1;
+            pRepos = parse_pRepos;
+        } else if(!strcmp(pszRepo, parse_pRepos->pszId) && nFound == 1)
+        {
+            dwError = ERROR_TDNF_DUPLICATE_REPO_ID;
+            BAIL_ON_TDNF_ERROR(dwError);
         }
-        pRepos = pRepos->pNext;
+        parse_pRepos = parse_pRepos->pNext;
     }
```
oliverkurth: This is the wrong place for this check, for two reasons: this function will be called too late, or may actually not be called at all. It's utility function to find a repo, it's not intended to check the configuration. So it may also be called multiple times, and since this adds extra CPU cycles, makes it inefficient.

A better place for the check would be in `TDNFLoadRepoData()` in `client/repolist.c` at the end. You can loop through all repos there and check for dups.

Besides all that, this logic also has a flaw - we'd only detect duplicates for repositories we are looking for. If for example we are looking for `photon-bar`, and `photon-foo` is a dup, we wouldn't see that.

pytests/tests/test_baseurls.py
```diff
@@ -32,11 +32,37 @@ def teardown_test(utils):
...
+    filename = os.path.join(utils.config['repo_path'], "yum.repos.d", REPOFILENAME)
+
+    baseurls = "http://localhost:8080/photon-test"
+    utils.create_repoconf(filename, baseurls, reponame)
+
+    ret = utils.run(['tdnf',
+                     '--disablerepo=*', '--enablerepo={}'.format(reponame),
+                     'makecache'],
+                    cwd=workdir)
+    assert ret['retval'] == 0
```
oliverkurth: If we have multipe repos with the same id, `makecache` should already throw an error, so you should assert for `1037` here.

pytests/tests/test_baseurls.py
```diff
@@ -32,11 +32,37 @@ def teardown_test(utils):
...
-def test_multiple_baseurls(utils):
+def test_multiple_repoid(utils):
     reponame = TESTREPO
     workdir = WORKDIR
     utils.makedirs(workdir)
 
+    filename = os.path.join(utils.config['repo_path'], "yum.repos.d", REPOFILENAME)
+
+    baseurls = "http://localhost:8080/photon-test"
+    utils.create_repoconf(filename, baseurls, reponame)
```
oliverkurth: Does this create another repo with an id that already exists? That's not clear to me.

pytests/tests/test_baseurls.py
```diff
@@ -32,11 +32,37 @@ def teardown_test(utils):
         os.remove(filename)
 
 
-def test_multiple_baseurls(utils):
+def test_multiple_repoid(utils):
```
oliverkurth: Why is this test in `test_baseurls.py`? It has nothing to do with multiple base URLs.

oliverkurth (changes requested):

client/repoutils.c
shivania2: Added check to identify duplicate repo id in TDNFLoadRepoData() in client/repolist.c

pytests/tests/test_baseurls.py
shivania2: Added test case in test_repolist.py

pytests/tests/test_baseurls.py
shivania2: Done

pytests/tests/test_baseurls.py
shivania2: Yes, This creates another repo with the id [cfg-test], which already exist. I have removed this code in latest changes

client/repoutils.c
shivania2: Done

client/repolist.c
```diff
@@ -115,18 +127,35 @@ TDNFLoadRepoData(
...
+    dwError = TDNFAllocateMemory(nRepoId + 1, sizeof(char *), (void **)&ppRepoId);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    pRepoParse = pReposAll;
+
+    while (pRepoParse != NULL) {
+        uint32_t hash = hashString(pRepoParse->pszId) % nRepoId;
+        if (ppRepoId[hash] != NULL && strcmp(ppRepoId[hash], pRepoParse->pszId) == 0) {
+            dwError = ERROR_TDNF_DUPLICATE_REPO_ID;
+            BAIL_ON_TDNF_ERROR(dwError);
```
oliverkurth: We don't need to use hashes. Just loop twice, it's okay, even when it's O(n^2) - there aren't that many repo configs.

pytests/tests/test_baseurls.py
```diff
@@ -33,7 +33,7 @@ def teardown_test(utils):
 
 
 def test_multiple_baseurls(utils):
-    reponame = TESTREPO
```
oliverkurth, line 36: Why do you change this?

pytests/tests/test_repolist.py
```diff
@@ -126,3 +127,23 @@ def test_repolist_invalid(utils):
 def test_repolist_memcheck(utils):
     ret = utils.run_memcheck(['tdnf', 'repolist'])
     assert ret['retval'] == 0
+
+
+# multiple repoid
+def test_multiple_repoid(utils):
+    reponame = 'bar1.repo'
+    repofile_bar1 = os.path.join(utils.config['repo_path'], 'yum.repos.d', reponame)
+    utils.edit_config(
```
oliverkurth, line 137: Wouldn't you need to add the repo with id `bar` twice to get the error?

oliverkurth (changes requested):

client/repolist.c
```diff
@@ -115,12 +117,27 @@ TDNFLoadRepoData(
 
         TDNF_SAFE_FREE_MEMORY(pszRepoFilePath);
         pszRepoFilePath = NULL;
-
         /* may have added multiple repos, go to last one */
         while (*ppRepoNext)
             ppRepoNext = &((*ppRepoNext)->pNext);
     }
 
+    pRepoParsePre = pReposAll;
```
sshedi:
```
for (pRepoParsePre = pReposAll;
        pRepoParsePre;
        pRepoParsePre = pRepoParsePre->pNext) {
    for (pRepoParseNext = pRepoParsePre->pNext;
            pRepoParseNext;
            pRepoParseNext = pRepoParseNext->pNext) {
        if (!strcmp(pRepoParsePre->pszId, pRepoParseNext->pszId)) {
            dwError = ERROR_TDNF_DUPLICATE_REPO_ID;
            BAIL_ON_TDNF_ERROR(dwError);
        }
    }
}
```

I would suggest checking of baseurl instead of ID. ID is not a primary key but baseurl is.
If baseurls are same, throw duplicate error.

sshedi (changes requested):

client/repolist.c
oliverkurth: No, that's just plain wrong. The id is the primary key. There is no issue at all if a `baseurl` is duplicated - it's not advisable, but not really an issue.

Having duplicate ids however is a problem.That is why we are doing this.

pytests/tests/test_baseurls.py
oliverkurth: Can you please revert this? I think you left it over accidentally.

client/repolist.c
```diff
@@ -115,12 +117,27 @@ TDNFLoadRepoData(
...
+        while (pRepoParseNext->pNext != NULL) {
+            if (!strcmp(pRepoParsePre->pszId, pRepoParseNext->pNext->pszId)) {
+                dwError = ERROR_TDNF_DUPLICATE_REPO_ID;
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+            else
+                pRepoParseNext = pRepoParseNext->pNext;
+        }
+        pRepoParsePre = pRepoParsePre->pNext;
+    }
```
oliverkurth: Thank you, this looks good. I liked the version using `for` better though because it makes the algorithm easier to understand (and I can only see it from Shreenidhi's comment).

client/repolist.c
```diff
@@ -115,12 +117,27 @@ TDNFLoadRepoData(
...
 
+    pRepoParsePre = pReposAll;
+
+    while (pRepoParsePre != NULL && pRepoParsePre->pNext != NULL) {
+        pRepoParseNext = pRepoParsePre;
+ 
+        while (pRepoParseNext->pNext != NULL) {
+            if (!strcmp(pRepoParsePre->pszId, pRepoParseNext->pNext->pszId)) {
+                dwError = ERROR_TDNF_DUPLICATE_REPO_ID;
+                BAIL_ON_TDNF_ERROR(dwError);
```
oliverkurth, line 131: Please also add a `pr_err` with an error message that also prints the id, so we know exactly which id is duplicated.

oliverkurth (changes requested):
This looks good, just minor comments.

I think the error in the tests is because of the change you left over in `test_baseulrd.py`, even though the failure is in `test_setopt_reposdiir.py` - this may be in bug not from this PR. So please revert that and try again.

client/repolist.c
```diff
@@ -115,12 +117,22 @@ TDNFLoadRepoData(
...
         /* may have added multiple repos, go to last one */
         while (*ppRepoNext)
             ppRepoNext = &((*ppRepoNext)->pNext);
     }
 
+    for (pRepoParsePre = pReposAll; pRepoParsePre; pRepoParsePre = pRepoParsePre->pNext) {
+
+        for (pRepoParseNext = pRepoParsePre->pNext; pRepoParseNext; pRepoParseNext = pRepoParseNext->pNext) {
+            if (!strcmp(pRepoParsePre->pszId, pRepoParseNext->pszId)) {
+                pr_err("ERROR: Duplicate Repoid (%s)\n", pRepoParsePre->pszId);
```
oliverkurth: Nit: better is "duplicate repo id: '%s'".

oliverkurth (commented):
Thanks again! Just a tiny nit.

We need to check why the tests are failing - it looks to me that we already had duplicate repo ids in the test and we are only noticing them now.

oliverkurth:
> We need to check why the tests are failing - it looks to me that we already had duplicate repo ids in the test and we are only noticing them now.

That was indeed the issue. I have a fix here: https://github.com/oliverkurth/tdnf/commit/a859f3399bc4aec7ba5f042f40fd9ab0f56bdbd8 . You can apply those changes to yours.

Proof that that succeeds: https://github.com/oliverkurth/tdnf/actions/runs/6356463459

oliverkurth (approved):

PR #453, opened 2023-10-09
Set a user agent for http/https tdnf requests

shivania2:
tdnf does not set a user agent for http/https requests and the repository `pkgs.k8s.io` did block requests without user agents (because its the default policy in AWS WAF). This issue is fixed in kubernetes upstream repo. But to avoid this kind of issue in future, adding user agent for http/https tdnf requests

### DIFF
--- a/client/config.c
+++ b/client/config.c
@@ -13,6 +13,9 @@
 #include "../llconf/entry.h"
 #include "../llconf/ini.h"
 
+#define USERAGENT_HEADER_MAX_LENGTH 256
+#define OS_CONF_FILE "/etc/os-release"
+
 int
 TDNFConfGetRpmVerbosity(
     PTDNF pTdnf
@@ -24,6 +27,59 @@ TDNFConfGetRpmVerbosity(
         nLogLevel = pTdnf->pArgs->nRpmVerbosity;
     }
     return nLogLevel;
+}
+
+static uint32_t TDNFParseOSInfo(PTDNF_CONF pConf)
+{
+    char buf[USERAGENT_HEADER_MAX_LENGTH];
+    char *name = NULL;
+    char *version = NULL;
+    uint32_t dwError = 0;
+    FILE *file = NULL;
+
+    file = fopen(OS_CONF_FILE, "r");
+
+    if (!file) {
+        pr_info("Warning: %s file is not present in the system\n", OS_CONF_FILE);
+        return 0;
+    }
+
+    while (fgets(buf, USERAGENT_HEADER_MAX_LENGTH, file))
+    {
+        if (strncmp("ID=", buf, sizeof("ID=")-1) == 0)
+        {
+            if (sscanf(buf, "ID=\"%[^\"]\"", buf) == 1) {
+                dwError = TDNFAllocateString(buf, &name);
+            } else if (sscanf(buf, "ID=%s", buf) == 1) {
+                dwError = TDNFAllocateString(buf, &name);
+            }
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+        else if (strncmp("VERSION_ID=", buf, sizeof("VERSION_ID=")-1) == 0)
+        {
+            if (sscanf(buf, "VERSION_ID=\"%[^\"]\"", buf) == 1) {
+                dwError = TDNFAllocateString(buf, &version);
+            } else if (sscanf(buf, "VERSION_ID=%s", buf) == 1) {
+                dwError = TDNFAllocateString(buf, &version);
+            }
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+    }
+
+    pConf->pszOSName = name;
+    pConf->pszOSVersion = version;
+
+cleanup:
+    if(file)
+    {
+        fclose(file);
+    }
+    return dwError;
+
+error:
+    TDNF_SAFE_FREE_MEMORY(name);
+    TDNF_SAFE_FREE_MEMORY(version);
+    goto cleanup;
 }
 
 uint32_t
@@ -40,6 +96,7 @@ TDNFReadConfig(
     char *pszPkgLocksDir = NULL;
     char *pszProtectedDir = NULL;
 
+    const char *pszTdnfVersion = NULL;
     const char *pszProxyUser = NULL;
     const char *pszProxyPass = NULL;
 
@@ -90,6 +147,9 @@ TDNFReadConfig(
             BAIL_ON_TDNF_ERROR(dwError);
         }
     }
+
+    dwError = TDNFParseOSInfo(pConf);
+    BAIL_ON_TDNF_ERROR(dwError);
 
     /* cn_conf == NULL => we will not reach here */
     /* coverity[var_deref_op] */
@@ -191,6 +251,17 @@ TDNFReadConfig(
             BAIL_ON_TDNF_ERROR(dwError);
         }
     }
+
+    pszTdnfVersion = TDNFGetVersion();
+
+    if (pConf->pszOSName == NULL)
+        TDNFAllocateString("UNKNOWN", &pConf->pszOSName);
+
+    if (pConf->pszOSVersion == NULL)
+        TDNFAllocateString("UNKNOWN", &pConf->pszOSVersion);
+
+    dwError = TDNFAllocateStringPrintf(&pConf->pszUserAgentHeader, "tdnf/%s %s/%s", pszTdnfVersion, pConf->pszOSName, pConf->pszOSVersion);
+    BAIL_ON_TDNF_ERROR(dwError);
 
     /* if plugins are not enabled explicitely,
        we have to disable them because it's the default */
@@ -326,6 +397,9 @@ TDNFFreeConfig(
         TDNF_SAFE_FREE_MEMORY(pConf->pszVarReleaseVer);
         TDNF_SAFE_FREE_MEMORY(pConf->pszVarBaseArch);
         TDNF_SAFE_FREE_MEMORY(pConf->pszBaseArch);
+        TDNF_SAFE_FREE_MEMORY(pConf->pszUserAgentHeader);
+        TDNF_SAFE_FREE_MEMORY(pConf->pszOSName);
+        TDNF_SAFE_FREE_MEMORY(pConf->pszOSVersion);
         TDNF_SAFE_FREE_STRINGARRAY(pConf->ppszExcludes);
         TDNF_SAFE_FREE_STRINGARRAY(pConf->ppszMinVersions);
         TDNF_SAFE_FREE_STRINGARRAY(pConf->ppszPkgLocks);
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -194,6 +194,15 @@ TDNFDownloadFile(
         BAIL_ON_TDNF_ERROR(dwError);
     }
 
+    if(!IsNullOrEmptyString(pTdnf->pConf->pszUserAgentHeader))
+    {
+        dwError = curl_easy_setopt(
+                      pCurl,
+                      CURLOPT_USERAGENT,
+                      pTdnf->pConf->pszUserAgentHeader);
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
     dwError = TDNFRepoApplyProxySettings(pTdnf->pConf, pCurl);
     BAIL_ON_TDNF_ERROR(dwError);
 
--- a/include/tdnftypes.h
+++ b/include/tdnftypes.h
@@ -265,6 +265,9 @@ typedef struct _TDNF_CONF
     char* pszBaseArch;
     char* pszVarReleaseVer;
     char* pszVarBaseArch;
+    char* pszUserAgentHeader;
+    char* pszOSName;
+    char* pszOSVersion;
     char** ppszExcludes;
     char** ppszMinVersions;
     char** ppszPkgLocks;


client/config.c
```diff
@@ -13,6 +13,8 @@
 #include "../llconf/entry.h"
 #include "../llconf/ini.h"
 
+#define UserAgentHeaderMaxLength 50
```
oliverkurth: Use ALL CAPS for macros.

client/config.c
```diff
@@ -26,6 +28,51 @@ TDNFConfGetRpmVerbosity(
...
+    uint32_t dwError = 0;
+    FILE *file = NULL;
+    file = fopen("/etc/os-release", "r");
+
+    if (!file) {
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(2*UserAgentHeaderMaxLength, sizeof(char), (void**)&line);
```
oliverkurth: You can use a temp buffer on the stack, then use `TDNFAllocate String()` to the the `pConf` member.

client/config.c
```diff
@@ -26,6 +28,51 @@ TDNFConfGetRpmVerbosity(
...
+
+    dwError = TDNFAllocateMemory(UserAgentHeaderMaxLength, sizeof(char), (void**)&pConf->pszOSName);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFAllocateMemory(UserAgentHeaderMaxLength, sizeof(char), (void**)&pConf->pszOSVersion);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    while (fgets(line, 2*UserAgentHeaderMaxLength, file)) {
+        if (strstr(line, "ID=") != NULL) {
+            sscanf(line, "ID=\"%[^\"]\"", pConf->pszOSName);
```
oliverkurth: Have you tested that this works with both unquoted and quoted settings?

client/config.c
```diff
@@ -192,6 +242,32 @@ TDNFReadConfig(
...
+    strcat(pszUserAgentHeader, "/");
+
+    if (pConf->pszOSVersion != NULL)
+        strcat(pszUserAgentHeader, pConf->pszOSVersion);
+    else
+        strcat(pszUserAgentHeader, "UNKNOWN_OS_VERSION");
+
+    dwError = TDNFAllocateString(pszUserAgentHeader, &pConf->pszUserAgentHeader);
+    BAIL_ON_TDNF_ERROR(dwError);
+
```
oliverkurth: This can be made much simpler with `TDNFAllocateStringPrintf()`.

client/prototypes.h
```diff
@@ -543,6 +543,11 @@ TDNFConfGetRpmVerbosity(
     PTDNF pTdnf
     );
 
+uint32_t
+TDNFParseOSInfo(
```
oliverkurth: You can make this a static function in `config.c`.

client/config.c
```diff
@@ -192,6 +242,32 @@ TDNFReadConfig(
...
+    dwError = TDNFAllocateMemory(UserAgentHeaderMaxLength, sizeof(char), (void**)&pszUserAgentHeader);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+	strcpy(pszUserAgentHeader, "User-Agent:tdnf/");
+    if (pConf->pszVarReleaseVer != NULL) {
+	    strcat(pszUserAgentHeader, pConf->pszVarReleaseVer);
+        strcat(pszUserAgentHeader, " ");
+    }
+    else
+        strcat(pszUserAgentHeader, "UNKNOWN ");
```
oliverkurth: Nit: use "unknown". No need to yell.

client/config.c
```diff
@@ -192,6 +242,32 @@ TDNFReadConfig(
...
+
+    if (pConf->pszOSName != NULL)
+	    strcat(pszUserAgentHeader, pConf->pszOSName);
+    else
+        strcat(pszUserAgentHeader, "UNKNOWN_OS_NAME");
```
oliverkurth: Also here, and and just "unknown" is enough.

client/config.c
```diff
@@ -192,6 +242,32 @@ TDNFReadConfig(
         }
     }
 
+    dwError = TDNFAllocateMemory(UserAgentHeaderMaxLength, sizeof(char), (void**)&pszUserAgentHeader);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+	strcpy(pszUserAgentHeader, "User-Agent:tdnf/");
+    if (pConf->pszVarReleaseVer != NULL) {
```
oliverkurth: Use the `tdnf` version. The `releaseVer` is something different.

oliverkurth (changes requested):
We also need a test.

client/config.c
```diff
@@ -91,6 +139,8 @@ TDNFReadConfig(
         }
     }
 
+    TDNFParseOSInfo(pConf);
```
sshedi: validate return value here

client/config.c
```diff
@@ -192,6 +242,32 @@ TDNFReadConfig(
         }
     }
 
+    dwError = TDNFAllocateMemory(UserAgentHeaderMaxLength, sizeof(char), (void**)&pszUserAgentHeader);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+	strcpy(pszUserAgentHeader, "User-Agent:tdnf/");
```
sshedi: `"User-Agent:tdnf/"` storing these keywords in macros would help

client/prototypes.h
sshedi: Can we use config parser for parsing `/etc/os-release`?

Also, we have some info on OS with `uname` APIs? can we use that here as well?

client/config.c
```diff
@@ -192,6 +242,32 @@ TDNFReadConfig(
...
+
+    if (pConf->pszOSName != NULL)
+	    strcat(pszUserAgentHeader, pConf->pszOSName);
+    else
+        strcat(pszUserAgentHeader, "UNKNOWN_OS_NAME");
+
+    strcat(pszUserAgentHeader, "/");
+
+    if (pConf->pszOSVersion != NULL)
+        strcat(pszUserAgentHeader, pConf->pszOSVersion);
```
sshedi: use `snprintf` here.

client/config.c
sshedi: use `snprintf` for this

client/config.c
```diff
@@ -26,6 +28,51 @@ TDNFConfGetRpmVerbosity(
...
 
+uint32_t
+TDNFParseOSInfo(
+    PTDNF_CONF pConf
+    )
+{
+    char *line = NULL;
+    uint32_t dwError = 0;
+    FILE *file = NULL;
+    file = fopen("/etc/os-release", "r");
```
sshedi: new line before line 39 and store etc/os-release path in macro

client/config.c
```diff
@@ -26,6 +28,51 @@ TDNFConfGetRpmVerbosity(
...
+
+    if (!file) {
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(2*UserAgentHeaderMaxLength, sizeof(char), (void**)&line);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    dwError = TDNFAllocateMemory(UserAgentHeaderMaxLength, sizeof(char), (void**)&pConf->pszOSName);
```
sshedi: 50 bytes for name & version seems like an overkill and allocate memory based on line length inside loop.

shivania2:
> We also need a test.

Exploring the possible test case

client/config.c
```diff
@@ -26,6 +29,49 @@ TDNFConfGetRpmVerbosity(
...
+
+    file = fopen(OSCONFFILE, "r");
+
+    if (!file) {
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+
+    while (fgets(temp, USERAGENTHEADERMAXLENGTH, file)) {
+        if (strstr(temp, "ID=") != NULL && name_found == 0) {
```
sshedi: use strncmp("ID=", temp, sizeof("ID=")-1);

no need of name_found and it will make this logic bit more robust

client/config.c
```diff
@@ -26,6 +29,49 @@ TDNFConfGetRpmVerbosity(
...
+    FILE *file = NULL;
+
+    file = fopen(OSCONFFILE, "r");
+
+    if (!file) {
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+
+    while (fgets(temp, USERAGENTHEADERMAXLENGTH, file)) {
```
sshedi: Use `getline` instead. Example here https://stackoverflow.com/questions/3501338/c-read-file-line-by-line

client/config.c
```diff
@@ -26,6 +29,49 @@ TDNFConfGetRpmVerbosity(
     return nLogLevel;
 }
 
+static uint32_t TDNFParseOSInfo(PTDNF_CONF pConf)
+{
+    char temp[USERAGENTHEADERMAXLENGTH];
+    char name[10];
```
sshedi: Dynamically allocate memory for name and version and simply assign them to pConf->pszOSName and version.
Later it gets freed anyway.

client/prototypes.h
oliverkurth: > Can we use config parser for parsing /etc/os-release?

No.

> Also, we have some info on OS with uname APIs? can we use that here as well?

I don't think it's needed.

client/config.c
oliverkurth: It should be `TDNFAllocateStringPrintf()`, see my comment above.

client/config.c
oliverkurth: If we use `TDNFAllocateStringPrintf()` it will be a format string.

client/config.c
oliverkurth: `fgets()` is just fine.

client/config.c
```diff
@@ -13,6 +13,9 @@
 #include "../llconf/entry.h"
 #include "../llconf/ini.h"
 
+#define USERAGENTHEADERMAXLENGTH 50
```
oliverkurth: Please make this better readable, use like `USERAGENT_HEADER_MAXLENGTH`

client/config.c
```diff
@@ -26,6 +29,52 @@ TDNFConfGetRpmVerbosity(
...
+    FILE *file = NULL;
+
+    file = fopen(OSCONFFILE, "r");
+
+    if (!file) {
+        dwError = errno;
+        BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+    }
+
+    dwError = TDNFAllocateMemory(10, sizeof(char), (void**)&name);
```
oliverkurth: Don't pre-allocate, parse the strings into the temp buffer first, then use `TDNFAllocateString()` to allocate the exact amount you need.

oliverkurth (changes requested):

client/config.c
```diff
@@ -192,6 +245,17 @@ TDNFReadConfig(
...
 
+    pszTdnfVersion = TDNFGetVersion();
+
+    if (pConf->pszOSName == NULL)
+        TDNFAllocateString("UNKNOWN", &pConf->pszOSName);
+
+    if (pConf->pszOSVersion == NULL)
+        TDNFAllocateString("UNKNOWN", &pConf->pszOSVersion);
+
+    dwError = TDNFAllocateStringPrintf(&pConf->pszUserAgentHeader, "User-Agent:tdnf/%s %s/%s", pszTdnfVersion, pConf->pszOSName, pConf->pszOSVersion);
```
oliverkurth: You need a space after "User-Agent:". But, I think it doesn't need to be part of the string at all. See example at https://curl.se/libcurl/c/CURLOPT_USERAGENT.html 
```
curl_easy_setopt(curl, CURLOPT_USERAGENT, "Dark Secret Ninja/1.0");
```

Also make sure it's compliant with RFC 7231, see https://www.rfc-editor.org/rfc/rfc7231#page-46

Also, I don't see anywhere where you are setting the header for curl. At some point you will have to call `curl_easy_setopt(curl, CURLOPT_USERAGENT, ...`

client/config.c
sshedi: `fgets` works here but we assume that max line length is 50 bytes, with `getline` we don't need not make any assumptions and fits perfectly for this use case.

oliverkurth: 50 characters is too short, we can just create a 256 byte buffer on the stack, and no serious `/etc/os-release` file will have lines any longer, and even if we can skip the rest. We can use `getline()`, I just think it's overkill.

client/config.c
```diff
@@ -91,6 +141,9 @@ TDNFReadConfig(
         }
     }
 
+    dwError = TDNFParseOSInfo(pConf);
+    BAIL_ON_TDNF_ERROR(dwError);
```
oliverkurth, line 152: We should not bail here if the error is that the file wasn't found. In that case put "unknown" as the OS and version.

client/config.c
```diff
@@ -26,6 +29,50 @@ TDNFConfGetRpmVerbosity(
...
+    char temp[USERAGENT_HEADER_MAX_LENGTH];
+    char *name = NULL;
+    char *version = NULL;
+    uint32_t dwError = 0;
+    FILE *file = NULL;
+
+    file = fopen(OS_CONF_FILE, "r");
+
+    if (!file) {
+        dwError = errno;
```
sshedi: print a warning here and return 0 perhaps?

client/config.c
```diff
@@ -26,6 +29,50 @@ TDNFConfGetRpmVerbosity(
     return nLogLevel;
 }
 
+static uint32_t TDNFParseOSInfo(PTDNF_CONF pConf)
+{
+    char temp[USERAGENT_HEADER_MAX_LENGTH];
```
sshedi: I would still say we should avoid static allocation here and use getline. It's upto you on what you want to use.

Also, be aware that fgets doesn't always null terminate the buffer. By chance if there is a line which is more than 50 bytes in os-release file, this might end up crashing tdnf.

client/config.c
```diff
@@ -326,6 +390,9 @@ TDNFFreeConfig(
         TDNF_SAFE_FREE_MEMORY(pConf->pszVarReleaseVer);
         TDNF_SAFE_FREE_MEMORY(pConf->pszVarBaseArch);
         TDNF_SAFE_FREE_MEMORY(pConf->pszBaseArch);
+        TDNF_SAFE_FREE_MEMORY(pConf->pszUserAgentHeader);
```
sshedi, line 400: Can we have a function of it's own to free up pConf structure? this is too much.

client/config.c
```diff
@@ -91,6 +139,11 @@ TDNFReadConfig(
         }
     }
 
+    dwError = TDNFParseOSInfo(pConf);
+    if (dwError != ENOENT) {
```
sshedi: I have a question. If we the os-release file doesn't exist and it's considered not a fatal error, why should we consider other errors in `TDNFParseOSInfo` as fatal? why not use some default values in case of failure and make the return type void?

client/config.c
```diff
@@ -26,6 +29,54 @@ TDNFConfGetRpmVerbosity(
...
+    {
+        fclose(file);
+    }
+    return dwError;
+
+error:
+    TDNF_SAFE_FREE_MEMORY(name);
+    TDNF_SAFE_FREE_MEMORY(version);
+    goto cleanup;
+
```
sshedi: delete the empty line please

client/config.c
sshedi: Shivani, any thoughts on this? I also gave somewhat similar comment yesterday.
https://github.com/vmware/tdnf/pull/453#discussion_r1360174875

client/config.c
sshedi: Any thoughts on this?

client/config.c
oliverkurth: I think `/etc/os-release` not present is an acceptable condition, but failing to parse it for any reason is not. It may be some deeper issue (a bug in the parser, file corrupted or whatever) , and we shouldn't proceed.

client/config.c
oliverkurth: Good point I think. Doesn't have to be necessarily in the change, but good to have.

client/config.c
oliverkurth: I replied to that comment. I think this way is good.

client/config.c
oliverkurth: That's not what the man page says:
```
fgets()  reads in at most one less than size characters from stream and stores them into the buffer pointed to by s.  Reading stops after an EOF or a newline.  If a newline is read, it is stored into
       the buffer.  A terminating null byte ('\0') is stored after the last character in the buffer.
```
Instead it seems to take care that it is aways terminated with 0.

client/config.c
```diff
@@ -26,6 +29,54 @@ TDNFConfGetRpmVerbosity(
...
+    if (!file) {
+        pr_info("Warning: %s file is not present in the system\n", OS_CONF_FILE);
+        return 0;
+    }
+
+    while (fgets(buf, USERAGENT_HEADER_MAX_LENGTH, file))
+    {
+        if (strncmp("ID=", buf, sizeof("ID=")-1) == 0)
+        {
+            sscanf(buf, "ID=%s\n", buf);
```
oliverkurth: You need to take care of quotes, which may or may not be present. For example this is in Ubuntu:
```
$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
...
```
This is in Photon:
```
# cat /etc/os-release 
NAME="VMware Photon OS"
VERSION="5.0"
ID=photon
VERSION_ID=5.0
...
```
Note that Ubuntu has `VERSION_ID` in quotes, but Photon does not. You need to check for quotes and remove them if there.

oliverkurth (changes requested):

sshedi (approved):

pytests/tests/test_useragent.py
oliverkurth, line 1: Thanks for writing a test! Are you going to finish this? I don't see where we are testing the header sent by tdnf (we are not calling it at all here).

client/config.c
oliverkurth: Ah, this *is* in `TDNFFreeConfig()` already ... I think the UI tricked us here.

oliverkurth (commented):
As stated per DM, let's drop the test change, it takes too much effort. I verified locally that the header is sent on requests. The change is good, I can merge this, without a test.

shivania2:
> As stated per DM, let's drop the test change, it takes too much effort. I verified locally that the header is sent on requests. The change is good, I can merge this, without a test.

Thank You. I will remove the test case

PR #464, opened 2024-01-11
refactor --setopts handling

oliverkurth:
This refactors the implementation of `--setopt=` for parity with `dnf`: any setting with `--setopt` should override config settings, or in the case of lists (like `baseurl`) append to them. With the patterns `--setopt=<repo>.foo=` a setting in the repo config will be reset.

This also adds support for all rpm transaction flags, and support for setting them in the config file.


common/utils.c
```diff
@@ -1273,3 +1271,16 @@ TDNFChecksumFromHexDigest(
 error:
     goto cleanup;
 }
+
+/* return true if str is a valid reponame */
+int TDNFStrIsValidRepoName(const char *str)
+{
+    /* should start with alnum or '@' */
+    /* '@ as a first character to allow "@System" and "@cmdline" */
+    if (!str || !*str || !(isalnum(*str) || (*str == '@')))
```
sshedi: nit: Use `IsNullOrEmptyString`

common/strings.c
```diff
@@ -169,6 +169,85 @@ TDNFSplitStringToArray(
...
+
+    for (i = 0; i < n1; i++) {
+        (*pppszArray0)[n0 + i] = ppszArray1[i];
+    }
+    (*pppszArray0)[n] = NULL;
+
+    /* intentionally free only the pointer list,
+       and not the strings, because those have been moved to
+       pppszArray0 */
+    TDNF_SAFE_FREE_MEMORY(ppszArray1);
```
sshedi, line 205: Once this gets freed, what happens to `pppszArray0` ? at line 198 it is just pointer assignment right?

client/config.c
```diff
@@ -210,10 +244,45 @@ TDNFConfigFromCnfTree(PTDNF_CONF pConf, struct cnfnode *cn_top)
         {
             SET_STRING(pConf->pszPluginPath, cn->value);
         }
+        else if  (strcmp(cn->name, TDNF_CONF_KEY_TSFLAGS) == 0)
+        {
+            if (cn->value == NULL || strcmp(cn->value, "") == 0) {
+                pConf->rpmTransFlags = RPMTRANS_FLAG_NONE;
+            } else {
+                char *value = strdup(cn->value);
```
sshedi: validate strdup retval for null

client/config.c
```diff
@@ -210,10 +244,45 @@ TDNFConfigFromCnfTree(PTDNF_CONF pConf, struct cnfnode *cn_top)
...
+                int i;
+
+                for (str = value; ; str = NULL){
+                    token = strtok_r(str, " ", &saveptr);
+                    if (token == NULL)
+                        break;
+
+                    for (i = 0; rpmtransflags_map[i].name != NULL; i++) {
+                        if (strcmp(token, rpmtransflags_map[i].name) == 0){
+                            pConf->rpmTransFlags |= rpmtransflags_map[i].flag;
```
sshedi, line 265: should `rpmTransFlags` should be set to zero before entering loop?

llconf/nodes.c
```diff
@@ -157,6 +172,23 @@ void cnfnode_setname(struct cnfnode *cn, const char *name)
...
+ * @param cn pointer to a node
+ * @param name pointer to the new name
+ * @param n length of the name to be used
+ */
+void cnfnode_setname_n(struct cnfnode *cn, const char *name, size_t n)
+{
+	if(cn){
+		if(cn->name) free(cn->name);
+		if(name)
+			cn->name = strndup(name, n);
```
sshedi, line 186: null check missing

client/config.c
```diff
@@ -314,6 +383,51 @@ TDNFReadConfig(
         pConf->nSkipSignature = 1;
     }
 
+    if (pConf->pszRepoDir == NULL)
+        pConf->pszRepoDir = strdup(TDNF_DEFAULT_REPO_LOCATION);
```
sshedi: null check

client/config.c
```diff
@@ -314,6 +383,51 @@ TDNFReadConfig(
...
+    if (pConf->pszCacheDir == NULL)
+        pConf->pszCacheDir = strdup(TDNF_DEFAULT_CACHE_LOCATION);
```
sshedi: null check

common/strings.c
oliverkurth: Yes. The ownership for the strings gets transfered from `ppszArray1` to `pppszArray0`. The latter is owned by the caller.

client/config.c
oliverkurth: No, it should definitely not. This function may be called multiple times - first for the config which may set flags, and then again for the `--setopt` options, which may set additional flags without resetting them. Only if we explicirely set them to "" should they be reset. See https://dnf.readthedocs.io/en/latest/command_ref.html:

"Values for configuration options like excludepkgs, includepkgs, installonlypkgs and tsflags are appended to the original value, they do not override it. However, specifying an empty value (e.g. --setopt=tsflags=) will clear the option."

client/config.c
oliverkurth: I modified the `SET_STRING` macro and used that.

sshedi (approved):

Issue #467, opened 2024-02-29
baseurl does not properly handle http GET query parameter


SeanDougherty:
### Describe the bug

When utilizing a baseurl that is hosted with access control measures in place, tdnf has difficulty attempting to properly append paths to the baseurl.

E.g.,
A package repository is hosted in a cloud service and access to anything within the cloud storage is handled through passing of http query parameters
(For example, in `https://github.com/vmware/tdnf/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml` url, the substring `?assignees=&labels=bug&projects=&template=bug-report.yml` is the query parameter)
In the cloud storage situation, the unauthenticated baseurl may look like: `foo.cloudservice.storage.com/reporoot` and the authenticated baseurl will look like: `foo.cloudservice.storage.com/reporoot?<authenticationtoken>`.

This authentication token must be appended to every HTTP get request when interacting with the cloud storage service.

In TDNF, there is no handling of this query substring, so the final path will get appended after this query parameter.
> `foo.cloudservice.storage.com/reporoot?<authenticationtoken>/repodata/repomd.xml`
instead of 
> `foo.cloudservice.storage.com/reporoot/repodata/repomd.xml?<authenticationtoken>`

This prevents the use of tdnf with secured package repositories.

When a baseurl contains an http query parameter, tdnf fails to retrieve the repomd.xml and will cease to function, causing a non-zero exit code. 

### Reproduction steps

1. Create a private container in Azure Storage (Please reach out if I can help with this)
2. Create a .repo file that points to the authenticated baseurl of the storage account
3. Attempt to call `tdnf makecache`
...

### Expected behavior

tdnf makecache, and all other tdnf commands should behave the same as with an unauthenticated baseurl

oliverkurth:
If I understand correctly, you have this configured in your `*.repo` file:
```
baseurl=https://foo.cloudservice.storage.com/reporoot?<authenticationtoken>
```
And `tdnf` just mindlessly appends `/repodata/repomd.xml` to it, yielding an invalid URL, correct?

SeanDougherty:
This is correct! Thank you for the fast response 👍

oliverkurth:
Thinking about this, I get more questions:

[...]

PR #469, opened 2024-03-05
Allow Query String Parameters in the URL

adithyaj:
As discussed in #467, when the baseURL has appended query string parameters, tdnf does not recognize this as a query string parameter and treats this as a standard baseURL and appends paths like the `repomd.xml` to the end of this. For query strings, the expectation is that these parameters would appear at the very end of the URL, after the paths have been appended. 

The following is a change to the `common/utils.c` path to update the URL in the event we see a query string.

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -810,6 +810,8 @@ TDNFJoinPath(char **ppszPath, ...)
     char *pszNodeCopy = NULL;
     char *pszResult = NULL;
     int nLength = 0;
+    int nIsRemoteUrl = 0;
+    char *pszUrlQueryStringCopy = NULL;
 
     if (!ppszPath)
     {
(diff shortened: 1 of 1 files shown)


common/utils.c
```diff
@@ -836,6 +838,19 @@ TDNFJoinPath(char **ppszPath, ...)
          * before stripping all leading slashes */
         if (i == 0)
         {
+            /* we also check to see if the base path is a remote URL; for URLs
```
oliverkurth, line 841: I don't really like that you are using `TDNFJoinPath()` for this, which is a generic function to join path elements (much like python's `os.path.join()`. I think it's not very efficient, because this code will now be used in all places where we join paths, which are mostly local.

Instead, I think we should create another version of this function that is only used where we need this functionality  of preserving the URL query string.

oliverkurth: @adithyaj , are you going to address my comments?

oliverkurth:
Is this still an issue? @adithyaj , will you address my review comments? Otherwise, I will close this.

oliverkurth:
Closing this, looks like the issue is stalled.

Feel free to reopen, or create another PR with my comments addressed.

PR #470, opened 2024-03-14
Address some cppcheck warnings

sshedi:


client/config.c
```diff
@@ -119,7 +119,6 @@ TDNFReadConfig(
 
     /* defaults */
     pConf->nGPGCheck = 0;
-    pConf->nInstallOnlyLimit = 1;
```
oliverkurth, line 122: Why are you removing this?

llconf/modules.c
```diff
@@ -237,12 +236,12 @@ int cnfmodule_unparse(struct cnfmodule *cm, FILE* fout,
 int cnfmodule_unparse_file(struct cnfmodule *cm, const char *fname,
 			   struct cnfnode *cn_root)
 {
-	FILE *fout;
 	int ret = -1;
 
 	if(fname == NULL)
 		fname = cm->default_file;
 	if(fname){
+	    FILE *fout;
```
oliverkurth: this looks misaligned

llconf/nodes.c
```diff
@@ -377,7 +377,8 @@ char *cnfnode_path(struct cnfnode *cn)
 	*q0 = 0;
 
 	for(cn_parent = cn; cn_parent; cn_parent = cn_parent->parent){
-		char *p, *q;
+		const char *p;
+        char *q;
```
oliverkurth, line 381: misaligned

plugins/metalink/utils.c
```diff
@@ -471,8 +459,12 @@ TDNFXmlParseData(
...
     if(!elementInfo || !elementInfo->ml_ctx || IsNullOrEmptyString(elementInfo->filename) || (elementInfo->dwError != 0))
     {
-        elementInfo->dwError = ERROR_TDNF_INVALID_PARAMETER;
-        BAIL_ON_TDNF_ERROR(elementInfo->dwError);
+        if (elementInfo) {
+            elementInfo->dwError = ERROR_TDNF_INVALID_PARAMETER;
+            BAIL_ON_TDNF_ERROR(elementInfo->dwError);
+        }
+        pr_err("%s:%d elementInfo is NULL\n", __func__, __LINE__);
+        return;
```
oliverkurth: Good catching the `NULL` pointer check, but I think there should be a `BAIL_ON_TDNF_ERROR` here.

solv/tdnfquery.c
```diff
@@ -505,10 +505,7 @@ SolvGenerateCommonJob(
     {
         queue_empty(&queueJob);
         queue_push2(&queueJob, SOLVER_SOLVABLE_ALL, 0);
-        if (pQuery->queueRepoFilter.count)
-        {
-            selection_filter(pPool, &queueJob, &pQuery->queueRepoFilter);
-        }
+        selection_filter(pPool, &queueJob, &pQuery->queueRepoFilter);
```
oliverkurth, line 508: Why are you removing the `if (...` here?

oliverkurth (changes requested):
Just a few minor comments, but you need to fix the CI failures.

client/config.c
sshedi: Check line 126, this is overwritten

solv/tdnfquery.c
sshedi: We enter inside this block based on same condition, so it's redundant.

llconf/nodes.c
sshedi: I changed tabs to spaces in this file. Hope that's okay.

oliverkurth:
Looks good, thank you!

PR #477, opened 2024-06-11
implement 'mirrorlist' for repos

oliverkurth:
Fetch mirror list if set. Variables in the `mirrorlist` link will be expanded, just like for `baseurl`. The downloaded list file will be reused, but expires after the same time as meta data expires. URLs will be tried in order - same behavior as if all URLs were set directly with `baseurl`.

### DIFF
--- a/client/repo.c
+++ b/client/repo.c
@@ -538,7 +538,7 @@ TDNFGetRepoMD(
     char *pszRepoMDUrl = NULL;
     char *pszTmpRepoDataDir = NULL;
     char *pszTmpRepoMDFile = NULL;
-    char *pszBaseUrlFile = NULL;
+    char *pszMirrorFile = NULL;
     char *pszTempBaseUrlFile = NULL;
     char* pszLastRefreshMarker = NULL;
     PTDNF_REPO_METADATA pRepoMDRel = NULL;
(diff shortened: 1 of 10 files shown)


client/repo.c
```diff
@@ -555,6 +555,48 @@ TDNFGetRepoMD(
...
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        if (stat(pszMirrorFile, &st) < 0) {
+            if (errno == ENOENT)
+                needDownload = 1;
+            else {
+                dwError = errno;
+                BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+            }
+        } else if ((now - st.st_ctime) > pRepoData->lMetadataExpire)
```
sshedi, line 587: what happens if system time is wrong?

client/repo.c
```diff
@@ -555,6 +555,48 @@ TDNFGetRepoMD(
...
+            BAIL_ON_TDNF_ERROR(dwError);
+        }
+
+        dwError = TDNFReadFileToStringArray(pszMirrorFile, &pRepoData->ppszBaseUrls);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        /* remove comments from mirror file */
+        int i, j;
+        for(i = 0, j = 0; pRepoData->ppszBaseUrls[i]; i++) {
+            if (pRepoData->ppszBaseUrls[i][0] == '#')
```
sshedi, line 601: use strspn here and ignore empty lines, `#` preceded by spaces or tabs
or is it omitted at the time of populating this buffer?

client/repo.c
```diff
@@ -555,6 +555,48 @@ TDNFGetRepoMD(
...
+                continue;
+            if (i != j) {
```
sshedi, line 603: I think this can be done:
```
if (i != j) {
    pRepoData->ppszBaseUrls[j] = pRepoData->ppszBaseUrls[i];
}
j++;
```

client/repolist.c
```diff
@@ -507,6 +507,10 @@ TDNFLoadReposFromFile(
             {
                 pRepo->pszMetaLink = strdup(cn->value);
             }
+            else if (strcmp(cn->name, TDNF_REPO_KEY_MIRRORLIST) == 0)
+            {
+                pRepo->pszMirrorList = strdup(cn->value);
```
sshedi, line 556: validate for NULL return

common/config.h
```diff
@@ -70,6 +71,7 @@
 
 //file names
 #define TDNF_REPO_METADATA_MARKER         "lastrefresh"
+#define TDNF_REPO_METADATA_MIRRORLIST     "mirrorlist"
```
sshedi, line 76: is it okay to do:
`#define TDNF_REPO_METADATA_MIRRORLIST TDNF_REPO_KEY_MIRRORLIST`
and keep them together?

sshedi:
Are you planning to add test for this?

client/repo.c
oliverkurth: Nothing we can do about that, we don't have any superpowers. This is the same as for repo data expiry.

client/repo.c
oliverkurth: I need to check if that is allowed in mirror files.

common/config.h
oliverkurth: These are two different things, they just happen to have the same value. One is a config setting key, the other is a file name. They could be different.

client/repo.c
oliverkurth: I thought about this, I think I leave it as is. I wasn't able to find out how dnf parses the mirror file, but I will assume that a comment has to start with `#` at the beginning of a line.

common/config.h
sshedi: In https://github.com/vmware/tdnf/pull/479 you have done the opposite of this :)

pytests/tests/test_mirrorlist.py
```diff
@@ -0,0 +1,84 @@
...
+    assert ret['retval'] == 0
+
+    pkgname = utils.config["mulversion_pkgname"]
+    utils.erase_package(pkgname)
+    ret = utils.run(['tdnf',
+                     '-y', '--nogpgcheck',
+                     '--disablerepo=*', '--enablerepo={}'.format(reponame),
+                     'install', pkgname],
+                    cwd=workdir)
+    assert utils.check_package(pkgname)
```
sshedi, line 84: please remove the created temp repo files and other stuff post testing

common/config.h
oliverkurth: It's not the same pattern. There we have an option common in both main config and repos, therefore they should have the same value. Here, we could just as well have a different value, like "mirror.list" for the file name.

pytests/tests/test_mirrorlist.py
oliverkurth: Files are removed, see lines 25ff , or am I missing something?

PR #479, opened 2024-06-12
make sslverify an option in main, setting the default for repos

oliverkurth:
Adds the `sslverify` option to the main config, setting the default for repos. This is to avoid having to set this for a big number of repositories indvidually.

The same principle can be applied to many other options, but this was the most pressing issue.

`gpgcheck` behaves the same way, but there it implemented in another (more complex way) - this should be unified.

### DIFF
--- a/common/config.h
+++ b/common/config.h
@@ -37,6 +37,7 @@
 #define TDNF_CONF_KEY_NO_PLUGINS          "noplugins"
 #define TDNF_CONF_KEY_PLUGIN_PATH         "pluginpath"
 #define TDNF_CONF_KEY_PLUGIN_CONF_PATH    "pluginconfpath"
+#define TDNF_CONF_KEY_SSL_VERIFY          "sslverify"
 #define TDNF_PLUGIN_CONF_KEY_ENABLED      "enabled"
 #define TDNF_CONF_KEY_EXCLUDE             "excludepkgs"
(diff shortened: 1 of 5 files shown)


common/config.h
```diff
@@ -60,7 +61,7 @@
 #define TDNF_REPO_KEY_RETRIES             "retries"
 #define TDNF_REPO_KEY_MINRATE             "minrate"
 #define TDNF_REPO_KEY_THROTTLE            "throttle"
-#define TDNF_REPO_KEY_SSL_VERIFY          "sslverify"
+#define TDNF_REPO_KEY_SSL_VERIFY          TDNF_CONF_KEY_SSL_VERIFY
```
sshedi, line 64: please keep this next to TDNF_CONF_KEY_SSL_VERIFY

sshedi (commented):
please add a test for this

common/config.h
oliverkurth: No, this is a `TDNF_REPO_KEY_` macro, why should this be under `TDNF_CONF_KEY_` macros?

sshedi: because they are same, will be easy to read if they are together

oliverkurth: I think it would be confusing. I want main options together, and repo options together.

Maybe we could add a section for common options, in another change when we add more.

sshedi (approved):

PR #480, opened 2024-06-12
implement vars

oliverkurth:
Implement `vars`. These are variables set by creating files in a directory. The names of the variable are the file names, and their value the (stripped) content. By default, `tdnf` will search the directories `/etc/tdnf/vars /etc/dnf/vars /etc/yum/vars`, and this can be set with `varsdir` in the main configuration. Variables are expanded in `baseurl` and `metalink`. This is in addition to the variables`releasever` and `basearch`.

### DIFF
--- a/client/repo.c
+++ b/client/repo.c
@@ -46,6 +46,11 @@ TDNFInitRepo(
     BAIL_ON_TDNF_ERROR(dwError);
 
     if (pRepoData->nHasMetaData) {
+        dwError = TDNFUtilsMakeDirs(pszRepoDataDir);
+        if (dwError == ERROR_TDNF_ALREADY_EXISTS)
+        {
+            dwError = 0;
+        }
         dwError = TDNFGetRepoMD(pTdnf,
                                 pRepoData,
                                 pszRepoDataDir,
--- a/client/varsdir.c
+++ b/client/varsdir.c
@@ -1,0 +1,118 @@
+/*
+ * Copyright (C) 2024 Broadcom, Inc. All Rights Reserved.
+ *
+ * Licensed under the GNU Lesser General Public License v2.1 (the "License");
+ * you may not use this file except in compliance with the License. The terms
+ * of the License are located in the COPYING file of this distribution.
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <ctype.h>
+#include <dirent.h>
+#include <errno.h>
+
+#include "../llconf/nodes.h"
+
+struct cnfnode *parse_varsdirs(char *dirs[])
+{
+    struct cnfnode *cn_top = NULL;
+    DIR *fdir = NULL;
+    FILE *fptr = NULL;
+    int i;
+
+    cn_top = create_cnfnode("(root)");
+
+    for (i = 0; dirs[i]; i++) {
+        fdir = opendir(dirs[i]);
+        if (!fdir) {
+            if (errno == ENOENT)
+                /* ignore non-existing directories */
+                continue;
+            else
+                goto error;
+        }
+
+        struct dirent *dent;
+        while((dent = readdir(fdir))) {
+            char buf[256], path[256];
+            char *p;
+
+            p = dent->d_name;
+            /* skip disallowed filenames */
+            /* See https://dnf.readthedocs.io/en/latest/conf_ref.html#varfiles-label :
+               "Filenames may contain only alphanumeric characters and underscores and be in lowercase." */
+            while(*p && ((*p == '_') || isdigit(*p) || islower(*p)))
+                p++;
+            if (*p) {
+                continue;
+            }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-truncation"
+            snprintf(path, sizeof(path), "%s/%s", dirs[i], dent->d_name);
+#pragma GCC diagnostic pop
+            fptr = fopen(path, "rt");
+            if (fptr == NULL) {
+                goto error;
+            }
+            if (fgets(buf, sizeof(buf), fptr) == NULL) {
+                /* might be EOF which is fine */
+                if (errno)
+                    goto error;
+            }
+            /* strip trailing spaces and newlines */
+            p = buf + strlen(buf)-1;
+            while (p >= buf && isspace(*p)) p--;
+            p[1] = 0;
+
+            struct cnfnode *cn_var = create_cnfnode(dent->d_name);
+            cnfnode_setval(cn_var, buf);
+            append_node(cn_top, cn_var);
+
+            fclose(fptr);
+        }
+        closedir(fdir);
+    }
+    return(cn_top);
+
+error:
+    if (fdir) closedir(fdir);
+    if (fptr) fclose(fptr);
+    if (cn_top) destroy_cnftree(cn_top);
+    return NULL;
+}
+
+char *replace_vars(struct cnfnode *cn_vars, const char *source)
+{
+    char dest[256];
+    char *p, *d;
+    const char *q, *s;
+    struct cnfnode *cn_var;
+
+    d = dest;
+    s = source;
+    while (*s && d < dest + sizeof(dest)-1) {
+        char name[256];
+        while(*s && *s != '$' && (d < dest + sizeof(dest)-1))
+            *d++ = *s++;
+        if (!*s)
+            break;
+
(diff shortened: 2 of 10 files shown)


client/repo.c
```diff
@@ -46,6 +46,11 @@ TDNFInitRepo(
     BAIL_ON_TDNF_ERROR(dwError);
 
     if (pRepoData->nHasMetaData) {
+        dwError = TDNFUtilsMakeDirs(pszRepoDataDir);
+        if (dwError == ERROR_TDNF_ALREADY_EXISTS)
```
sshedi, line 50: can you please move this check to TDNFUtilsMakeDirs function itself?

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+
+    cn_top = create_cnfnode("(root)");
+
+    for (i = 0; dirs[i]; i++) {
+        fdir = opendir(dirs[i]);
+        if (!fdir) {
+            if (errno == ENOENT)
+                /* ignore non-existing directories */
+                continue;
+            else
```
sshedi, line 33: else is redundant here
or just use:
if (errno != ENOENT) goto error;

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+#include <string.h>
+#include <ctype.h>
+#include <dirent.h>
+#include <errno.h>
+
+#include "../llconf/nodes.h"
+
+struct cnfnode *parse_varsdirs(char *dirs[])
+{
+	struct cnfnode *cn_top = NULL;
```
sshedi: nit: indent

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        struct dirent *dent;
+        while((dent = readdir(fdir))) {
+            char buf[256], path[256];
+            char *p;
+
+            p = dent->d_name;
+            /* skip disallowed filenames */
+            while(*p && ((*p == '_') || isdigit(*p) || islower(*p))) p++;
```
sshedi: please move p++ to next line, it is confusing to read

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+                goto error;
+        }
+
+        struct dirent *dent;
+        while((dent = readdir(fdir))) {
+            char buf[256], path[256];
+            char *p;
+
+            p = dent->d_name;
+            /* skip disallowed filenames */
```
sshedi, line 43: can you please explain a bit on why this skipping needed?

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+            while(*p && ((*p == '_') || isdigit(*p) || islower(*p))) p++;
+            if (*p) {
+                continue;
+            }
+#pragma GCC diagnostic push
```
sshedi, line 51: Does this warning occur even if you use `char path[PATH_MAX]`?

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+#pragma GCC diagnostic ignored "-Wformat-truncation"
+            snprintf(path, sizeof(path), "%s/%s", dirs[i], dent->d_name);
+#pragma GCC diagnostic pop
+            fptr = fopen(path, "rt");
+            if (fptr == NULL) {
+                goto error;
+            }
+            if (fgets(buf, sizeof(buf), fptr) == NULL) {
```
sshedi, line 59: check errono as well before concluding this as error, fgets returns NULL for EOF as well.

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+
+            struct cnfnode *cn_var = create_cnfnode(dent->d_name);
+            cnfnode_setval(cn_var, buf);
+            append_node(cn_top, cn_var);
+
+            fclose(fptr);
+        }
+        closedir(fdir);
+    }
+    return(cn_top);
```
sshedi, line 77: nit: `return cn_top;`

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+char *replace_vars(struct cnfnode *cn_vars, const char *source)
+{
+    char dest[256];
+    char *p, *d;
+    const char *q, *s;
+    struct cnfnode *cn_var;
+
+    d = dest;
+    s = source;
+    while (*s && d < dest + sizeof(dest)-1) {
```
sshedi, line 95: dest + sizeof(dest)-1 -> store this in a variable, no need to calculate it every time

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+    return(cn_top);
+
+error:
+    if (fdir) closedir(fdir);
+    if (fptr) fclose(fptr);
+    if (cn_top) destroy_cnftree(cn_top);
+    return NULL;
+}
+
+char *replace_vars(struct cnfnode *cn_vars, const char *source)
```
sshedi, line 86: validate function args

client/varsdir.c
```diff
@@ -0,0 +1,113 @@
...
+    while (*s && d < dest + sizeof(dest)-1) {
+        char name[256];
+        while(*s && *s != '$' && (d < dest + sizeof(dest)-1))
+            *d++ = *s++;
+        if (!*s)
+            break;
+
+        s++;
+        p = name;
+        while(*s && (p < name + sizeof(name)-1) && (isdigit(*s) || islower(*s)))
```
sshedi, line 104: store name + sizeof(name)-1 in a variable

common/config.h
```diff
@@ -82,6 +83,7 @@
 // repo defaults
 #define TDNF_DEFAULT_REPO_LOCATION        "/etc/yum.repos.d"
 #define TDNF_DEFAULT_CACHE_LOCATION       "/var/cache/tdnf"
+#define TDNF_DEFAULT_VARS_DIRS            "/etc/tdnf/vars /etc/dnf/vars /etc/yum/vars"
```
sshedi, line 86: can we make this a comma separated list?

common/strings.c
```diff
@@ -119,46 +119,51 @@ TDNFStringSepCount(
...
     char ***pppszTokens
     )
 {
     uint32_t dwError = 0;
-    char *pszTok = NULL;
-    char *pszState = NULL;
     char **ppszToks = NULL;
-    size_t nCount = 0;
-    size_t nIndex = 0;
+    const char *p, *p0;
```
sshedi: reduce the scope of p0 variable and please try to use a better name

sshedi:
And please don't forget to add a test for this.

client/repo.c
oliverkurth: Yeah, I had the same thought.

client/varsdir.c
oliverkurth: That doesn't make sense.Three cases:
- we have an error: `fdir == NULL && errno != ENOENT` => goto error
- dir does not exist: `fdir == NULL && errno == ENOENT` => ignore and continue with next one
- dir does exist: `fdir != NULL` => proceed (read it and do stuff)

client/varsdir.c
oliverkurth: See https://dnf.readthedocs.io/en/latest/conf_ref.html#varfiles-label : "Filenames may contain only alphanumeric characters and underscores and be in lowercase. "

client/varsdir.c
oliverkurth: I think so, since `PATH_MAX` has the same value, and this is also the size of `dent->d_name`. So the result can always be greater. It's okay to truncate, since the path name would be too long otherwise.

client/varsdir.c
oliverkurth: Good point - can happen with an empty file.

client/varsdir.c
oliverkurth: I like this better because it makes it clear what ewe are checking. Calculation is left to the optimizer.

common/config.h
oliverkurth: No, lists are always space separated. Same as for `baseurl` and others.

client/varsdir.c
oliverkurth: a tab sneaked in there somehow...

common/strings.c
oliverkurth: Variable name is perfect.

oliverkurth:
> And please don't forget to add a test for this.

Done.

sshedi (approved):

PR #481, opened 2024-06-13
print error and bail if a cmd line argument is an empty string

oliverkurth:
Failures because of empty arguments are hard to debug, catch them early.

### DIFF
--- a/tools/cli/lib/parseargs.c
+++ b/tools/cli/lib/parseargs.c
@@ -264,6 +264,12 @@ TDNFCliParseArgs(
 
         while (optind < argc)
         {
+            if (argv[optind][0] == 0) {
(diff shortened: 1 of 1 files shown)


sshedi:
Changes LGTM. Can we also validate URLs?

```
root@ph5dev [ ~ ]# tdnf install -y https:/packages.vmware.com/photon/5.0/photon_updates_5.0_x86_64/x86_64/vsftpd-3.0.5-4.ph5.x86_64.rpm
Nothing to do.
```
(https:/ instead of https://)
And tdnf doesn't throw any error.

oliverkurth:
> Changes LGTM. Can we also validate URLs?

Interesting. I can look into it, but this is not related to this change. You can create  an issue if you want to keep track of it.

PR #489, opened 2024-07-28
remove broken code and ignore the size field in metalink data

oliverkurth:
When the `size` tag was present in metalink file, the metalink plugin would throw an error:
```
$ sudo tdnf --repoid=epel --releasever=9 makecache
Loaded plugin: tdnfmetalink
Loaded plugin: tdnfrepogpgcheck
epel                                     23689 100%
XML Parser Warning: size is invalid value: 

Unable to parse metalink, ERROR: code=1622
Error: TDNFMetalinkGetBaseURLs 1622
Plugin error: metalink plugin error: (null)

Error: Failed to synchronize cache for repo 'Extra Packages for Enterprise Linux 9 - aarch64'
```
The code that checks the `size` field was wrong and useless, so removing it, and adding the field to the test file.

Future enhancements can evaluate the `size` field (and others, like `verification`), but with this change `tdnf` will be able to use metalink files from EPEL.

### DIFF
--- a/plugins/metalink/utils.c
+++ b/plugins/metalink/utils.c
@@ -472,24 +472,8 @@ TDNFXmlParseData(
         elementInfo->dwError = TDNFParseFileTag(userData);
         BAIL_ON_TDNF_ERROR(elementInfo->dwError);
     }
-    else if(!strcmp(elementInfo->startElement, TAG_NAME_SIZE))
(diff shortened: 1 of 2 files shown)


oliverkurth:
Replaced with PR #491 , closing.

PR #491, opened 2024-07-29
fix metalink 'size' field parsing

oliverkurth:
Fix metalink `size` field parsing. Alternative to PR #489 .

### DIFF
--- a/plugins/metalink/utils.c
+++ b/plugins/metalink/utils.c
@@ -455,7 +455,6 @@ TDNFXmlParseData(
     )
 {
     struct MetalinkElementInfo* elementInfo = (struct MetalinkElementInfo*)userData;
-    char *size = NULL;
 
     if(!elementInfo || !elementInfo->ml_ctx || IsNullOrEmptyString(elementInfo->filename) || elementInfo->dwError)
     {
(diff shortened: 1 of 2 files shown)


plugins/metalink/utils.c
```diff
@@ -467,28 +466,30 @@ TDNFXmlParseData(
...
     if(!strcmp(elementInfo->startElement, TAG_NAME_FILE))
     {
         elementInfo->dwError = TDNFParseFileTag(userData);
         BAIL_ON_TDNF_ERROR(elementInfo->dwError);
     }
     else if(!strcmp(elementInfo->startElement, TAG_NAME_SIZE))
     {
-        //Get File Size.
-        TDNFAllocateStringN(val, len, &size);
+        char size_buf[12], *p = size_buf;
```
tapakund, line 480: could you explain why did we hardcode the size_buf to 12 bytes?

oliverkurth: It's more than enough to hold a 32 bit integer, which in turn is way more than needed for the file size of `repomd.xml`.

plugins/metalink/utils.c
```diff
@@ -467,28 +466,30 @@ TDNFXmlParseData(
...
+        const char *q = val;
 
-        if(!size)
+        while (*q && q < val + len && p < size_buf + sizeof(size_buf) && isdigit(*q))
```
oliverkurth: This should actually be `p < size_buf + sizeof(size_buf) - 1`. I'll fix it later.

oliverkurth: Done.

PR #496, opened 2024-09-20
Fix small cache test by returning original cache path

sameluch:
Small test fix discovered while making the tests compatible with Azure Linux. This fixes out of space errors on later tests which are incompatible with the small_cache_path if this test is run.


oliverkurth:
Thanks again!

PR #497, opened 2024-09-20
Add virtual repo snapshot feature to tdnf

sameluch:
This PR adds some functionality built for Azure Linux to tdnf for the broader userbase, virtual repo snapshots. This is the ability to emulate an older snapshot of all enabled repos at a specific timestamp.

Added with this PR is a new config option `snapshottime=<posix_time>` which can be added to tdnf.conf to load as a global version of another added command line option `--snapshottime=<posix_time>`. The snapshot emulates as if the command was run against repos at the given timestamp. All packages which are published to a repo with a timestamp later than the the specified `--snapshottime` will be occluded from the solver and thus not considered for list, install, upgrade, and other commands which hit repositories.

Additionally, if some repos are desired to be not snapshot while executing a given path `--snapshotexcluderepos=<repo_id1>[,<repo_id2>,...]`. Each repo specified in `--snapshotexcluderepos` will NOT have the `--snapshottime` applied to it, if applicable.


oliverkurth:
Wow, that's a big change. It will take me a while to review this. I have a few question:

What is the use case? Is it just for testing, for example test an upgrade scenario? Or do you want to tag a specific time for a stable build?

Have you considered alternative approaches? For example, wouldn't it be easier to just create a "real" repository with all packages older than the snapshot time (if space is an issue, maybe just use symlinks?). I think this could be done using simple shell scripts, much simpler than this change.

How reliable is the "file" attribute, which is set (I guess) by `createrepo`? Does it use the actual file time stamp, or current time?

Have you thought about implementing this as a plugin? I know the plugin interface would probably need changes for this. But I think the use case is rather limited.

client/repo.c
```diff
@@ -24,13 +24,30 @@ TDNFInitRepo(
...
+    // set local POSIX limit if conf or cmd line opt is present
+    if (pTdnf->pConf != NULL && pTdnf->pConf->pszSnapshotTime!= NULL) 
+    {
+        pszSnapshotTime = pTdnf->pConf->pszSnapshotTime;
+    }
+
+    // take command line over config if both are present
+    for (pSetOpt = pTdnf->pArgs->pSetOpt; pSetOpt; pSetOpt = pSetOpt->pNext) 
+    {
+        if(strncmp(pSetOpt->pszOptName, TDNF_CONF_KEY_SNAPSHOT_TIME, strlen(TDNF_CONF_KEY_SNAPSHOT_TIME)) == 0)
```
sshedi, line 45: store `strlen(TDNF_CONF_KEY_SNAPSHOT_TIME)` in a variable before entering loop

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+    {
+        // print end element to file
+        dwError = printElementEndToFile(pTracking->pbOutfile, name);
+        if (dwError)
+        {
+            BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+        }
+
+    } 
+    else if (pTracking->nInPackage && !pTracking->nTimeFound)
```
sshedi, line 506: Checking pTracking->nInPackage for non zero is redundant here.

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+    {
+        // print to file
+        fprintf(pTracking->pbOutfile, "%s", pszLintedCharData);
+        if (ferror(pTracking->pbOutfile))
+        {
+            dwError = ERROR_TDNF_TIME_FILTER_IO;
+            BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+        }
+    }
+    else if (pTracking->nInPackage && !pTracking->nTimeFound)
```
sshedi, line 584: Checking pTracking->nInPackage for non zero is redundant here.

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+        *nElementBufferMaxLen = TDNF_DEFAULT_TIME_FILTER_BUFF_SIZE;
+    }
+    *pszElementBuffer = malloc(*nElementBufferMaxLen * sizeof(char));
+    
+    dwError = checkAndResizeBuffer(pszElementBuffer, nElementBufferMaxLen, strlen(pszElementName) + 4);
+    if (dwError)
+    {
+        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+    }
+    sprintf(*pszElementBuffer, "</%s>", pszElementName);
```
sshedi, line 243: use snprintf

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+
+    if (pszStringToEscape == NULL || ppszDestStr == NULL)
+    {
+        dwError = ERROR_TDNF_TIME_FILTER_GENERAL;
+        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+    }
+
+    // allocate new string for linted string
+    int nStrToLintLen = (strlen(pszStringToEscape) + 1); // add one for null char
+    char * pszLintedStr = malloc(nStrToLintLen * sizeof(char));
```
sshedi, line 82: Use `TDNFAllocateString`

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
 
+// #### XML FILTER CODE ####
+
+/***
+* Resize the buffer specified by ppszCharBuffer and update pnBufferMaxLen
+* to the length of the newly resized buffer if the nLengthToAdd would overflow
+* the buffer.
+***/
+uint32_t 
+checkAndResizeBuffer(char ** ppszCharBuffer, int * pnBufferMaxLen, int nLengthToAdd) {
```
sshedi, line 20: maybe make this a static function?

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+    goto cleanup;
+}
+
+/***
+* allocate a new string in ppszDestStr location with the linted description,
+* all '&', '<', and '>' characters will be replaced with the xml escape
+* character versions of each in line.
+***/
+uint32_t
+xmlEscapeCharLinter(const char * pszStringToEscape, char ** ppszDestStr) {
```
sshedi, line 68: make this a static function

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+        {
+            case '&':
+                pszCharToAdd = amp;
+                break;
+            case '>':
+                pszCharToAdd = gt;
+                break;
+            case '<':
+                pszCharToAdd = lt;
+                break;
```
sshedi, line 108: add a default case

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+        }
+
+        // add linted char
+        if (pszCharToAdd == NULL)
+        {
+            pszLintedStr[i + nOffset] = pszStringToEscape[i];
+        }
+        else
+        {
+            strcat(pszLintedStr, pszCharToAdd);
```
sshedi, line 129: keep track of pszLintedStr index and use it directly. No need to call strcat everytime

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+}
+
+/***
+* allocate a new buffer to location pszElementBuffer of the size
+* nElementBufferMax or greater (in the case resizing is needed).
+* a formatted start element with the name and attrs specified will be
+* placed in the newly allocated buffer.
+***/
+uint32_t 
+addElementStartToBuffer(char ** pszElementBuffer, int * nElementBufferMax, const char * pszElementName, const char ** ppszAttrs) {
```
sshedi, line 151: make this static function

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+    *pszElementBuffer = malloc(*nElementBufferMax * sizeof(char));
+    
+    char * pszLintedAttrVal = NULL;
+    char * pszTempBuffer = NULL;
+    dwError = checkAndResizeBuffer(pszElementBuffer, nElementBufferMax, strlen(pszElementName) + 2);
+    if (dwError)
+    {
+        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+    }
+    sprintf(*pszElementBuffer, "<%s", pszElementName);
```
sshedi, line 174: use snprintf

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+    if (nTempMaxLen >= TDNF_MAX_FILTER_INPUT_THRESHOLD)
+    {
+        dwError = ERROR_TDNF_TIME_FILTER_GENERAL;
+        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+    }
+
+    // only realloc if the size changed
+    if (nTempMaxLen != *pnBufferMaxLen)
+    {
+        pszTempCharBuffer = realloc(*ppszCharBuffer, nTempMaxLen);
```
sshedi, line 44: Try to make use of `TDNFReAllocateMemory`

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+        dwError = ERROR_TDNF_TIME_FILTER_GENERAL;
+        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+    }
+
+    // set default buffer max length
+    if (*nElementBufferMax == 0)
+    {
+        *nElementBufferMax = TDNF_DEFAULT_TIME_FILTER_BUFF_SIZE;
+    }
+    *pszElementBuffer = malloc(*nElementBufferMax * sizeof(char));
```
sshedi, line 165: use `TDNFAllocateMemory` or `TDNFAllocateString`

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+        // free temp variables
+        free(pszTempBuffer);
+        pszTempBuffer = NULL;
+        free(pszLintedAttrVal);
+        pszLintedAttrVal = NULL;
+    }
+    strcat(*pszElementBuffer, ">");
+    
+cleanup:
+    if (pszLintedAttrVal)
```
sshedi, line 206: no need of this if, free handles this now

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+ * 
+ ***/
+void 
+TDNFFilterStartElement(void *userData, const char * name, const char ** attrs) {
+    uint32_t dwError = 0;
+    char * pszStartElementBuffer = NULL;
+    // load tracking data
+    TDNFFilterData * pTracking = (TDNFFilterData *)userData;
+    int nAddNewLineAfterStart = pTracking->nPrevElement == 0;
+    char szNewLineBuffer[2];
```
sshedi, line 338: `char szNewLineBuffer[2] = {};` this will zero out the whole array

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+    {
+        bzero(szNewLineBuffer, sizeof(szNewLineBuffer)); // don't assume memory zero'd
+    }
+
+    // increment depth
+    pTracking->nDepth += 1;
+    pTracking->nPrevElement = 0;
+
+    // new package to parse or currently parsing package info
+    if (strcmp(name, "package") == 0 || pTracking->nInPackage)
```
sshedi, line 353: change the order of this condition check, this might save a strcmp call

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+ ***/
+void 
+TDNFFilterEndElement(void * userData, const char * name) {
+    uint32_t dwError = 0;
+    char * pszElementBuffer = NULL;
+    // load tracking data
+    TDNFFilterData * pTracking = (TDNFFilterData *)userData;
+
+    // decrement depth
+    pTracking->nDepth -= 1;
```
sshedi, line 493: `pTracking->nDepth--;` makes i easy to read

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+
+    int nEndElementBufferLength = TDNF_DEFAULT_TIME_FILTER_BUFF_SIZE;
+    char * pszEndElement = NULL;
+
+    dwError = addElementEndToBuffer(&pszEndElement, &nEndElementBufferLength, pszElementName);
+    if (dwError)
+    {
+        BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+    }
+    fprintf(pbOutfile, "%s", pszEndElement);
```
sshedi, line 310: have a wrapper function for this, repeated in many places

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+    goto cleanup;
+}
+
+/***
+ * 
+ ***/
+uint32_t 
+printElementEndToFile(FILE * pbOutfile, const char * pszElementName) {
+    uint32_t dwError = 0;
+    if (pbOutfile == NULL)
```
sshedi, line 296: check using IsNullOrEmptyString

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+        // add end element to buffer
+        dwError = addElementEndToBuffer(&pszElementBuffer, &nEndElementBufferLen, name);
+        if (dwError)
+        {
+            BAIL_ON_TDNF_TIME_FILTER_ERROR(dwError);
+        }
+        int nEndElementLen = strlen(pszElementBuffer);
+
+        dwError = checkAndResizeBuffer(&(pTracking->pszElementBuffer), &(pTracking->nBufferMaxLen), nEndElementLen);
+        if (dwError)
```
sshedi, line 520: why this check? BAIL_ON_TDNF_TIME_FILTER_ERROR does it already

solv/tdnfrepo.c
```diff
@@ -9,6 +9,766 @@
...
+
+/***
+ * 
+ ***/
+char * 
+TDNFFilterFile(const char * pszInFilePath, const char * pszSnapshotTime) {
+    // vars
+    uint32_t dwError = 0;
+    TDNFFilterData pData;
+    bzero(&pData, sizeof(TDNFFilterData));
```
sshedi, line 618: please don't mix declarations and function calls

sshedi (changes requested):
I just did one round of review. Need to understand the logic and usage better.

Some generic comments:
- Avoid sprintf, strcpy and prefer snprintf, strncpy
- Use already available tdnf APIs for memory allocations
- Make functions static if not used outside of file
- Add test cases and try to cover as various scenarios

Please sign the SLA before anything else.

oliverkurth:
> I just did one round of review. Need to understand the logic and usage better.

Thank you @sshedi . But let's not waste our time until @sameluch has answered my questions. I think there are better ways to achieve the same goal.

sameluch:
> > I just did one round of review. Need to understand the logic and usage better.
> 
> Thank you @sshedi . But let's not waste our time until @sameluch has answered my questions. I think there are better ways to achieve the same goal.

Hey @oliverkurth, I'll have some answers coming in soon. I was out of town with little internet access so, but back now to work through this larger change. I will sign the CLA as soon as I am able/allowed.
My apologies for opening a large change request and disappearing haha

@sshedi Thanks for the review I will consider the changes as we decide to move forward here and for future development within the tdnf repo.

sameluch:
> Wow, that's a big change. It will take me a while to review this. I have a few question:
> 
> What is the use case? Is it just for testing, for example test an upgrade scenario? Or do you want to tag a specific time for a stable build?
> 
> Have you considered alternative approaches? For example, wouldn't it be easier to just create a "real" repository with all packages older than the snapshot time (if space is an issue, maybe just use symlinks?). I think this could be done using simple shell scripts, much simpler than this change.
> 
> How reliable is the "file" attribute, which is set (I guess) by `createrepo`? Does it use the actual file time stamp, or current time?
> 
> Have you thought about implementing this as a plugin? I know the plugin interface would probably need changes for this. But I think the use case is rather limited.

@oliverkurth here's some answers for your above questions. Let me know if anything is not particularly clear

Use Case
The main focus of this feature is system stability from the client side without much additional overhead of space or calculation. It gives the option to system maintainers and users to "freeze" their system at a point in time for the purposes of stability. Additionally, should updates ever be required to the snapshot, you can simply just update the timestamp in tdnf.conf or use a new time on --snapshottime within a script or otherwise. The goal is client side customization of a repository snapshot without the maintenance overhead of creation/recreation of a manual repo each time.
This could have many uses such as, A-B testing and image recreation as couple cases the top of my head.

Alternate Approaches I've Considered
- Adding a similar option/function to libsolv which is currently used to read in repository and for package dependency calculation. Though adding this filtering directly to the package manager as an option seemed much more apt for the actual implementation overhead.
- Creation of individual repos for a snapshot given a desired time, this requires much more maintenance by the user or system maintainer. At least for our purposes, this did not seem like the right approach at this time.

The "file" attribute is added by `createrepo` and is present within each `<hash>-primary.xml` file. The time stamps used are those from the files, both the "build" and "file" times are collected this way. The "file" time is when the file was added to the repo, the "build" time is when the file in question was actually generated or built. This has been very reliable across remote repos and my local testing.

I did not consider a plugin implementation of this. Do you have an idea of what the changes might look like in that form? I don't quite see how a plugin interface would fit into this solution?
The actual implementation details of the current feature have the enabled, targeted repo files being temporarily copied and altered such that the data being fed to libsolv are those of the effective timestamp being used to leverage the solver in an intuitive way without the requirement of individual repo management.

oliverkurth:
Thanks for replying, @sameluch .

I am not convinced that other approaches are more overhead. Here are some ideas:
* creating a separate repo could be simply done with python scripts (or golang if you prefer), which go through the metadata and select just packages with older timestamps. If local, this could be done with symlinks. No impact to tdnf.
* after downloading the primary metadata (`*-primary.xml.gz`), filter it for the timestamps, and replace the original one. This could be done in one of multiple possible ways, for example a plugin, or possibly with a separate script, without impacting tdnf code, or just minimally.
* if this needs to be in tdnf itself, instead of rewriting the metadata file as this change does (IIUC), you can use the `considered` map of the libsolv package pool to filter out packages (looks like you still need to parse the xml separately because apparently the `file` timestamp is not exposed by libsolv, although the `build` time is, see `SOLVABLE_BUILDTIME` in https://github.com/openSUSE/libsolv/blob/master/doc/libsolv-constantids.txt). The `considered` map is used for example (but not only) for the `excludes` feature.

> The "file" time is when the file was added to the repo, 

What does "added" mean? Which time does `createrepo` use here? Its current time, or the file modification time? Something else? Can you point me to the code in `createrepo` that does this?

oliverkurth:
> It gives the option to system maintainers and users to "freeze" their system at a point in time for the purposes of stability.

What you can do is dump the set of installed packages, to json, like:
```
root [ / ]# tdnf -j list --installed | jq | head
[
  {
    "Name": "bash",
    "Arch": "aarch64",
    "Evr": "5.2-2.ph5",
    "Repo": "@System"
  },
  {
...
```

Then use a script, and install on the target with `tdnf install bash=5.2-2.ph5 ...` etc. So instead of a timestamp, which is pretty unreliable, you use a set of versions.

oliverkurth:
Closing this. https://github.com/vmware/tdnf/pull/518 should address the feature.

PR #502, opened 2024-10-10
Remove pthread_mutex API usage from TDNF.

sshedi:
As a single-threaded application, mutexes are unnecessary.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -10,7 +10,7 @@
 
 uid_t gEuid;
 
-static TDNF_ENV gEnv = {0};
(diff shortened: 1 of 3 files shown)


oliverkurth:
Looks good to me. Why was that there in the first place?

sshedi:
No idea. Totally unnecessary.

oliverkurth:
It's very old, from 2016. It may have made sense if this code was used as a library by a 3rd party application (like old pmd), but we don't have that.

PR #506, opened 2024-12-02
CI/CD fixes

sshedi:
I started this work with an idea to segregate package list in docker files but found some other issues along the way.

- We were using Fedora-34 for testing, so switched to using latest (F41)
- When I switched to F41, lot of spec issues popped up and all were correct.
- Ended up fixing many spec files in test repo
- No need to pass both rootdir option and topdir option to rpmbuild, unless we are doing something in chroot and this caused regressions in F41 while build rpms from specs
- F41 is not happy with `default` Key-Curve for generating gpg keys (throws unknown elliptic curve error), so switched to RSA


ci/docker-entrypoint.sh
```diff
@@ -1,14 +1,23 @@
...
+
 set -e
 
 rm -rf build
 mkdir -p build
 cd build || exit 1
 
-mkdir -p /usr/lib/sysimage/tdnf
-cmake -DHISTORY_DB_DIR=/usr/lib/sysimage/tdnf .. && make -j32 && make python -j32 && make check -j32 || exit
+JOBS=$(( $(nproc) / 2 ))
```
oliverkurth: Why 50% of CPUs? Why not all?

pytests/tests/test_cache.py
```diff
@@ -235,7 +234,8 @@ def test_cache_directory_out_of_disk_space(utils):
 
     switch_cache_path(utils, orig_cache_path)
     clean_cache(utils)
-    clean_small_cache(utils)
+    utils.run(f"rm -rf {small_cache_path}/*".split())
```
oliverkurth: Please use `shutil.rmtree`. Try to avoid shell commands in python.

pytests/tests/test_cache.py
```diff
@@ -218,15 +214,18 @@ def test_download_vs_cache_size_multiple_packages(utils):
...
+    if try_mount_small_cache():
+        print("Failed to mount small cache directory")
+        return
+
     small_cache_path = utils.config['small_cache_path']
     orig_cache_path = utils.tdnf_config.get('main', 'cachedir')
     switch_cache_path(utils, small_cache_path)
     enable_cache(utils)
-    clean_small_cache(utils)
+    utils.run(f"rm -rf {small_cache_path}/*".split())
```
oliverkurth: Please use `shutil.rmtree`. Try to avoid shell commands in python.

.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
             - name: run container with tests
               working-directory: ${{ github.workspace }}
               run: |
-                docker run --security-opt seccomp:unconfined --rm -e DIST -v$(pwd):/build -w/build ${DIST}/tdnf-build ./ci/docker-entrypoint.sh
+                docker run --privileged --security-opt seccomp:unconfined --rm -e DIST -v$(pwd):/build -w/build ${DIST}/tdnf-build ./ci/docker-entrypoint.sh
```
oliverkurth, line 19: Why do we need `--privileged` now?

ci/Dockerfile.fedora
```diff
@@ -1,15 +1,7 @@
...
-        python3-pyOpenSSL python3 python3-devel valgrind gpgme-devel \
-        expat-devel openssl-devel sqlite-devel rpm-sign which python3-pip \
-        shadow-utils sudo e2fsprogs util-linux
-
-RUN pip3 install flake8
-
-RUN mkdir -p /var/lib/tdnf
+COPY ci/prep.sh /prep.sh
+RUN chmod +x /prep.sh
+RUN /prep.sh
```
oliverkurth, line 5: IMHO, I think this makes it more complex than needed.

oliverkurth (commented):
Thank you for this! It looks mostly good to me, but I have a few comments.

I always dislikes the `_topdir` in the spec files, but never had the energy to fix it. Thanks!

ci/docker-entrypoint.sh
sshedi: Good practice. Better to leave half the cores for other components usage.

pytests/tests/test_cache.py
sshedi: I added a helper function for this, `shutil.rmtree` deletes the directory, not just the content within. Here `small_cache_path` is a mount point, so it will fail.

ci/Dockerfile.fedora
sshedi: I understand. But easy to maintain going forward and if we are trying to test tdnf in a host (not in container), we can directly run this script and get all dependencies.

.github/workflows/ci.yml
sshedi: small_cache test tries to mount a disk within container, mount won't work if the container is not privileged.

ci/prep.sh
```diff
@@ -0,0 +1,63 @@
...
+    curl-devel
+    glib
+    glibc-debuginfo
+    shadow
+    zlib-devel
+  )
+  tdnf -y upgrade --refresh
+  tdnf remove -y toybox
+  tdnf -y install --enablerepo=photon-debuginfo ${photon_packages[@]}
+  pip3 install pytest flake8
```
oliverkurth: We have flake in Photon:
```
okurth@photon-dev [ ~/poi-harness ]$  (main)$ tdnf search flake
python3-pyflakes : A simple program which checks Python source files for errors
okurth@photon-dev [ ~/poi-harness ]$  (main)$ tdnf repoquery --list python3-pyflakes
/usr/bin/pyflakes
```
Can we use that?

sshedi: They are not the same. From the pyflakes project page: https://github.com/PyCQA/pyflakes
```
If you require more options and more flexibility, you could give a look to [Flake8](https://pypi.org/project/flake8/) too.
```

ci/docker-entrypoint.sh
oliverkurth: It will fail if we have just CPU:
```
$ echo $(( 1 / 2 ))
0
```
And `make` does not like it:
```
# make -j0
make: the '-j' option requires a positive integer argument
```

oliverkurth: Using `$(( ($(nproc)+1) / 2 ))` should fix it.

sshedi: Correct, good catch. I thought of this scenario once and forgot to make the changes. Thanks.

PR #508, opened 2024-12-04
Fix updateinfo script

sshedi:
Move instructions from bin/tdnf-cache-updateinfo to 02-tdnf-updateinfo, when motd triggers it, it will generate the right data.


etc/motdgen.d/02-tdnf-updateinfo.sh
oliverkurth, line 1: nit: change the copyright year

oliverkurth, line 1: How does this file get installed? I see you removed the old name from the install script and the spec file, why is that not replaced with the new name?

sshedi: This file is not removed from getting packaged in the spec
I removed `tdnf-cache-update-info` script.

There was a timer, a service file for populating updateinfo.
When the timer went off, it was triggering the service and that in turn tdnf-cache-update-info and that in turn was triggering 02-tdnf-updateinfo.sh.

Timer -> Updateinfo service -> /bin/tdnf-cache-update-info -> /etc/motgend.d/02-tdnf-updateinfo.sh

This is too much for such a trivial task.

PR #518, opened 2025-04-25
implement snapshot option for repos

oliverkurth:
Add `snaspshot` option for repositories. It needs to be set to a file which contains a list of packages which will be accepted for the repository. Other packages in that repository will be disabled.

This is to implement snapshots of repositories that contain packages of different versions that are continuously updated by adding new packages. A snapshot file should list the current packages at the time of an update.

The snapshot file needs to be a list of package names wity versions, in the form `<name>=<evr>`, like `pkg-name=1.2-4.ph5`.

Example to create a snapshot of today, assuming `photon-updates` contains just the latest current versions:

With `tdnf` >= 3.6.0
```
tdnf repoquery --repoid=photon-updates --available --qf "%{name}=%{evr}" > snapshot-20250425.list
```
or
```
tdnf list --repoid photon-updates --available -j | jq -r '.[] | [.Name, .Evr] | join("=")' > snapshot-20250425.list
```
Or create a list of all installed packages, for example to reproduce the current set of installed packages in another host:
```
tdnf repoquery --installed --qf "%{name}=%{evr}" > snapshot.list
```

This can be used later by configuring a repository, assuming the `photon` repo contains all packages of all versions:
```
[photon-20250425]
name = VMware Photon Linux $releasever ($basearch)
baseurl = https://packages.vmware.com/photon/$releasever/photon_$releasever_$basearch
gpgkey = file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY file:///etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY-4096
gpgcheck = 1
enabled = 0
skip_if_unavailable = 1
skip_md_filelists = 1
snapshot = /etc/yum.repos.d/snapshot-20250425.list
```

Note that the snapshot is *per repository*. Packages in other repositories will still be installable, regardless if they are in the list or not. Therefore, to use the repository above, disable all other that may have other versions or use `--repoid=photon-20250425`.

### DIFF
--- a/solv/simplequery.c
+++ b/solv/simplequery.c
@@ -54,6 +54,26 @@ int split_nevra(char *nevra, char **name, char **evr, char **arch)
     return 0;
 }
 
+/* split string like "pkg-name=1.2.4-1.ph5"
+   into "pkg-name" and "1.2.4-1.ph5" 
+   There is no arch
+*/
+static
+int split_name_equals_evr(char *nevr, char **name, char **evr)
+{
+    char *p = nevr;
+    while(*p && *p != '=')
+        p++;
+    if (*p)
+        *p = 0;
+    else
+        return -1;
+    *evr = p+1;
+    *name = nevr;
+
+    return 0;
+}
+
 /* Find packages by nevra as specified with ids. Must be either installed
    or not as set by the 'installed' flag. Adds result to qresult, can
    be multiples if package is in multiple repos. */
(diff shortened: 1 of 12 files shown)


solv/simplequery.c
```diff
@@ -54,6 +54,26 @@ int split_nevra(char *nevra, char **name, char **evr, char **arch)
...
+   into "pkg-name" and "1.2.4-1.ph5" 
+   There is no arch
+*/
+static
+int split_name_equals_evr(char *nevr, char **name, char **evr)
+{
+    char *p = nevr;
+    while(*p && *p != '=')
+        p++;
+    if (*p)
```
sshedi, line 67: check p for NULL as well
```
while (p && *p && *p != '=')

if (!p || !*p)
    return -1;

*p = 0;
...
```

sshedi (approved):

solv/simplequery.c
oliverkurth: It's just a quick helper function. The parameter is already checked in the caller `SolvFindSolvablesByNEqualsEvrFromRepo()`.

PR #522, opened 2025-05-10
set rpm verify flag when gpgcheck enabled in tdnf.conf

yepmunchun:


yepmunchun:
When `gpgcheck=1` in `/etc/tdnf/tdnf.conf`, unsigned package still can pass through the installation's `Testing transaction`, and proceed to `Running transaction`.

image

This patch sets the rpm verify flag to enable signature verification when `gpgcheck` enabled in `tdnf.conf`. 

image

oliverkurth:
Thank you for this. This is a valid bug. However, the solution is more complex because we also have the flag `gpgcheck` in the repo configuration, which should override the global config, and all this should be overridable with the `--nogpgcheck` command line option. I am working on a more comprehensive solution.

oliverkurth:
Thanks for raising the issue, but I am going to close this in favor of PR #523 .

PR #523, opened 2025-05-14
do not install unsigned packages if gpgcheck is enabled

oliverkurth:
This started with addressing the issue in https://github.com/vmware/tdnf/pull/522 (unsigned packages were installed even when `gpgcheck` is enabled), and developed into a major overhaul of the signature checking code.

Changes made:
* make sure `gpgcheck` can be consistently enabled globally in `tdnf.conf` which sets the default, and on a per repo basis. Repo config inherits from the global config if unset in the repo config.
* make the `--nogpgcheck` disable the check globally
* refactor setting the rpm verify flags and level
* move signature check to the `TDNFGPGCheckPackage()` function used by `TDNFTransAddInstallPkg()` and away from `TDNFRunTransaction()` to ensure settings on per repo basis are used.
* move internally the `--skipdigest` and `--skipsignature` options to settings in the global config, to make code simpler
* add more tests

### DIFF
--- a/client/config.c
+++ b/client/config.c
@@ -301,6 +301,19 @@ TDNFReadConfig(
     dwError = TDNFConfigFromCnfTree(pConf, cn_conf->first_child);
     BAIL_ON_TDNF_ERROR(dwError);
 
+    /* override from cmd line */
+    if (pTdnf->pArgs->nNoGPGCheck) {
+        pConf->nGPGCheck = 0;
+    }
+
+    /* these two have no config setting (future?) */
+    if (pTdnf->pArgs->nSkipDigest) {
+        pConf->nSkipDigest = 1;
+    }
+    if (pTdnf->pArgs->nSkipSignature) {
+        pConf->nSkipSignature = 1;
+    }
+
     pszTdnfVersion = TDNFGetVersion();
 
     if (pConf->pszOSName == NULL)
--- a/client/gpgcheck.c
+++ b/client/gpgcheck.c
@@ -210,7 +210,6 @@ TDNFGPGCheckPackage(
 {
     uint32_t dwError = 0;
     Header rpmHeader = NULL;
-    int nGPGSigCheck = 0;
     FD_t fp = NULL;
     char** ppszUrlGPGKeys = NULL;
     char* pszLocalGPGKey = NULL;
@@ -218,9 +217,34 @@ TDNFGPGCheckPackage(
     int nRemote = 0;
     int i;
     int nMatched = 0;
+    char *pszTmp = NULL;
+    int nSavedVfyLevel = 0;
+    rpmVSFlags savedVSFlags = 0;
 
-    dwError = TDNFGetGPGSignatureCheck(pTdnf, pRepo, &nGPGSigCheck, NULL);
-    BAIL_ON_TDNF_ERROR(dwError);
+    if(pTS == NULL || pTdnf == NULL || pRepo == NULL || IsNullOrEmptyString(pszFilePath))
+    {
+        dwError = ERROR_TDNF_INVALID_PARAMETER;
+        BAIL_ON_TDNF_ERROR(dwError);
+    }
+
+    nSavedVfyLevel = rpmtsVfyLevel(pTS->pTS);
+    savedVSFlags = rpmtsVSFlags(pTS->pTS);
+
+    if (pRepo->nGPGCheck)
+    {
+        int level = RPMSIG_VERIFIABLE_TYPE;
+        if (pTdnf->pConf->nSkipSignature) {
+            level &= ~RPMSIG_SIGNATURE_TYPE;
+            rpmtsSetVSFlags(pTS->pTS, rpmtsVSFlags(pTS->pTS) | RPMVSF_MASK_NOSIGNATURES);
+        }
+        if (pTdnf->pConf->nSkipDigest) {
+            level &= ~RPMSIG_DIGEST_TYPE;
+            rpmtsSetVSFlags(pTS->pTS, rpmtsVSFlags(pTS->pTS) | RPMVSF_MASK_NODIGESTS);
+        }
+        rpmtsSetVfyLevel(pTS->pTS, level);
+    } else {
+        rpmtsSetVfyLevel(pTS->pTS, RPMSIG_NONE_TYPE);
+    }
 
     fp = Fopen (pszFilePath, "r.ufdio");
     if(!fp)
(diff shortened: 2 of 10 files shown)


client/gpgcheck.c
```diff
@@ -234,10 +241,21 @@ TDNFGPGCheckPackage(
...
-
     Fclose(fp);
     fp = NULL;
 
+    if (nGPGSigCheck) {
+        /* refuse to install an unsigned package if gpgcheck is enabled */
+        if (((pszTmp = headerGetAsString(rpmHeader, RPMTAG_SIGPGP)) == NULL) &&
+            ((pszTmp = headerGetAsString(rpmHeader, RPMTAG_SIGGPG)) == NULL) &&
+            ((pszTmp = headerGetAsString(rpmHeader, RPMTAG_DSAHEADER)) == NULL) &&
+            ((pszTmp = headerGetAsString(rpmHeader, RPMTAG_RSAHEADER)) == NULL))
```
yepmunchun, line 269: Besides unsigned case, there could be signed but invalid signature, which cannot be detected by NULL header check.
In the beginning of each rpm transaction, `rpmtsClean` reset the rpm verify level to the default `_pkgverify_level` defined in `/usr/lib/rpm/macros`.

image

which is default to `digest` only.
to ensure the rpm validate the signature, we will need `rpmtsSetVfyLevel` with `RPMSIG_VERIFIABLE_TYPE` to enable the `signature` bit.

image

oliverkurth:
Thanks again - you are right. But it needs to be on per repository level. We run `rpmReadPackageFile()` in `TDNFGPGCheckPackage()` (in `gpgcheck.c`), that can check for the signature by setting the appropriate flags. We also have options to skip signature and digest checking. It's a bit messy, fixing this properly takes some work.

client/config.c
```diff
@@ -301,6 +301,19 @@ TDNFReadConfig(
     dwError = TDNFConfigFromCnfTree(pConf, cn_conf->first_child);
     BAIL_ON_TDNF_ERROR(dwError);
 
+    /* override from cmd line */
+    if (pTdnf->pArgs->nNoGPGCheck) {
+        pConf->nGPGCheck = 0;
+    }
+
+    /* these two have no config setting (future?) */
```
sshedi, line 309: Prefix this comment with `TODO:`; I think this can be done as a commit in this same PR

Also, why not do:
```
pConf->nGPGCheck = !(pTdnf->pArgs->nNoGPGCheck);
pConf->nSkipDigest = !!(pTdnf->pArgs->nSkipDigest);
pConf->nSkipSignature = !!(pTdnf->pArgs->nSkipSignature);
```

client/gpgcheck.c
```diff
@@ -210,17 +210,41 @@ TDNFGPGCheckPackage(
...
+    if (pRepo->nGPGCheck)
+    {
+        int level = RPMSIG_VERIFIABLE_TYPE;
+        if (pTdnf->pConf->nSkipSignature) {
+            level &= ~RPMSIG_SIGNATURE_TYPE;
+            rpmtsSetVSFlags(pTS->pTS, rpmtsVfyFlags(pTS->pTS) | RPMVSF_MASK_NOSIGNATURES);
+        }
+        if (pTdnf->pConf->nSkipDigest) {
+            level &= ~RPMSIG_DIGEST_TYPE;
+            rpmtsSetVSFlags(pTS->pTS, rpmtsVfyFlags(pTS->pTS) | RPMVSF_MASK_NODIGESTS);
```
sshedi: Replace `rpmtsVfyFlags(pTS->pTS)` with value from `savedVfyflags`, same in line 238

client/gpgcheck.c
```diff
@@ -234,17 +258,28 @@ TDNFGPGCheckPackage(
                   fp,
                   pszFilePath,
                   &rpmHeader);
-
     Fclose(fp);
     fp = NULL;
 
+    if (pRepo->nGPGCheck && !pTdnf->pConf->nSkipSignature) {
+        /* refuse to install an unsigned package if gpgcheck is enabled */
+        if (((pszTmp = headerGetAsString(rpmHeader, RPMTAG_SIGPGP)) == NULL) &&
```
sshedi, line 266: `pszTmp` is freed at the end, but it is assigned four times. This should result in 3 missing frees, right?

client/repolist.c
```diff
@@ -641,6 +641,11 @@ TDNFLoadReposFromFile(
         if (pRepo->pszName == NULL)
             pRepo->pszName = strdup(pRepo->pszId);
 
+        /* override from cmd line */
+        if (pTdnf->pArgs->nNoGPGCheck) {
+            pRepo->nGPGCheck = 0;
```
sshedi, line 646: nit: `pRepo->nGPGCheck = !(pTdnf->pArgs->nNoGPGCheck);`

pytests/tests/test_signature.py
```diff
@@ -26,19 +26,71 @@ def setup_test_function(utils):
...
+# expect success
+def test_install_unsigned_skipsignature(utils):
+    set_gpgcheck(utils, True, repo='photon-test-unsigned')
+    set_repo_key(utils, DEFAULT_KEY)
+    pkgname = utils.config["sglversion_pkgname"]
+    ret = utils.run(['tdnf', '--skipsignature', '--repoid', 'photon-test-unsigned', 'install', '-y', pkgname])
+    assert ret['retval'] == 0
+    assert utils.check_package(pkgname)
+
+
```
sshedi, line 93: Thanks for adding the tests. Neatly done. Can you please add one for `--skipdigest` as well? I didn't find it anywhere currently.

sshedi (changes requested):

client/gpgcheck.c
oliverkurth: No, because it's conditional. The condition stops to be evaluated when the first call to `headerGetAsString()` returns a non-NULL pointer, therefore we will have a string allocated once at most.

client/repolist.c
oliverkurth: That's not the same thing. We do not want to set `pRepo->nGPGCheck = 1` always when we do not have the option set. It may have been configured to be `0` already via config.

pytests/tests/test_signature.py
oliverkurth: Yes, good point.

client/gpgcheck.c
oliverkurth: No, we may have changed the flags above, if `nSkipSignature` is set.

client/config.c
oliverkurth: I can add `TODO`. Or d you mean adding the options to the config?

`pConf->nGPGCheck = !(pTdnf->pArgs->nNoGPGCheck);`
I already commented below - we do not necessarily want `gpgcheck` enabled when the option is not given. This would override any config setting.

```
pConf->nSkipDigest = !!(pTdnf->pArgs->nSkipDigest);
pConf->nSkipSignature = !!(pTdnf->pArgs->nSkipSignature);
```
Same argument, although we do not have those options in the config - but we may implement them.

sshedi: > Or d you mean adding the options to the config?

This, if it is not too much to have in one PR.

> I already commented below - we do not necessarily want gpgcheck enabled when the option is not given. This would override any config setting.

Ack.

client/gpgcheck.c
sshedi: Can you please explain a bit more? Am I missing something here?
https://github.com/rpm-software-management/rpm/blob/84170932172ba6a6e735169c398130b3b206bd82/lib/rpmts.cc#L625

https://github.com/rpm-software-management/rpm/blob/84170932172ba6a6e735169c398130b3b206bd82/lib/rpmts.cc#L615

`vsflags` and `vfyflags` are different here.

oliverkurth: Ah, good catch! It should be `rpmtsVSFlags()` instead if `rpmtsVfyFlags()`. Probably a copy/paste error.

oliverkurth: Fixed.

pytests/tests/test_signature.py
sshedi: Please address this, otherwise LGTM 👍

client/config.c
oliverkurth: Let's skip implementing `skipdigest`/`skipsigmature` as config options in this PR. It can be done in a different PR, if needed. The use case is very rare.

pytests/tests/test_signature.py
oliverkurth: I added a test for `--skipdigest`.

PR #529, opened 2025-06-04
Misc fixes

sshedi:
This patch series contains a bunch of fixes and improvements. Please check commit messages.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -11,8 +11,7 @@
 uid_t gEuid;
 
 static int gInitialized;
-
-static tdnflock instance_lock;
+static int instanceLockFd = -1;
 
 static void TdnfExitHandler(void);
 static void IsTdnfAlreadyRunning(void);
@@ -24,7 +23,7 @@ static void TdnfExitHandler(void)
         return;
     }
 
-    tdnflockFree(instance_lock);
+    tdnfLockFree(TDNF_INSTANCE_LOCK_FILE, instanceLockFd);
 }
 
 static void IsTdnfAlreadyRunning(void)
@@ -34,11 +33,10 @@ static void IsTdnfAlreadyRunning(void)
         return;
     }
 
-    instance_lock = tdnflockNewAcquire(TDNF_INSTANCE_LOCK_FILE,
-                                       "tdnf_instance");
-    if (!instance_lock)
+    instanceLockFd = tdnfLockAcquire(TDNF_INSTANCE_LOCK_FILE);
+    if (instanceLockFd < 0)
     {
-        pr_err("Failed to acquire tdnf_instance lock\n");
+        pr_err("Failed to acquire tdnfInstance lock\n");
     }
 }
 
--- a/client/api.c
+++ b/client/api.c
@@ -15,6 +15,7 @@ static int instanceLockFd = -1;
 
 static void TdnfExitHandler(void);
 static void IsTdnfAlreadyRunning(void);
+static void alterSpecStr(const char *pszSpec, char *buf);
 
 static void TdnfExitHandler(void)
 {
@@ -922,6 +923,25 @@ error:
     goto cleanup;
 }
 
+static void
+alterSpecStr(const char *pszSpec, char *buf)
+{
+    char *dst = buf;
+    const char *src = pszSpec;
+
+    *dst++ = '*';
+    if (*src != '/')
+        *dst++ = '/';
+
+    size_t len = strlen(src);
+    while (len && src[len - 1] == '/')
+      --len;
+
+    strncpy(dst, src, len);
+
+    dst[len] = '\0';
+}
+
 uint32_t
 TDNFProvides(
     PTDNF pTdnf,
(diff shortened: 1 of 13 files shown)


common/lock.c
```diff
@@ -8,254 +8,88 @@
...
-
-error:
-	tdnflock_free(lock);
-
-	return lock;
-}
+static int tdnfCreteLockFile(void);
 
-static void tdnflock_free(tdnflock lock)
+static int tdnfCreteLockFile(void)
```
oliverkurth: Typo, I guess you mean "tdnfCreateLockFile".

common/lock.c
```diff
@@ -8,254 +8,88 @@
 
 #include "includes.h"
 
-static void tdnflock_free(tdnflock lock);
-static void tdnflock_release(tdnflock lock);
-static int tdnflock_acquire(tdnflock lock, int mode);
-static tdnflock tdnflock_new(const char *lock_path, const char *descr);
+static int gLockFd;
+static const char *lockPath = TDNF_INSTANCE_LOCK_FILE;
```
oliverkurth: Does this ever change? If not, use `TDNF_INSTANCE_LOCK_FILE` directly.

common/lock.c
```diff
@@ -8,254 +8,88 @@
...
-		info.l_type = F_UNLCK;
-		info.l_whence = SEEK_SET;
-		info.l_start = 0;
-		info.l_len = 0;
-		info.l_pid = 0;
-		(void) fcntl(lock->fd, F_SETLK, &info);
-	}
+    int ret;
+
+    if (IsNullOrEmptyString(lockPath))
```
oliverkurth, line 52: If `lockPath` never changes, this condition will never be true.

common/lock.c
```diff
@@ -8,254 +8,88 @@
...
+    if (gLockFd >= 0)
+    {
+        if (flock(gLockFd, LOCK_UN))
+        {
+          pr_err("ERROR: failed to unlock: '%s'\n", lockPath);
+        }
+        (void) close(gLockFd);
+    }
+
+    if (!access(lockPath, F_OK) && remove(TDNF_INSTANCE_LOCK_FILE))
```
oliverkurth: Here you use the macro `TDNF_INSTANCE_LOCK_FILE` for `remove()` already, but in `access()` and above you use `lockPath`. No matter what you prefer (and I prefer the macro unless there is a chance it changes), make it consistent.

oliverkurth (changes requested):

client/api.c
```diff
@@ -920,6 +921,25 @@ TDNFAddCmdLinePackages(
     goto cleanup;
 }
 
+static void
```
oliverkurth, line 926: This function prepends a `*` always? So "foo/bar" becomes "*foo/bar"?  Why?

oliverkurth: I understand what you are trying to achieve here, but I think this needs more work. For example, you can use the `PATH` for executables. I think `dnf` already does something like this - find out how and maybe we can emulate that.

oliverkurth: I any case, add comments about what this is doing for future generations.

oliverkurth (changes requested):

tools/cli/lib/api.c
```diff
@@ -678,192 +684,166 @@ TDNFCliRepoSyncCommand(
...
+        {"license", pPkgInfo->pszLicense},
+        {"url", pPkgInfo->pszURL},
+        {NULL, NULL}
+    };
+
+    for (int i = 0; tagMappings[i].tag; ++i) {
+        if (!strcmp(tag, tagMappings[i].tag)) {
+            pszVal = tagMappings[i].member;
+            *pszRet = strdup(pszVal);
+            return dwError;
```
oliverkurth, line 721: Remove this line and the `strdup()` above - `pszVal` will be copied below (line 739) if set.

oliverkurth (changes requested):
I like the `format_string()` refactor - something I wanted to do but found no time for yet. Thank you!

tools/cli/lib/api.c
sshedi: They are different.  In line 729 we dynamically allocate memory to `pszVal` and here it is not a memory address fro heap.

common/lock.c
sshedi: My idea here is to make lock.c like a library which we should scale if we need multiple locks at some point.
Thanks for the inputs, I kinda rushed while making this modification, tried to do it right this time.

client/api.c
sshedi: dnf5 handles this in a slight complex way, https://github.com/rpm-software-management/dnf5/blob/main/dnf5/commands/provides/provides.cpp#L78

I think prepending `*` is easier and more user friendly way to achieve the same result. And users can simply do `tdnf whatprovides/provides grub` (for instance) and they get desired results. I don't like doing things on behalf of users, that's a recipe for failure at some point but I think this is one exemption. What say you?

I also improved the code slightly, if user has provided `*` as first character in the query, the string doesn't get modified.

oliverkurth: I don't like this. I think this can be really annoying when `tdnf` prints out tons of matches just because it always prepends a wildcard. I think we should do it the way dnf does it. We could even improve on it by using the `PATH` env var, and maybe do something similar for libraries.

PR #531, opened 2025-07-31
tdnf-config: fix -f/--file option for get/remove and other commands

oliverkurth:
The `-f` option was only used for the `create` command, and ignored for all other commands.

### DIFF
--- a/tools/config/main.c
+++ b/tools/config/main.c
@@ -24,6 +24,7 @@
 #define ERR_NO_SETTING  4
 #define ERR_REPO_EXISTS 5
 #define ERR_JSON        6
+#define ERR_INTERNAL    7
 
 #define pr_err(fmt, ...) \
     fprintf(stderr, fmt, ##__VA_ARGS__)
@@ -129,6 +130,7 @@ void remove_repo(struct cnfnode *cn_root, const char *repo)
     } else
         fail(ERR_NO_REPO, "repo '%s' not found\n", repo);
 }
+
 static
 char *get_repodir(const char *main_config)
 {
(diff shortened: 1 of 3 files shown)


tools/config/main.c
```diff
@@ -161,6 +163,8 @@ struct cnfnode *find_repo(const char *repodir, const char *repo, char **pfilenam
     glob_t globbuf =  {0};
     int i, rc = 0;
 
+    check_cond(strlen(repodir) + 7 < sizeof(pattern)-1);
+
     snprintf(pattern, sizeof(pattern), "%s/*.repo", repodir);
```
sshedi: You can check `snprintf` retval for errors instead of line 166.
`if (retval < 0 || retval >= sizeof(pattern)) { error; ...}`

tools/config/main.c
```diff
@@ -220,13 +224,18 @@ int write_file(struct cnfnode *cn_root, const char *filename)
     int rc = 0;
     char buf[256];
 
+    check_cond(strlen(filename) + 4 < sizeof(buf));
```
sshedi: same here

tools/config/main.c
```diff
@@ -257,6 +266,16 @@ struct json_dump *cnftree2json(struct cnfnode *cn_root)
     return jd;
 }
 
+int str_is_valid_repo_name(const char *str)
```
sshedi, line 270: Please make this a static function

sshedi (commented):
Please add few tests for this

tools/config/main.c
oliverkurth: Done.

tools/config/main.c
oliverkurth: Thank you, I changed it.

tools/config/main.c
oliverkurth: Done.

PR #535, opened 2025-08-25
glob packages uninstall fix when all repos disabled.

sshedi:
Do not exclude @System repo while applying repo filter

    Excluding @System repo will break package removals with globs with all
    repos disabled.

    Steps to reproduce:
    ```
    tdnf install -y openjdk11
    tdnf remove -y openjdk\* --disablerepo=*
    ```
    The second command errors out.

### DIFF
--- a/pytests/tests/test_downgrade.py
+++ b/pytests/tests/test_downgrade.py
@@ -21,8 +21,10 @@ def teardown_test(utils):
 
 
 def test_downgrade_no_arg(utils):
-    ret = utils.run(['tdnf', 'downgrade'])
-    assert ret['retval'] == 1011
+    cmd = "tdnf downgrade -y --disablerepo=* --enablerepo=photon-test".split()
+    ret = utils.run(cmd)
+    assert ret['retval'] == 0
+    assert "Downgrading: tdnf-test-multiversion" in " ".join(ret["stdout"])
 
 
 def test_downgrade_install(utils):
--- a/solv/tdnfquery.c
+++ b/solv/tdnfquery.c
@@ -249,7 +249,7 @@ SolvApplySinglePackageFilter(
 {
     uint32_t dwError = 0;
     char** ppCopyOfpkgNames = NULL;
-    if(!pQuery || !pszPackageName || IsNullOrEmptyString(pszPackageName))
+    if(!pQuery || IsNullOrEmptyString(pszPackageName))
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
--- a/solv/tdnfquery.c
+++ b/solv/tdnfquery.c
@@ -399,13 +399,10 @@ SolvAddAvailableRepoFilter(
     pool = pQuery->pSack->pPool;
     FOR_REPOS(i, pRepo)
     {
-        if (strcasecmp(SYSTEM_REPO_NAME, pRepo->name))
-        {
-            queue_push2(
-                &pQuery->queueRepoFilter,
-                SOLVER_SOLVABLE_REPO | SOLVER_SETREPO | SOLVER_SETVENDOR,
-                pRepo->repoid);
-        }
+        queue_push2(
+            &pQuery->queueRepoFilter,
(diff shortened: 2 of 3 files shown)


pytests/tests/test_downgrade.py
```diff
@@ -21,8 +21,23 @@ def teardown_test(utils):
 
 
 def test_downgrade_no_arg(utils):
+    cmd = "rpm -qa".split()
+    ret = utils.run(cmd)
+    assert ret['retval'] == 0
+
+    for p in ret['stdout']:
```
oliverkurth: This removes all test packages, right? If this is the intention, this deserves a comment, and better yet, a separate function. Also, I think the filtering you do is a little fragile, and may accidentally remove packages from the system (the container or VM, wherever the test is running).

pytests/tests/test_downgrade.py
```diff
@@ -21,8 +21,23 @@ def teardown_test(utils):
...
+            continue
+        if ".ph" in p or ".fc" in p:
+            continue
+
+        print(f"Removing {p} ...")
+        cmd = f"tdnf remove -y {p}".split()
+        ret = utils.run(cmd)
+        assert ret['retval'] == 0
+
     ret = utils.run(['tdnf', 'downgrade'])
```
oliverkurth: This downgrades *all* packages in the test system. This may have severe side effects.

solv/tdnfquery.c
```diff
@@ -399,13 +399,10 @@ SolvAddAvailableRepoFilter(
...
     FOR_REPOS(i, pRepo)
     {
-        if (strcasecmp(SYSTEM_REPO_NAME, pRepo->name))
-        {
-            queue_push2(
-                &pQuery->queueRepoFilter,
-                SOLVER_SOLVABLE_REPO | SOLVER_SETREPO | SOLVER_SETVENDOR,
-                pRepo->repoid);
-        }
+        queue_push2(
```
oliverkurth, line 402: Removing the condition means that we would list installed packages. "available" means packages that are in the repositories, but not installed packages. I need to double check, but I think this would break the `list --available` and `repoquery --available` commands.

oliverkurth (changes requested):

pytests/tests/test_downgrade.py
sshedi: Yeah, that's why used `.ph` & `.fc` filtering. Feel free to suggest better approach on this.

pytests/tests/test_downgrade.py
sshedi: Right, will try to keep only test repo enabled. Thought of keeping it as it is with no args but you are right. This needs further fixing. But, the previous `1011` retval checking was also wrong.

solv/tdnfquery.c
sshedi: You are right, `@System` packages get listed when the commands you suggested are invoked. Should I add a filter before enabling `@System` repo? Like only during `remove` or `erase` operations.

pytests/tests/test_downgrade.py
sshedi: And there is a catch, there is no `-y` option specified here, so nothing really happens.

solv/tdnfquery.c
oliverkurth: The problem is in the `TDNFGetGlobPackages()` function, called by `TDNFPrepareAllPackages()`. It calls `SolvFindAvailablePkgByName` even for `remove` operations. I think the right solution would be to add a parameter to `TDNFGetGlobPackages()` to indicate if we are looking for installed or for available packages, and call `SolvFindInstalledPkgByName()` or `SolvFindAvailablePkgByName` depending on that flag. We are looking for installed packages when the operation is `ALTER_ERASE`, or `ALTER_AUTOERASE`. And I think also for `ALTER_DOWNGRADE`, `ALTER_UPGRADE` and `ALTER_REINSTALL`.

oliverkurth: See https://github.com/vmware/tdnf/pull/537 .

sshedi:
Closing this in favour of https://github.com/vmware/tdnf/pull/537

PR #536, opened 2025-08-25
Improve error message while failing an operation

sshedi:
Currently if a third party rpm is installed in Photon and if we do: `tdnf downgrade` it fails and error we see is:
'Error(1011) : No matching packages'
And we will not know which package caused the failure.

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -255,6 +255,10 @@ TDNFCreateRepoFromDirectory(
     BAIL_ON_TDNF_ERROR(dwError);
 
     dwError = TDNFIsDir(pszPath, &nIsDir);
+    if (dwError) {
+        pr_err("CreateRepoFromDir: Error while operating on '%s', '%s'\n",
+                pszPath, strerror(errno));
+    }
     BAIL_ON_TDNF_ERROR(dwError);
 
     if (!nIsDir)
@@ -321,6 +325,10 @@ TDNFCreateRepoFromPath(
     if (pszPath[0] == '/')
     {
         dwError = TDNFIsDir(pszPath, &nIsDir);
+        if (dwError) {
+            pr_err("CreateRepoFromPath: Error while operating on '%s', '%s'\n",
+                    pszPath, strerror(errno));
+        }
         BAIL_ON_TDNF_ERROR(dwError);
 
         if (nIsDir)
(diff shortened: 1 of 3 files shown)


sshedi:
Another simple way to see this issue is;
```
rpm -Uvh https://packages.vmware.com/photon/5.0/photon_debuginfo_5.0_x86_64/x86_64/aide-debuginfo-0.17.4-5.ph5.x86_64.rpm

rpm -qa | grep '\.ph' > x.txt
tdnf install -qy $(cat x.txt) --disablerepo=* --enablerepo=photon-updates
Or
tdnf downgrade -qy $(cat x.txt) --disablerepo=* --enablerepo=photon-updates
```

client/repolist.c
```diff
@@ -255,6 +255,10 @@ TDNFCreateRepoFromDirectory(
     BAIL_ON_TDNF_ERROR(dwError);
 
     dwError = TDNFIsDir(pszPath, &nIsDir);
+    if (dwError) {
+        pr_err("(%s): Error while operating on '%s', '%s'\n",
+                __func__, pszPath, strerror(errno));
+    }
```
oliverkurth: We can instead use a macro.

I actually use this occasionally for debug builds:
```
$ diff include/tdnf-common-defines.h include/tdnf-common-defines.h.trace
18a19,20
> #define DO_TRACE 1
>
22a25
>             if (DO_TRACE) printf("error %d in %s at %s:%d\n", dwError, __FUNCTION__, __FILE__, __LINE__); \
```
We could add a build switch that enables this debugging code, and we will see a trace for every  `BAIL_ON_TDNF_ERROR` , including function name and line number.

oliverkurth (changes requested):

client/repolist.c
sshedi: This is a good macro while experimenting. Thanks. But this will make user experience difficult.
The two or three places where I added the error message are places which fail with no specific info. There could be more but I found these at the moment. Being a tdnf active user, I got confused and had no idea what the error message meant, so decided to add them.

oliverkurth: Curious, what *was* the issue? I am okay with adding an error message, but please without the function name and line number. We do not have that anywhere else.

Also for the other change below. The other error messages look okay to me.

sshedi: I was taking package list backup and doing some install and removes and then tried to restore the packages from the backup file I took earlier. And started seeing this confusing errors. I used function name because error messages look identical, and I don't like same error messages in multiple places. Will use something else as a marker in error message.

PR #537, opened 2025-08-27
fix globbing to match installed packages for erase/upgrade/downgrade/reinstall ops

oliverkurth:
This should supersede https://github.com/vmware/tdnf/pull/535 . I cherry-picked two commits from there for proper credits.

When using globs for packages, we should search installed packages for  erase, upgrade, downgrade and reinstall operations, not available packages.

### DIFF
--- a/client/resolve.c
+++ b/client/resolve.c
@@ -123,10 +123,17 @@ TDNFPrepareAllPackages(
 
            if(TDNFIsGlob(pszPkgName))
            {
+               int nIsInstalled = (nAlterType == ALTER_ERASE ||
+                                   nAlterType == ALTER_AUTOERASE ||
+                                   nAlterType == ALTER_UPGRADE ||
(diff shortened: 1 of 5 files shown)


client/resolve.c
```diff
@@ -123,10 +123,13 @@ TDNFPrepareAllPackages(
 
            if(TDNFIsGlob(pszPkgName))
            {
+               int nIsInstalled = (nAlterType == ALTER_ERASE || nAlterType == ALTER_AUTOERASE || nAlterType == ALTER_UPGRADE || nAlterType == ALTER_DOWNGRADE || nAlterType == ALTER_REINSTALL);
```
sshedi: Thanks for this fix, let's use a switch case, makes it easy to read. LGTM otherwise.

oliverkurth: I broke up that line, but not using a `switch` (it's just two cases). But this makes it easier to read.

PR #541, opened 2025-09-18
Use docker buildx in workflow

sshedi:


oliverkurth:
Thank, looks good to me.

PR #543, opened 2025-10-20
add option --nocmdlinegpgcheck to disable gpgcheck just for packages …

oliverkurth:
This disables GPG check just for local packages specified on the command line, but does not change any GPG check configuration for other packages that need to be installed as dependencies. This is useful if you trust the local packages, for example because they were built locally.

Using `nogpcheck` would also disable GPG checking for other packages that are pulled as dependencies.

### DIFF
--- a/pytests/tests/test_rpm_cmdline.py
+++ b/pytests/tests/test_rpm_cmdline.py
@@ -176,3 +176,35 @@ def test_reinstall_as_file(utils):
     assert utils.check_package(pkgname)
     assert "Nothing to do" not in "\n".join(ret['stderr'])
     assert "Reinstalling" in "\n".join(ret['stdout'])
+
+
+# test something like "tdnf install /path/to/pkg.rpm"
+# with nocmdlinegpgcheck option
+def test_install_as_file_nocmdlinegpgcheck(utils):
+    pkgname = utils.config["sglversion_pkgname"]
+    path = get_pkg_file_path(utils, pkgname)
+
+    # make sure we will fail if option isn't set
+    ret = utils.run(['tdnf', 'install', '-y', '--setopt=gpgcheck=1', path])
+    assert ret['retval'] != 0
+    assert not utils.check_package(pkgname)
+
+    ret = utils.run(['tdnf', 'install', '-y', '--setopt=gpgcheck=1', '--nocligpgcheck', path])
+    assert ret['retval'] == 0
+    assert utils.check_package(pkgname)
+
+
+# test something like "tdnf install /path/to/pkg.rpm"
+# with gpgcheck set to 1, but cligpgcheck set to 0
+def test_install_as_file_nocmdlinegpgcheck_conf(utils):
+    pkgname = utils.config["sglversion_pkgname"]
+    path = get_pkg_file_path(utils, pkgname)
+
+    # make sure we will fail if option isn't set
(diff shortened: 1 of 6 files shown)


pytests/tests/test_rpm_cmdline.py
```diff
@@ -176,3 +176,13 @@ def test_reinstall_as_file(utils):
     assert utils.check_package(pkgname)
     assert "Nothing to do" not in "\n".join(ret['stderr'])
     assert "Reinstalling" in "\n".join(ret['stdout'])
+
+
+# test something like "tdnf install /path/to/pkg.rpm"
```
sshedi, line 181: Add a negative test as well, where installation fails when option not speicified.

tools/cli/lib/parseargs.c
```diff
@@ -40,6 +40,7 @@ static struct option pstOptions[] =
     {"noautoremove",  no_argument, &_opt.nNoAutoRemove, 1},
     {"nodeps",        no_argument, &_opt.nNoDeps, 1},
     {"nogpgcheck",    no_argument, &_opt.nNoGPGCheck, 1},  //--nogpgcheck
+    {"nocmdlinegpgcheck", no_argument, &_opt.nNoCmdLineGPGCheck, 1},  //--nolocalgpgcheck
```
sshedi: How about we avoid adding a new option all together and make `nogpgcheck` a special arg?
```
tdnf --nogpgcheck  --> (nothing specified)Disable gpgcheck entirely
tdnf --nogpgcheck=*  --> same as above
tdnf --nogpgcheck=<repo-name>  --> Disable gpgcheck only for specified repo
tdnf --nogpgcheck=repo1,repo2,...  --> Disable gpgcheck only for specified repos
tdnf --nogpgcheck=cmdline  --> Do what this PR does
```

This will be more useful and easy to explain to users while retaining existing behaviour.

oliverkurth: Thank you for the suggestion, but in 3.6, we have this functionality already: you can set any config option for a repository like `--setopt=photon-updates.nogpgcheck=0`. This would also work for the command line packages with `--setopt=@cmdline.nogpgcheck=0`. So this new option would be just a shortcut for the command line, and I also intend to port this to 3.5 (where we do not have a way to override any arbitrary config variable) . I think we should leave it at that.

sshedi: I understand the ease of implementation in this approach but we will introduce another command line arg which has to be maintained forever. It would be good to enhance existing `--nogpgecheck` arg and make consumers understand and do when they want to use this.

Thanks for the info on `--setopt=@cmdline.nogpgcheck=0`, TIL.

oliverkurth: I think the tech debt that this adds is smaller than adding your suggestion. It would be the first option that has optional arguments, and I am worried that this will have side effects (for example the argument may be confused with a command). And the use case for the command line is much more common than singling out one or more repositories. Since there is a way in 3.6 anyway to accomplish the same thing, I'd very much prefer the implementation as is in this PR.

sshedi: Valid argument. Two suggestions:
1. `--nocmdlinegpgcheck` -> `--nocligpgcheck`
2. And I think it is worth giving a global override for this through `/etc/tdnf/tdnf.conf` if user doesn't want to use the flag each time.

PR #548, opened 2025-11-04
Fix tdnf list and tdnf search commands while multiple keys are given at a time

sshedi:
Currently tdnf returns with zero exit status when a command like `tdnf search tdnf invalid linux` or `tdnf search linux invalid123 tdnf` is invoked, if we are passing args in bulk, and if there are multiple keys with no results tdnf doesn't show any errors. This patch is an attempt to address this issue. Corresponding tests are added.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -550,6 +550,10 @@ TDNFListInternal(
     *pdwCount = dwCount;
 
 cleanup:
+    if (!dwError && pQuery && pQuery->queryError)
+    {
+        dwError = pQuery->queryError;
+    }
     if(pQuery)
     {
         SolvFreeQuery(pQuery);
@@ -558,6 +562,7 @@ cleanup:
     {
         SolvFreePackageList(pPkgList);
     }
+
     return dwError;
 error:
     if(ppPkgInfo)
@@ -869,7 +874,6 @@ TDNFProvides(
 
     dwError = SolvCreateQuery(pTdnf->pSack, &pQuery);
     BAIL_ON_TDNF_ERROR(dwError);
-
     dwError = SolvApplySinglePackageFilter(pQuery, pszSpec);
     BAIL_ON_TDNF_ERROR(dwError);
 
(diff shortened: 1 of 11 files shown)


oliverkurth:
I double checked with Fedora's `dnf`, and it does not behave that way. Only if none of the arguments match does it fail. What is your use case here, or why do you want this behavior change?

sshedi:
> I double checked with Fedora's `dnf`, and it does not behave that way. Only if none of the arguments match does it fail. What is your use case here, or why do you want this behavior change?

Yes, I checked in fedora as well before making this change and I think it is an issue in dnf as well. While debugging an issue I got a big list of packages (500+ entires) and I wanted to see what were missing in my system uding `tdnf list <pkg-list>` but it didn't show anything.

I don't mind if we don't error out but we should be explicit about mishits with tdnf list and tdnf search.

sshedi:
Something like:
```
$ tdnf list linux invalid tdnf
linux ....
No match found for 'invalid'
tdnf ....

And exit with zero status
```

```
$ tdnf list search invalid123 tdnf
linux ....
...
No match found for 'invalid123'
...
tdnf ....
...
And exit with zero status
```

oliverkurth:
I am a little worried about users using scripts relying on existing behavior. How about adding a command line option that changes to the new behavior?

sshedi:
> I am a little worried about users using scripts relying on existing behavior. How about adding a command line option that changes to the new behavior?

Not one more command line option please; there are so many already. Let this be a breaking change, users should be aware that something failed silently in their frameworks which should be fixed. We can document it as a breaking change in tdnf. 

Something like this in this project
https://github.com/canonical/cloud-init/blob/a2209a9e1d3ea990a4699f16dfeb64c12d1e5a2f/doc/rtd/reference/breaking_changes.rst#L3

oliverkurth:
When I do
```
./bin/tdnf list glibc nosuchpkg
```
I see no results for `glibc`, just an error:
```
No match found for 'nosuchpkg'
Error(1011) : No matching packages
```
I think we should show the partial matches.

PR #550, opened 2025-11-19
make snapshot files fetchable by URL

oliverkurth:
Make snapshot files fetchable by URL.

If the snapshot file is not local, it will be downloaded to the cache, unless it already exists and is not expired, with the same expiry time as the whole repo. The file will be downloaded as `snapshot-<hash>`, where the hash is calculated from the URL - this is to make sure that the file will be downloaded even if a previous version exists but the URL has changed. On cleanup all snapshot files will be removed, by matching the pattern.

The behavior for local snapshot files has not changed. It will be read directly, and not copied to cache.

### DIFF
--- a/client/repo.c
+++ b/client/repo.c
@@ -23,7 +23,7 @@ TDNFApplySnapshot(
     int i;
     Queue qResult = {0};
 
-    if(!pRepoData || !pRepoData->pszSnapshotUrl || !pSack || !pSack->pPool)
+    if(!pRepoData || !pRepoData->pszSnapshotFile || !pSack || !pSack->pPool)
     {
         dwError = ERROR_TDNF_INVALID_PARAMETER;
         BAIL_ON_TDNF_ERROR(dwError);
@@ -32,7 +32,7 @@ TDNFApplySnapshot(
 
     queue_init(&qResult);
 
-    dwError = TDNFReadFileToStringArray(pRepoData->pszSnapshotUrl, &ppszSnapshotPackages);
+    dwError = TDNFReadFileToStringArray(pRepoData->pszSnapshotFile, &ppszSnapshotPackages);
     BAIL_ON_TDNF_ERROR(dwError);
 
     for(i = 0; ppszSnapshotPackages[i]; i++) {
@@ -475,6 +475,7 @@ TDNFGetRepoMD(
     char *pszTmpRepoDataDir = NULL;
     char *pszTmpRepoMDFile = NULL;
     char *pszMirrorFile = NULL;
+    char *pszSnapshotFile = NULL;
     char *pszTempBaseUrlFile = NULL;
     char* pszLastRefreshMarker = NULL;
     PTDNF_REPO_METADATA pRepoMDRel = NULL;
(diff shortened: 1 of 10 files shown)


client/repo.c
```diff
@@ -693,6 +694,56 @@ TDNFGetRepoMD(
     BAIL_ON_TDNF_ERROR(dwError);
     *ppRepoMD = pRepoMD;
 
+    if (pRepoData->pszSnapshotUrl) {
+        time_t now = time(NULL);
+        int needDownload = 0, nIsRemote = 0;
+        struct stat st = {0};
+
+        dwError = TDNFUriIsRemote(pRepoData->pszSnapshotUrl, &nIsRemote);
```
sshedi: Check if it is local file first and then see if it is valid url and there should be a error message thrown accordingly.

client/repo.c
```diff
@@ -693,6 +694,56 @@ TDNFGetRepoMD(
...
+                }
+            } else if ((now - st.st_ctime) > pRepoData->lMetadataExpire)
+                needDownload = 1;
+
+            if (needDownload) {
+                dwError = TDNFDownloadFile(pTdnf, pRepoData, pRepoData->pszSnapshotUrl, pRepoData->pszSnapshotFile, pRepoData->pszId);
+                BAIL_ON_TDNF_ERROR(dwError);
+            }
+        } else {
+            if (pRepoData->pszSnapshotUrl[0] == '/') {
```
sshedi: Should we also support `file://`?

pytests/tests/test_snapshots.py
```diff
@@ -102,6 +102,18 @@ def test_list(utils):
...
+
+    ret = utils.run(["tdnf", "-j", "--repoid", REPONAME, "--available", f"--setopt=snapshot.{REPONAME}={snapshot_url}", "list"])
+    infolist = json.loads("\n".join(ret['stdout']))
+
+    # excluded packages should not be listed
+    for info in infolist:
+        nevr = f"{info['Name']}={info['Evr']}"
+        assert nevr not in EXCLUDED_PKGS
+
+
```
sshedi, line 147: Please add 3 more tests.
1. With a local file
2. With invalid url
3. With non existent local file

client/repo.c
oliverkurth: Done.

client/repo.c
oliverkurth: Thanks, I forgot. Done.

pytests/tests/test_snapshots.py
oliverkurth: Done.

client/repo.c
```diff
@@ -691,6 +692,60 @@ TDNFGetRepoMD(
...
+                            BAIL_ON_TDNF_SYSTEM_ERROR(dwError);
+                        }
+                    } else if ((now - st.st_ctime) > pRepoData->lMetadataExpire)
+                        needDownload = 1;
+
+                    if (needDownload) {
+                        dwError = TDNFDownloadFile(pTdnf, pRepoData, pRepoData->pszSnapshotUrl, pRepoData->pszSnapshotFile, pRepoData->pszId);
+                        BAIL_ON_TDNF_ERROR(dwError);
+                    }
+                } else if (strncmp(pRepoData->pszSnapshotUrl, "file://", 7) == 0) {
```
sshedi, line 739: JFYI: instead of 7 you can use `sizeof("file://") - 1`

sshedi (approved):

client/repo.c
oliverkurth: I know, but this makes it easier to read, IMHO.

PR #553, opened 2025-12-06
Enhance enablerepo, disablerepo

sshedi:
Accept comma-separated repository names and globs as well.

Currently, enabling or disabling specific repositories requires multiple flags, for example:
```
tdnf ...  --disablerepo=repo1 --disablerepo=repo2 --enablerepo=repo3 --enablerepo=repo4
```

This PR extends the behavior so that each flag can take a comma-separated list, allowing the same operation with:
```
tdnf ...  --disablerepo=repo1,repo2 --enablerepo=repo3,repo4
```

Globs are also supported.

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -726,33 +726,34 @@ TDNFRepoListFinalize(
     /* There could be overrides to enable/disable
        repo such as cmdline args, api overrides */
     for (cn = pTdnf->pArgs->cn_setopts->first_child; cn; cn = cn->next) {
-        if(strcmp(cn->name, "enablerepo") == 0) {
-            dwError = TDNFAlterRepoState(
-                          pTdnf->pRepos,
-                          1,
-                          cn->value);
-        }
-        else if(strcmp(cn->name, "disablerepo") == 0) {
-            dwError = TDNFAlterRepoState(
-                          pTdnf->pRepos,
-                          0,
-                          cn->value);
-        }
-        else if((strcmp(cn->name, "repo") == 0) ||
-                (strcmp(cn->name, "repoid") == 0)) {
-            if (!nRepoidSeen)
-            {
-                dwError = TDNFAlterRepoState(
-                              pTdnf->pRepos, 0, "*");
-                BAIL_ON_TDNF_ERROR(dwError);
-                nRepoidSeen = 1;
+        int isEnabled = -1;
+        char **ppszRepos = NULL;
+
+        if (!strcmp(cn->name, "enablerepo"))
+            isEnabled = 1;
+        else if (!strcmp(cn->name, "disablerepo"))
+            isEnabled = 0;
+
+        if (isEnabled != -1) {
+            dwError = TDNFSplitStringToArray(cn->value, ",", &ppszRepos);
+            BAIL_ON_TDNF_ERROR(dwError);
+
+            for (int i = 0; ppszRepos[i]; ++i) {
+              dwError = TDNFAlterRepoState(pTdnf->pRepos, isEnabled, ppszRepos[i]);
+              BAIL_ON_TDNF_ERROR(dwError);
             }
-            dwError = TDNFAlterRepoState(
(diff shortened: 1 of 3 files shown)


oliverkurth:
I wonder if we can use the `FNM_EXTMATCH` flag instead for `fnmatch()`, see the man page:
```
       FNM_EXTMATCH
              If this flag (a GNU extension) is set, extended patterns are supported, as introduced by 'ksh' and now supported by other shells.  The extended format is as follows, with pattern-list being a '|' separated list of patterns.
...
       '+(pattern-list)'
              The pattern matches if one or more occurrences of any of the patterns in the pattern-list match the input string.
...
```
This would accomplish the same thing, if we can do `--disablerepo=+(repo1,repo2)`.

I wish we could use patterns like `{foo,bar}`, but I do not think `fnmatch()` supports that, and I do not know what does.

sshedi:
> I wonder if we can use the `FNM_EXTMATCH` flag instead for `fnmatch()`, see the man page:
> 
> ```
>        FNM_EXTMATCH
>               If this flag (a GNU extension) is set, extended patterns are supported, as introduced by 'ksh' and now supported by other shells.  The extended format is as follows, with pattern-list being a '|' separated list of patterns.
> ...
>        '+(pattern-list)'
>               The pattern matches if one or more occurrences of any of the patterns in the pattern-list match the input string.
> ...
> ```
> 
> This would accomplish the same thing, if we can do `--disablerepo=+(repo1,repo2)`.
> 
> I wish we could use patterns like `{foo,bar}`, but I do not think `fnmatch()` supports that, and I do not know what does.

For TDNF use cases, `FNM_EXTMATCH` could be overkill. Regex matching is hard to remember and document. For now the existing implementation is enough I guess, using brackets and some other special characters. `TDNFIsGlob` already has some amount of this, that should suffice.

client/repolist.c
```diff
@@ -726,33 +726,34 @@ TDNFRepoListFinalize(
...
-        }
-        else if((strcmp(cn->name, "repo") == 0) ||
-                (strcmp(cn->name, "repoid") == 0)) {
-            if (!nRepoidSeen)
-            {
-                dwError = TDNFAlterRepoState(
-                              pTdnf->pRepos, 0, "*");
-                BAIL_ON_TDNF_ERROR(dwError);
-                nRepoidSeen = 1;
+        int isRepoFlag = -1;
```
oliverkurth: Nit: `isRepoFlag` is a confusing name, please rename to something like `isEnabled`.

client/repolist.c
```diff
@@ -851,6 +852,8 @@ TDNFAlterRepoState(
 
     for (int nMatch = 0; pRepos; pRepos = pRepos->pNext)
     {
+        nMatch = 0;
```
oliverkurth: Oh, looks like this fixes a bug? And you can remove the `nMatch = 0` in the `for(...` just above.

oliverkurth (changes requested):
Okay, I'll accept with just a few minor changes. Thank you for this!

client/repolist.c
sshedi: Yes, first commit has the details.

client/repolist.c
```diff
@@ -726,33 +726,34 @@ TDNFRepoListFinalize(
...
+
+            TDNF_SAFE_FREE_STRINGARRAY(ppszRepos);
+        } else if (!strcmp(cn->name, "repo") || !strcmp(cn->name, "repoid")) {
+            if (!nRepoidSeen) {
+              dwError = TDNFAlterRepoState(pTdnf->pRepos, 0, "*");
+              BAIL_ON_TDNF_ERROR(dwError);
+              nRepoidSeen = 1;
+            }
+
+            dwError = TDNFAlterRepoState(pTdnf->pRepos, 1, cn->value);
```
oliverkurth: I just noticed this - can we have the same enhancement here? For `--repoid=foo,bar`.

PR #555, opened 2025-12-20
Enhancements in help message and bash completions

sshedi:


etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -40,6 +40,61 @@ _tdnf__process_if_prev_is_option()
             COMPREPLY=( $(compgen -W "$opts" -- $cur) )
             return 0
             ;;
+        --forcearch)
+            opts="x86_64 aarch64 noarch"
+            COMPREPLY=( $(compgen -W "$opts" -- $cur) )
+            return 0
+            ;;
+        --releasever)
+            COMPREPLY=( $(compgen -W "7 8 9" -- $cur) )
```
oliverkurth: We only allow versions 7, 8 and 9? This is usually set to "5.0". It can be any version string. It might even be non-numeric.

oliverkurth:
I still see a few misalignments:
```
...
--refresh                  Refresh repository metadata before operation
                            Example: tdnf --refresh install package
...
--sec-severity <level>     Filter by security severity (Critical|Important|Moderate|Low)
                           Example: tdnf updateinfo --sec-severity=Critical
--security                  Only show security updates
                            Example: tdnf --security update
--setopt <option>=<value>  Set configuration option
                            Example: tdnf --setopt=keepcache=1 install package
--skip-broken              Skip packages with broken dependencies
                           Example: tdnf --skip-broken install pkg
```
There are more, just check.

Typo also:
```
--download-path <dir>      Directory to download packages to
                            Exampl tdnf reposync --download-path=/mnt/repo
```

I will install it also, to see if tab completion works as expected.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -40,6 +40,61 @@ _tdnf__process_if_prev_is_option()
             COMPREPLY=( $(compgen -W "$opts" -- $cur) )
             return 0
             ;;
+        --forcearch)
+            opts="x86_64 aarch64 noarch"
```
oliverkurth: I don't think `noarch` makes sense here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -40,6 +40,61 @@ _tdnf__process_if_prev_is_option()
...
+        --arch)
+            opts="x86_64 aarch64 noarch"
+            COMPREPLY=( $(compgen -W "$opts" -- $cur) )
+            return 0
+            ;;
+        --metadata-path)
+            COMPREPLY=( $(compgen -d -- $cur) )
+            return 0
+            ;;
+        --whatdepends|--whatrequires|--whatprovides|--whatobsoletes|--whatconflicts|--whatrecommends|--whatsuggests|--whatsupplements|--whatenhances)
```
oliverkurth, line 77: This should be packages:
```
root@photon-e440344eaaf3 [ ~ ]# tdnf repoquery --whatdepends toybox
raspberrypi-firmware-pi3-1.20230106-2.ph5.noarch
...
```

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,32 +111,99 @@ _tdnf__process_if_cmd()
...
+            # After repoquery, offer both repoquery-specific options and package names
+            local repoquery_opts="--available --duplicates --extras --file --installed --userinstalled --upgrades --whatconflicts --whatdepends --whatenhances --whatobsoletes --whatprovides --whatrecommends --whatrequires --whatsuggests --whatsupplements --changelogs --conflicts --depends --enhances --list --location --obsoletes --provides --qf --recommends --requires --requires-pre --suggests --supplements --source"
+            local pkg_names="$(tdnf list --available 2>/dev/null | awk '{print $1}' | cut -d'.' -f1) $(tdnf list --installed 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)"
+            opts="$repoquery_opts $pkg_names"
             ;;
         update|upgrade)
-            opts=$(tdnf list --upgrades | awk '{print $1}' | cut -d'.' -f1)
+            opts=$(tdnf list --upgrades 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)
+            ;;
+        check|help|makecache|nocligpgcheck)
```
oliverkurth: `nocligpgcheck` is an option, not a command.

tools/cli/lib/help.c
```diff
@@ -9,113 +9,407 @@
...
+ "--forcearch <arch>         Force architecture (x86_64|aarch64|noarch)\n"
+ "                           Example: tdnf --forcearch=x86_64 install\n"
+
+ "-h, --help                 Show this help message\n"
+ "                           Example: tdnf --help\n"
+
+ "-i, --installroot <path>   Set install root path (must be absolute)\n"
+ "                           Example: tdnf -i /mnt/chroot install pkg\n"
+
+ "--json                     Output in JSON format\n"
```
oliverkurth: Also `-j`.

tools/cli/lib/help.c
```diff
@@ -9,113 +9,407 @@
...
+ "--noautoremove             Do not remove automatically installed dependencies\n"
+ "                           Example: tdnf --noautoremove remove pkg\n"
+
+ "--nodeps                   Skip dependency checks(requires --downloadonly)\n"
+ "                           Example: tdnf --downloadonly --nodeps install package\n"
+
+ "--nogpgcheck               Skip GPG signature checks\n"
+ "                           Example: tdnf --nogpgcheck install pkg\n"
+
+ "--nocligpgcheck            Skip GPG signature checks for local rpms\n"
```
oliverkurth: Not local, but packages on the command line. For example, it would affect checking when run like `tdnf install --nopcligpgcheck http://foo.bar.com/package.rpm`.

Packages that are local, but in a repository (using `baseurl=file:///mylocal/rpms`) would NOT be affected.

tools/cli/lib/help.c
```diff
@@ -9,113 +9,407 @@
...
+ "--refresh                  Refresh repository metadata before operation\n"
+ "                            Example: tdnf --refresh install package\n"
+
+ "--releasever <version>     Set release version\n"
+ "                           Example: tdnf --releasever=5.0 install pkg\n"
+
+ "--repo <repoid>            Enable only this repository\n"
+ "                           Example: tdnf --repo=base install pkg\n"
+
+ "--repofromdir <dir>        Add repository from directory\n"
```
oliverkurth: This needs a repo id also, just like `repofrompath`.

tools/cli/lib/help.c
```diff
@@ -9,113 +9,407 @@
...
+ "                           Example: tdnf --repofromdir=/path/to/repo install package\n"
+
+ "--repofrompath <id>,<path> Add repository from path\n"
+ "                           Example: tdnf --repofrompath=local,/mnt/repo install package\n"
+
+ "--repoid <repoid>          Same as --repo\n"
+ "                           Example: tdnf --repoid=base install pkg\n"
+
+ "--rpmdefine <macro>        Define RPM macro\n"
+ "                           Example: tdnf --rpmdefine='_prefix /usr' install package\n"
```
oliverkurth: I think this is a bad example, I don't think it would have the expected effect. It's usually used to set the rpmdb location. I cannot think of another example, that doesn't mean they don't exist.

oliverkurth (changes requested):

sshedi:
Thanks for the review. Fixed everything I caught, PTAL again.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -40,12 +39,67 @@ _tdnf__process_if_prev_is_option()
...
+            return 0
+            ;;
+        --file)
+            COMPREPLY=( $(compgen -f -- $cur) )
+            return 0
+            ;;
+        --arch)
+            opts="x86_64 aarch64"
+            COMPREPLY=( $(compgen -W "$opts" -- $cur) )
+            return 0
```
oliverkurth, line 71: Sorry, you went too far here - for `--arch`, `noarch` is legit. It's used for `reposync` and `repoquery` to filter packages.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -40,12 +39,67 @@ _tdnf__process_if_prev_is_option()
...
+            ;;
+        --metadata-path)
+            COMPREPLY=( $(compgen -d -- $cur) )
+            return 0
+            ;;
+        --whatdepends|--whatrequires|--whatprovides|--whatobsoletes|--whatconflicts|--whatrecommends|--whatsuggests|--whatsupplements|--whatenhances)
+            opts=$(tdnf list --available 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)
```
oliverkurth: Why just `--available`? It could be used for installed packages that are not available.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -40,12 +39,67 @@ _tdnf__process_if_prev_is_option()
...
+            COMPREPLY=( $(compgen -W "$opts" -- $cur) )
+            return 0
+            ;;
+        --qf)
+            COMPREPLY=( $(compgen -W "" -- $cur) )
+            return 0
+            ;;
+        --to|--from)
+            COMPREPLY=( $(compgen -W "" -- $cur) )
+            return 0
```
oliverkurth, line 89: This takes numbers - can we complete for those?

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +110,110 @@ _tdnf__process_if_cmd()
...
+            if [ $1 -eq $(($COMP_CWORD - 1)) ]; then
                 opts="all enabled disabled"
+            else
+                return 0
+            fi
+            ;;
+        repoquery)
+            # After repoquery, offer both repoquery-specific options and package names
+            local repoquery_opts="--available --duplicates --extras --file --installed --userinstalled --upgrades --whatconflicts --whatdepends --whatenhances --whatobsoletes --whatprovides --whatrecommends --whatrequires --whatsuggests --whatsupplements --changelogs --conflicts --depends --enhances --list --location --obsoletes --provides --qf --recommends --requires --requires-pre --suggests --supplements --source"
+            local pkg_names="$(tdnf list --available 2>/dev/null | awk '{print $1}' | cut -d'.' -f1) $(tdnf list --installed 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)"
```
oliverkurth: I think you can just use `tdnf list` without `--available` or `--installed` and it will list all packages known. Makes this more efficient.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +110,110 @@ _tdnf__process_if_cmd()
...
-            opts=$(tdnf list --upgrades | awk '{print $1}' | cut -d'.' -f1)
+            opts=$(tdnf list --upgrades 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)
+            ;;
+        check|help|makecache)
+            # Commands that take no arguments
+            return 0
+            ;;
+        check-update)
+            # Optional package names
+            opts="$(tdnf list --available 2>/dev/null | awk '{print $1}' | cut -d'.' -f1) $(tdnf list --installed 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)"
```
oliverkurth: Same here, and all similar cases.

oliverkurth (changes requested):

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -40,12 +39,68 @@ _tdnf__process_if_prev_is_option()
...
+            opts="x86_64 aarch64 noarch"
+            COMPREPLY=( $(compgen -W "$opts" -- $cur) )
+            return 0
+            ;;
+        --metadata-path)
+            COMPREPLY=( $(compgen -d -- $cur) )
+            return 0
+            ;;
+        --whatdepends|--whatrequires|--whatprovides|--whatobsoletes|--whatconflicts|--whatrecommends|--whatsuggests|--whatsupplements|--whatenhances)
+            opts=$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)
```
oliverkurth: Since this is on the dev branch (or >= 3.6), we can use:
`tdnf repoquery --qf "%{name}" 2>/dev/null`

Also in lots of other places. It's more efficient.

However, it means we cannot backport this to 3.5 - which may be okay, since we can update tdnf to 3.6 in Photon 5.0rev2 soon.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
-            [ $1 -eq $(($COMP_CWORD - 1)) ] &&
+            if [ $1 -eq $(($COMP_CWORD - 1)) ]; then
                 opts="packages metadata dbcache plugins expire-cache all"
+            else
+                return 0
+            fi
             ;;
         downgrade)
-            opts=$(tdnf list --downgrades | awk '{print $1}' | cut -d'.' -f1)
+            opts=$(tdnf list --downgrades 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
             ;;
         autoerase|autoremove|erase|reinstall|remove)
-            opts=$(tdnf list --installed | awk '{print $1}' | cut -d'.' -f1)
+            opts=$(tdnf list --installed 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
                 opts="init update list rollback undo redo"
+            else
+                return 0
+            fi
+            ;;
+        count)
+            return 0
+            ;;
+        distro-sync)
+            opts=$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
             ;;
         install)
-            opts="$(tdnf list --available | awk '{print $1}' | cut -d'.' -f1) $(compgen -d -G '*.rpm')"
+            opts=$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
+            ;;
+        repoquery)
+            # After repoquery, offer both repoquery-specific options and package names
+            local repoquery_opts="--available --duplicates --extras --file --installed --userinstalled --upgrades --whatconflicts --whatdepends --whatenhances --whatobsoletes --whatprovides --whatrecommends --whatrequires --whatsuggests --whatsupplements --changelogs --conflicts --depends --enhances --list --location --obsoletes --provides --qf --recommends --requires --requires-pre --suggests --supplements --source"
+            local pkg_names=$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)
+            opts="$repoquery_opts $pkg_names"
             ;;
         update|upgrade)
-            opts=$(tdnf list --upgrades | awk '{print $1}' | cut -d'.' -f1)
+            opts=$(tdnf list --upgrades 2>/dev/null | awk '{print $1}' | cut -d'.' -f1)
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
+            ;;
+        check|help|makecache)
+            # Commands that take no arguments
+            return 0
+            ;;
+        check-update)
+            # Optional package names
+            opts="$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)"
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
+        info)
+            # Package names or scope options
+            local scope_opts="installed available updates downgrades recent all"
+            local pkg_names="$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)"
+            opts="$scope_opts $pkg_names"
+            ;;
+        list)
+            # Scope options or package names
+            local scope_opts="installed available updates downgrades recent all"
+            local pkg_names="$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)"
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
+            return 0
+            ;;
+        check-update)
+            # Optional package names
+            opts="$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)"
+            ;;
+        info)
+            # Package names or scope options
+            local scope_opts="installed available updates downgrades recent all"
+            local pkg_names="$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)"
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
+            # Optional repository names
+            opts=$(tdnf repolist all 2>/dev/null | awk '{if (NR > 1) print $1}')
+            ;;
+        search)
+            # Search terms - no easy way to list these, allow free-form input
+            return 0
+            ;;
+        update-to|upgrade-to)
+            # Package names (with optional version)
+            opts="$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)"
```
oliverkurth: Use `repoquery` here.

etc/bash_completion.d/tdnf-completion.bash
```diff
@@ -56,44 +111,110 @@ _tdnf__process_if_cmd()
...
+            ;;
+        updateinfo)
+            # Mode options or package names
+            local mode_opts="all info summary"
+            local pkg_names="$(tdnf list 2>/dev/null | awk '{print $1}' | sort -u | cut -d'.' -f1)"
```
oliverkurth: Use `repoquery` here.

oliverkurth (commented):
Looks good (thank you!), but I have one more suggestion: use `repoquery --qf` instead of `awk` and `cut` for `list` commands. I didn't suggest this earlier because it requires tdnf >= 3.6, but I realized that that is actually fine.

Also, I realized that we should have a simpler output version for `repolist`, and a way to get the current history count number. We can add that in another change.

etc/bash_completion.d/tdnf-completion.bash
sshedi: Yes, I understand and agree.

etc/bash_completion.d/tdnf-completion.bash
sshedi: This was missing, I added this support to repoquery.

etc/bash_completion.d/tdnf-completion.bash
sshedi: This command had an issue, fixed it.

oliverkurth:
Thank you, looks good!

PR #556, opened 2026-01-21
Check for RPMTAG_OPENPGP when detecting unsigned packages

sshedi:
This is needed if underlying RPM version used is 6.x

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ endif()
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
-### External dependency: rpm-devel
+# External dependency: rpm-devel
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(RPM REQUIRED rpm)
 find_package(OpenSSL REQUIRED)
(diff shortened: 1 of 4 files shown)


oliverkurth:
It would be good if we could check if the enum is defined using cmake. I asked gemini, and it suggests this:
```
include(CheckCSourceCompiles)

check_c_source_compiles("
#include <my_header.h>
int main() {
    int val = EXPECTED_VALUE;
    return 0;
}
" HAS_ENUM_VALUE)
```
Can you try implementing that to check for `RPMTAG_OPENPGP`?

sshedi:
> It would be good if we could check if the enum is defined using cmake. I asked gemini, and it suggests this:
> 
> ```
> include(CheckCSourceCompiles)
> 
> check_c_source_compiles("
> #include <my_header.h>
> int main() {
>     int val = EXPECTED_VALUE;
>     return 0;
> }
> " HAS_ENUM_VALUE)
> ```
> 
> Can you try implementing that to check for `RPMTAG_OPENPGP`?

Thank you for the suggestion. I added a global level check for `RPM-6.x`, we might need this macro in other places as well.

PR #557, opened 2026-02-19
Add configurable connect timeout for downloads

sshedi:
Introduce connect_timeout config option and apply it via CURLOPT_CONNECTTIMEOUT in remote downloads to avoid long hangs when DNS resolution succeeds but TCP connection does not establish.

### DIFF
--- a/client/config.c
+++ b/client/config.c
@@ -164,6 +164,10 @@ TDNFConfigFromCnfTree(PTDNF_CONF pConf, struct cnfnode *cn_top)
         {
             pConf->nCliGPGCheck = isTrue(cn->value);
         }
+        else if (strcmp(cn->name, TDNF_CONF_KEY_CONNECT_TIMEOUT) == 0)
+        {
+            pConf->nConnectTimeout = strtoi(cn->value);
+        }
         else if (strcmp(cn->name, TDNF_CONF_KEY_SSL_VERIFY) == 0)
         {
             pConf->nSSLVerify = isTrue(cn->value);
@@ -345,6 +349,7 @@ TDNFReadConfig(
     pConf->nOpenMax = TDNF_CONF_DEFAULT_OPENMAX;
     pConf->nInstallOnlyLimit = TDNF_CONF_DEFAULT_INSTALLONLY_LIMIT;
     pConf->nSSLVerify = TDNF_CONF_DEFAULT_SSLVERIFY;
+    pConf->nConnectTimeout = TDNF_CONF_DEFAULT_CONNECT_TIMEOUT;
 
     register_ini(NULL);
     mod_ini = find_cnfmodule("ini");
--- a/common/config.h
+++ b/common/config.h
@@ -56,6 +56,7 @@
 #define TDNF_CONF_KEY_VARS_DIRS           "varsdir"
 #define TDNF_CONF_KEY_CHECK_UPDATE_COMPAT "dnf_check_update_compat"
 #define TDNF_CONF_KEY_DISTROSYNC_REINSTALL_CHANGED "distrosync_reinstall_changed"
+#define TDNF_CONF_KEY_CONNECT_TIMEOUT     "connect_timeout"
 
 #define TDNF_PLUGIN_CONF_KEY_ENABLED      "enabled"
 
@@ -117,6 +118,7 @@
 #define TDNF_CONF_DEFAULT_OPENMAX            1024
 #define TDNF_CONF_DEFAULT_INSTALLONLY_LIMIT  2
 #define TDNF_CONF_DEFAULT_SSLVERIFY          1
+#define TDNF_CONF_DEFAULT_CONNECT_TIMEOUT    0  // seconds (unlimited)
 
 // repo default settings
 #define TDNF_REPO_DEFAULT_ENABLED            0
--- a/client/remoterepo.c
+++ b/client/remoterepo.c
@@ -218,6 +218,12 @@ TDNFDownloadFile(
     dwError = curl_easy_setopt(pCurl, CURLOPT_FOLLOWLOCATION, 1L);
     BAIL_ON_TDNF_CURL_ERROR(dwError);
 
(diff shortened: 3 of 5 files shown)


sshedi:
@oliverkurth if you have any idea on how this can be tested in a container, please let me know. I tried this manually.

Here are the steps:
```
1. ./bin/tdnf makecache -----> works
2. iptables -I OUTPUT -p tcp --dport 80 -j DROP (or 443 port)
3. ./bin/tdnf makecache -----> fails after 10 retries with 10 sec interval, can be modified through tdnf.conf
4. iptables -D OUTPUT -p tcp --dport 80 -j DROP
5. ./bin/tdnf makecache -----> works
```
In the existing code, step 3 hangs indefinitely (nearly an hour) to finish 10 retries

common/config.h
```diff
@@ -56,6 +56,7 @@
 #define TDNF_CONF_KEY_VARS_DIRS           "varsdir"
 #define TDNF_CONF_KEY_CHECK_UPDATE_COMPAT "dnf_check_update_compat"
 #define TDNF_CONF_KEY_DISTROSYNC_REINSTALL_CHANGED "distrosync_reinstall_changed"
+#define TDNF_CONF_KEY_CONNECT_TMOUT       "connect_timeout"
```
oliverkurth: Nit: I would prefer ` TDNF_CONF_KEY_CONNECT_TIMEOUT `.

We can afford two extra bytes ;-)

common/config.h
```diff
@@ -117,6 +118,7 @@
 #define TDNF_CONF_DEFAULT_OPENMAX            1024
 #define TDNF_CONF_DEFAULT_INSTALLONLY_LIMIT  2
 #define TDNF_CONF_DEFAULT_SSLVERIFY          1
+#define TDNF_CONF_DEFAULT_CONNECT_TMOUT      5  // seconds
```
oliverkurth: Set it to 0, so it defaults to no timeout - previous behavior.

If we really want a timeout be default though, I think 5 seconds is too short - maybe 60 seconds?

oliverkurth (changes requested):
For testing, I think you can use `iptables` also in the container. You may need to install it.

common/config.h
sshedi: The issue is happening due to 0 timeout, to establish initial steps of connection you hardly need less than a second. I think 5 sec is a realistic number. making this 60 seconds would implicitly mean, we end up wasting 10 minutes during each tdnf transaction.

sshedi: I have set it to 10 seconds, should be enough for any kind of slow connections.

oliverkurth:
If read this correctly: https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html :

"The connection phase includes the name resolve (DNS) and all protocol handshakes and negotiations until there is an established connection with the remote side."

It includes the DNS resolution. DNS can, under circumstances, require some time.

Also, the default is 300 seconds, it's not indefinitely: "Set this option to zero to switch to the default built-in connection timeout - 300 seconds."

sshedi:
> For testing, I think you can use `iptables` also in the container. You may need to install it.

It's not that simple, I think the container should be started in a different way to achieve it but I will add this to my backlog and will try to do it when I have bandwidth. Hope that's okay?

sshedi:
> If read this correctly: https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html :
> 
> "The connection phase includes the name resolve (DNS) and all protocol handshakes and negotiations until there is an established connection with the remote side."
> 
> It includes the DNS resolution. DNS can, under circumstances, require some time.
> 
> Also, the default is 300 seconds, it's not indefinitely: "Set this option to zero to switch to the default built-in connection timeout - 300 seconds."

IMO, if DNS + Handshake doesn't finish under 10 seconds, the consumer has bigger things to worry about than installing packages :)

oliverkurth:
> IMO, if DNS + Handshake doesn't finish under 10 seconds, the consumer has bigger things to worry about than installing packages :)

I am not convinced. DNS can sometimes take some time, especially in our fragile infra structure (gobuild, artifactory gerrit or whatever), and I already had to do workaround a do retries in some cases because it just happens, and I have no control over it.

How about a compromise - we leave the default at 0 so there is no change in behavior by default, but set it to 10 seconds in the config file. This way, it's easily reverted or changed.

sshedi:
> I already had to do workaround a do retries in some cases because it just happens, and I have no control over it.

Can you please give more info on this? I'm curious to know where this happened.

> How about a compromise - we leave the default at 0 so there is no change in behavior by default, but set it to 10 seconds in the config file. This way, it's easily reverted or changed.

I'm ok with this, will update the PR.

client/config.c
```diff
@@ -345,6 +349,7 @@ TDNFReadConfig(
     pConf->nOpenMax = TDNF_CONF_DEFAULT_OPENMAX;
     pConf->nInstallOnlyLimit = TDNF_CONF_DEFAULT_INSTALLONLY_LIMIT;
     pConf->nSSLVerify = TDNF_CONF_DEFAULT_SSLVERIFY;
+    pConf->nConnectTimeout = TDNF_CONF_DEFAULT_CONNECT_TMOUT;
```
oliverkurth: Please use `TDNF_CONF_DEFAULT_CONNECT_TIMEOUT`

common/config.h
```diff
@@ -117,6 +118,7 @@
 #define TDNF_CONF_DEFAULT_OPENMAX            1024
 #define TDNF_CONF_DEFAULT_INSTALLONLY_LIMIT  2
 #define TDNF_CONF_DEFAULT_SSLVERIFY          1
+#define TDNF_CONF_DEFAULT_CONNECT_TMOUT      0  // seconds (unlimited)
```
oliverkurth: Also here. `TDNF_CONF_DEFAULT_CONNECT_TIMEOUT`

oliverkurth (commented):
Just a nit.

PR #562, opened 2026-03-14
Fix package installs from command line repo

sshedi:
When installing packages from the command line, repo_internalize() must
be called first, before pool_addfileprovides() and pool_createwhatprovides()

Fixes: e8906072

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -832,9 +832,9 @@ TDNFAddCmdLinePackages(
         queue_push(pQueueGoal, id);
     }
 
+    repo_internalize(pTdnf->pSolvCmdLineRepo);
     pool_addfileprovides(pSack->pPool);
     pool_createwhatprovides(pSack->pPool);
-    repo_internalize(pTdnf->pSolvCmdLineRepo);
 
 cleanup:
     TDNF_SAFE_FREE_MEMORY(pszRPMPath);
--- a/pytests/repo/tdnf-test3.spec
+++ b/pytests/repo/tdnf-test3.spec
@@ -1,0 +1,24 @@
+Summary:    basic install test file.
+Name:       tdnf-test3
+Version:    1.0
+Release:    1
+Vendor:     VMware, Inc.
+License:    VMware
+Url:        http://www.vmware.com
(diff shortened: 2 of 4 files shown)


oliverkurth:
The fix and the test do not seem related - please separate into two commits, and update the title and description, or create a separate PR.

sshedi:
Changed the test to try the same using local repo.
I have kept server related configurations and some additional code, let's keep it in case if we need it someday for testing something quickly. What do you think? @oliverkurth

oliverkurth:
Thank for the changes.

I still have a few issues:
* we do not need to use `tdnf` to prepare the directory - we can just copy. This is more efficient.
* please create separate spec files instead of re-using `tdnf-test-one` and `tdnf-test-two` to isolate this
* remove the `server_reachable()` method - it's just dead code. If we ever need it tit's easy to add again, no rocket science. Also remove `packages_server` from the config.
* I am still not happy with the description - the root cause is that when installing packages from the command line, `repo_internalize()` must be called first, before `pool_addfileprovides()` and `pool_createwhatprovides()`. It has nothing to do with `disablerepo` or `installroot`, although those options are needed to isolate the test case.

sshedi:
> we do not need to use tdnf to prepare the directory - we can just copy. This is more efficient.

It is deliberate decision, downloadonly and repofrompath gets tested along with this. Using cp is easy but better to have tests like these which are real life use cases.

> please create separate spec files instead of re-using tdnf-test-one and tdnf-test-two to isolate this

Done, thought of reducing test time by not building additional packages.

> remove the server_reachable() method - it's just dead code. If we ever need it tit's easy to add again, no rocket science. Also remove packages_server from the config.

Done. The concern was about the time we waste while adding this again to figure things out, if we quickly want to test something while developing something before adding a proper test case.

> I am still not happy with the description - the root cause is that when installing packages from the command line, repo_internalize() must be called first, before pool_addfileprovides() and pool_createwhatprovides(). It has nothing to do with disablerepo or installroot, although those options are needed to isolate the test case.

Done.

PR #565, opened 2026-03-16
implement history id command

oliverkurth:
Useful addition to the `history` command, to be used in scripts to save the current state, so it can be easily reverted back to.

Example:
```
SAVED_ID=$(tdnf history id)
...
tdnf install foobar
...
tdnf hstory rollback ${SAVED_ID}
```


tools/cli/lib/api.c
```diff
@@ -1313,6 +1313,46 @@ TDNFCliRefresh(
...
+    {
+        jd = jd_create(0);
+        CHECK_JD_NULL(jd);
+        CHECK_JD_RC(jd_map_start(jd));
+        CHECK_JD_RC(jd_map_add_int(jd, "Id", nId));
+        pr_json(jd->buf);
+    }
+    else
+    {
+        pr_crit("%d\n", nId);
```
sshedi, line 1346: Instead of printing a plain integer, I think it is better to use a prefix.
```
pr_crit("LatestHistoryID: %d\n", nId);
```
This will help to keep the automation pipelines work based on a definite key value pair.

sshedi (approved):
LGTM, with a suggestion.

tools/cli/lib/api.c
oliverkurth: I want it very simple, so you can just do:
```
SAVED_ID=$(tdnf history id)
...
tdnf hstory rollback ${SAVED_ID}
```
without having to parse the output, without using `cut` or `sed` or whatever.

sshedi (approved):

PR #567, opened 2026-03-16
print url on error

oliverkurth:
To better find an issue if downloading fails, print out the complete URL on error.

### DIFF
--- a/pytests/tests/test_download_errors.py
+++ b/pytests/tests/test_download_errors.py
@@ -1,0 +1,123 @@
+#
+# Copyright (C) 2026 Broadcom, Inc. All Rights Reserved.
+#
+# Licensed under the GNU General Public License v2 (the "License");
+# you may not use this file except in compliance with the License. The terms
+# of the License are located in the COPYING file of this distribution.
+#
+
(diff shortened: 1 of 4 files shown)


pytests/tests/test_download_errors.py
```diff
@@ -0,0 +1,112 @@
+#
```
sshedi, line 1: One generic comment on this test module.
Please keep the error message check more specific.
For instance, for connection error:
```
Error: failed to download https://<url>/packages-stage/5.0/photon_5.0_x86_64/repodata/repomd.xml: Could not connect to server
````
Error check should be:
```
assert BAD_HTTP_URL in stderr and "Could not connect to server" in stderr
```
And likewise for invalid url "Could not resolve hostname"

sshedi (commented):
One good to have request in test module, code changes LGTM.

pytests/tests/test_download_errors.py
oliverkurth: Thanks, I will add that.

sshedi (approved):

PR #568, opened 2026-03-24
Support xml:base in repodata <location> elements

oliverkurth:
### Background

The RPM repodata format allows a `<location>` element in `primary.xml` to carry
an `xml:base` attribute that overrides the base URL used to locate the actual RPM
file:

```xml
<location xml:base="file:///srv/pool" href="attr-2.5.1-6.ph5.x86_64.rpm"/>
```

This is used in "merged repo" scenarios where multiple repositories share a common
RPM pool but each maintain their own repodata. Without `xml:base` support, the
only way to achieve this with tdnf was to use symlinks from each repo directory
into the shared pool — which is not always possible (e.g. on servers that don't
support symlinks).

`xml:base` may be:
- **Absolute** — `file:///srv/pool` or `http://host/pool`: the RPM location is
  independent of the repo's configured base URL
- **Relative** — `pool/`: resolved against the repo's base URL at runtime,
  keeping the repodata relocatable

### What was broken

libsolv correctly parses `xml:base` and stores it as `SOLVABLE_MEDIABASE` on the
solvable. However, `solvable_get_location()` only returns `SOLVABLE_MEDIADIR`/
`SOLVABLE_MEDIAFILE` (the `href` components) — it never consults
`SOLVABLE_MEDIABASE`. As a result, tdnf would ignore `xml:base` entirely and
construct wrong URLs by joining the repo's configured base URL with just the bare
`href` filename, losing the path specified in `xml:base`.

This affected all tdnf operations: `list`, `repoquery --location`, `install`,
`install --downloadonly`, and `--urls`.

### Changes

**`solv/tdnfpackage.c`**, **`solv/prototypes.h`**

Added `SolvGetPkgMediaBaseFromId()` to read `SOLVABLE_MEDIABASE` from a solvable.
Returns `NULL` (not an error) when absent, so normal repos without `xml:base` are
completely unaffected.

**`client/packageutils.c`**

Added a static helper `GetPkgLocationWithMediaBase()` which replaces all four call
sites of `SolvGetPkgLocationFromId()` that populate `pPkgInfo->pszLocation`. When
`SOLVABLE_MEDIABASE` is present it uses `TDNFJoinPath(mediabase, href)` to compose
the effective location. Since `mediabase` is always the first argument, URL schemes
such as `file:///` are preserved correctly by `TDNFJoinPath`. The composed value is:

- An absolute URL (e.g. `file:///srv/pool/pkg.rpm`) when `xml:base` is absolute —
  all downstream code then uses it directly
- A relative path (e.g. `pool/pkg.rpm`) when `xml:base` is relative — joined with
  the repo base URL by existing downstream code as normal

**`client/remoterepo.c`**

- `TDNFDownloadFileFromRepo`: when `pszLocation` already contains a `://` scheme
  (absolute `xml:base` case), skip joining with the repo base URL and pass it
  directly to `TDNFDownloadFile`.
- `TDNFCreatePackageUrl`: same guard for the same reason, used by
  `repoquery --location` and `--urls`.

**`client/rpmtrans.c`**

Added an early check in the package fetch path: when `pszLocation` starts with
`file://` (absolute `xml:base`), strip the scheme and try `access()` directly to
use the file in-place, before falling through to the existing loop that tries
joining base URLs.

**`pytests/tests/test_xmlbase.py`**

New test module covering all `xml:base` variants against both `file://` and
`http://` repositories:

| Class | Variant |
|---|---|
| `TestNoBase` | No `xml:base` — regression guard, existing behaviour unchanged |
| `TestRelativeBase` | Relative `xml:base` (e.g. `pool/`) |
| `TestAbsoluteFileBase` | Absolute `xml:base` with `file://` URL |
| `TestAbsoluteHttpBase` | Absolute `xml:base` with `http://` URL; also tests repodata via HTTP with packages from a different HTTP location |

Each class exercises: `list --available`, `repoquery --location`, `install`, and
`install --downloadonly`.

Repos are built on-the-fly using `createrepo_c --baseurl` (to set `xml:base`) and
`--outputdir` (to separate repodata from the RPM pool), with no new test packages
or spec files required.

## Note

The changes, including the description above, were entirely generated by Cursor, using the Claude Sonnet 4.6 agent, but with human feedback.

### DIFF
--- a/client/packageutils.c
+++ b/client/packageutils.c
@@ -9,6 +9,61 @@
 #define _GNU_SOURCE 1
 #include "includes.h"
 
+/*
+ * Retrieve the effective package location, incorporating xml:base if present.
+ *
+ * If the solvable has SOLVABLE_MEDIABASE set (from <location xml:base="..."/>),
+ * the mediabase and href are joined with TDNFJoinPath. Since mediabase is always
+ * the first argument, URL schemes (e.g. "file:///pool") are preserved correctly.
+ * The result is either an absolute URL (e.g. "file:///pool/pkg.rpm") or a
+ * relative path (e.g. "snap-merged/pkg.rpm") that downstream callers join with
+ * the repo base URL as normal.
+ *
+ * When no mediabase is present the bare href is returned unchanged, preserving
+ * existing behaviour for normal repositories.
+ */
+static uint32_t
+GetPkgLocationWithMediaBase(
+    PSolvSack pSack,
+    uint32_t dwPkgId,
+    char **ppszLocation)
+{
+    uint32_t dwError = 0;
+    char *pszLocation = NULL;
+    char *pszMediaBase = NULL;
+    char *pszComposed = NULL;
+
+    dwError = SolvGetPkgLocationFromId(pSack, dwPkgId, &pszLocation);
+    BAIL_ON_TDNF_ERROR(dwError);
+
+    if (pszLocation)
+    {
+        dwError = SolvGetPkgMediaBaseFromId(pSack, dwPkgId, &pszMediaBase);
+        BAIL_ON_TDNF_ERROR(dwError);
+
+        if (pszMediaBase)
+        {
+            dwError = TDNFJoinPath(&pszComposed, pszMediaBase, pszLocation, NULL);
+            BAIL_ON_TDNF_ERROR(dwError);
+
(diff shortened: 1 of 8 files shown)


sshedi (approved):
LGTM. Thank you.

PR #569, opened 2026-04-06
fix segfault when -c /dev/null (empty config file) is used

oliverkurth:
`tdnf -c /dev/null` would crash with a segmentation fault. Fixing this.


sshedi:
Is it even a valid use case? What is the point of using dev/null as input config file?

oliverkurth:
It's good for testing the built-in defaults.

sshedi:
Probably add a simple test for this?

tapakund (approved):
LGTM

PR #573, opened 2026-04-17
ensure history db directory exists

oliverkurth:
* ensure that the history db directory exists, to avoid failures
* Check for history db errors: the db may have been busy, and the error SQLITE_BUSY wasn't checked.
This lead to invalid entries written to the db, and this caused errors leter when read back.
This change also skips over these invalid entries, and adds a busy timeout (500 seconds),
to wait when busy instead of exting with SQLITE_BUSY immediately.
* fix a few test issues
* Fedora gcc showed a few warnings, also fixed.

### DIFF
--- a/ci/prep.sh
+++ b/ci/prep.sh
@@ -56,25 +56,8 @@ elif grep -qw "Photon" ${rel_file}; then
     zlib-devel
   )
 
-  tdnf-config edit photon-updates enabled=0
-  if [ -f /etc/yum.repos.d/photon-snapshot.repo ] ; then
-    tdnf-config edit photon-snapshot enabled=1
-  else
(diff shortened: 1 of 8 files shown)


ci/prep.sh
```diff
@@ -56,25 +56,19 @@ elif grep -qw "Photon" ${rel_file}; then
...
-gpgcheck=1
-enabled=1
-skip_if_unavailable=1
-skip_md_filelists=1
-EOF
-  fi
-
   tdnf -y upgrade --refresh
-  tdnf remove -y toybox
+  # Workaround for issue in toybox-0.8.9-11.
```
sshedi: New toybox with fix is available now and you can just do "/bin/toybox-toys -- install" to install all toys if you still need this

PR #575, opened 2026-05-13
ignore history absent error if a package is already installed

oliverkurth:
## Summary
Fixes an issue where package operations fail with `Error(1802) : History database does not exist` in fresh environments (like Docker containers) where the history database hasn't been created yet.

## Details
When running commands like `tdnf install <pkg>` where the package is already installed, `tdnf` attempts to mark the package as user-installed. Previously, this would fail if the history database was missing. 

This change gracefully handles a missing history database by catching and ignoring `ERROR_TDNF_HISTORY_NODB` in non-critical paths where it is acceptable for the database to be absent:
* `TDNFMarkAutoInstalledSinglePkg`
* `TDNFAddUserInstalledToJobs`
* `TDNFGetAutoInstalledOrphans`

Since packages are considered user-installed by default, a missing history database shouldn't block these operations.

### DIFF
--- a/client/api.c
+++ b/client/api.c
@@ -2022,7 +2022,6 @@ TDNFHistoryResolve(
     PTDNF_SOLVED_PKG_INFO *ppSolvedPkgInfo)
 {
     uint32_t dwError = 0;
-    int rc = 0;
     char **ppszPkgsNotResolved = NULL;
     Queue queueGoal = {0};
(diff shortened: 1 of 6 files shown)


sshedi:
Please add a test for this

oliverkurth:
> Please add a test for this

Thanks, done.

Test errors are unrelated, failed with the faulty `createrepo__c` package.

sshedi (approved):

PR #576, opened 2026-05-22
global SSL configuration inheritance and SSL_CERT_FILE environment variable

oliverkurth:
## Summary

This PR introduces global SSL configuration inheritance and `SSL_CERT_FILE` environment variable support to `tdnf`, ensuring that SSL settings follow a canonical precedence hierarchy.

Previously, `sslcacert` had to be explicitly defined in every single `.repo` file, which was problematic for transparent MITM proxies. Additionally, `tdnf` did not respect the standard `SSL_CERT_FILE` environment variable.

### Changes Made

- **Global SSL Configuration**: Added support for parsing `sslcacert`, `sslclientcert`, and `sslclientkey` in the `[main]` section of `tdnf.conf`. These settings are now inherited by all repositories unless explicitly overridden.
- **Environment Variable Support**: Added support for the `SSL_CERT_FILE` environment variable to override configuration files.
- **Canonical Precedence Hierarchy**: Enforced the standard configuration precedence (from highest to lowest priority):
  1. Command-line arguments (`--setopt=sslcacert=...` or `--setopt=repoid.sslcacert=...`)
  2. Environment variable (`SSL_CERT_FILE`)
  3. Per-repository configuration (`.repo` files)
  4. Global configuration (`tdnf.conf`)
- **Memory Leak Fix**: Fixed a pre-existing memory leak where `pszSnapshotUrl` and `pszSnapshotFile` were not being freed when destroying repository data structures. Consolidated `TDNFFreeReposInternal` and `TDNFFreeRepos` to prevent future drift.
- **Test Suite Updates**: Added a new pytest module (`test_ssl_config.py`) that spins up a local HTTPS server with self-signed certificates to verify the new inheritance and precedence logic. Updated `TestUtils.run()` in `conftest.py` to accept custom environment variables for testing.

### DIFF
--- a/ci/prep.sh
+++ b/ci/prep.sh
@@ -37,6 +37,7 @@ if grep -qw "Fedora" ${rel_file}; then
     glib2-devel
     libcurl-devel
     make
+    openssl
     python3-flake8
     rpm-devel
     rpm-sign
--- a/client/api.c
+++ b/client/api.c
@@ -2579,7 +2579,7 @@ TDNFCloseHandle(
     {
         if(pTdnf->pRepos)
         {
-            TDNFFreeReposInternal(pTdnf->pRepos);
+            TDNFFreeRepos(pTdnf->pRepos);
         }
         if(pTdnf->pConf)
         {
--- a/client/config.c
+++ b/client/config.c
@@ -177,6 +177,18 @@ TDNFConfigFromCnfTree(PTDNF_CONF pConf, struct cnfnode *cn_top)
         {
             pConf->nSSLVerify = isTrue(cn->value);
         }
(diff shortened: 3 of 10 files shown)


ci/prep.sh
```diff
@@ -37,6 +37,7 @@ if grep -qw "Fedora" ${rel_file}; then
     glib2-devel
     libcurl-devel
     make
+    openssl
```
sshedi, line 40: nit: openssl-devel is already present in this list, so this is redundant

sshedi (approved):

ci/prep.sh
oliverkurth: I had to add this because in Fedora there was no `/usr/bin/openssl`.

PR #581, opened 2026-06-15
support wildcards and relative expressions for repoqery

oliverkurth:
## Summary
Adds support for wildcards and relational strings to `tdnf repoquery --whatrequires`.

Previously, `--whatrequires` converted the query string to a libsolv ID using `pool_str2id` and matched using `solvable_matchesdep`, which only supported exact string matching. This prevented queries like `pkgconfig(*)` or `pkgconfig(gmp) >= 6.0` from resolving properly.

## Changes
- **Refactored `SolvApplyDepsFilter`**: Replaced the exact-match string parsing logic in `solv/tdnfquery.c` with libsolv's native `selection_make_matchdeps` API.
- **Enabled Glob and Relational Matching**: Applied the `SELECTION_GLOB | SELECTION_REL` flags, allowing libsolv to automatically parse wildcards and version constraints.
- **Optimized Match Filtering**: Updated the result queue filtering to use a libsolv `Map` to track matches resolved via `selection_solvables`. 
- **Tests**: Added a test case in `pytests/tests/test_repoquery.py` for wildcard queries.

## Example
```
tdnf repoquery --whatrequires 'pkgconfig(*)'
GConf-devel-3.2.6-3.ph5.aarch64
ImageMagick-c++-devel-7.1.2.19-1.1.ph5.aarch64
ImageMagick-devel-7.1.2.19-1.1.ph5.aarch64
ModemManager-devel-1.18.12-3.ph5.aarch64
...
```

### DIFF
--- a/pytests/tests/test_repoquery.py
+++ b/pytests/tests/test_repoquery.py
@@ -351,3 +351,12 @@ def test_queryformat(utils):
 
     assert BASE_PKG in '\n'.join(ret['stdout'])
     assert '\n'.join(ret['stdout']).startswith("http://localhost:8080/photon-test")
+
+
+def test_repoquery_wildcard_whatrequires(utils):
+    ret = utils.run(['tdnf',
+                     'repoquery',
+                     '--whatrequires',
+                     'tdnf*'])
+    assert ret['retval'] == 0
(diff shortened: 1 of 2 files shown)


sshedi (approved):
This is a good feature.

PR #584, opened 2026-06-26
Fix repoquery exit status on failure

sshedi:
Return with non zero exit status when repoquery is a miss.


oliverkurth:
This looks good, but you need to resolve the conflict.

PR #587, opened 2026-06-26
Show full list of not found packages while erroring out

sshedi:


oliverkurth:
Why do you want to merge into `stable-3.3`? Who uses that branch?

Also, please set labels for PRs for stable branches.

PR #589, opened 2026-07-07
Pressing a plain enter should not break the prompt

sshedi:


oliverkurth:
Thanks for this, but please be more descriptive, so we have a records if something comes with this change.

sshedi:
Assume there is a script which is doing a bunch of tdnf operations
```
#!/bin/bash

...
tdnf install <pkgs>

...
Some sort of random delay
...
tdnf remove <pkgs>
...
```
In the lifetime of above script, some thing might make user feel that operation is stuck or machine is frozen; and to test it if user presses enter key few times the "tdnf remove ..." post delay will abruptly stop the script without prompting user. This happened while I was trying some things with photon-upgrade recently.

oliverkurth:
Verified:
```
$ sudo ./bin/tdnf --releasever=5.0 --installroot=$(pwd)/installroot remove lsof

Removing:
lsof                                                       aarch64                                     4.96.4-2.ph5                                               @System                                     236.58k                                               0.00b

Total installed size: 236.58k
Total download size:   0.00b
Is this ok [y/N]:
Is this ok [y/N]:
Is this ok [y/N]:
Is this ok [y/N]:
Is this ok [y/N]: N
```

PR #592, opened 2026-07-14
fix: sort directory listings when discovering repos and local repo rpms

oliverkurth:
`readdir()` order is filesystem-dependent and not guaranteed stable across runs, which made repo config discovery (`repolist.c`) and local directory-repo rpm discovery (`tdnfrepo.c`) a source of non-deterministic solvable-Id assignment. Switch both to `scandir()`+`alphasort()` so repo/rpm processing order is consistent regardless of underlying filesystem.

### DIFF
--- a/client/config.c
+++ b/client/config.c
@@ -797,6 +797,17 @@ cleanup:
         }
         free(ppDirEntries);
     }
+    if (pppszArrayList)
+    {
+        /* on error, entries already read from files (but not yet moved
+         * into ppszNewLines) still own their string arrays; free them.
+         * Stop before the appended *pppszLines entry, which the caller
+         * still owns when we bail out before it is replaced below. */
+        for (i = 0; pppszArrayList[i] && pppszArrayList[i] != *pppszLines; i++)
+        {
+            TDNF_SAFE_FREE_STRINGARRAY(pppszArrayList[i]);
+        }
+    }
     TDNF_SAFE_FREE_MEMORY(pppszArrayList);
     TDNF_SAFE_FREE_MEMORY(pszFile);
     return dwError;
--- a/client/config.c
+++ b/client/config.c
@@ -700,8 +700,9 @@ TDNFReadConfFilesFromDir(
     )
 {
     uint32_t dwError = 0;
-    DIR *pDir = NULL;
-    struct dirent *pEnt = NULL;
+    struct dirent **ppDirEntries = NULL;
+    int nDirEntries = -1;
+    int nDirIndex;
     char *pszFile = NULL;
     char **ppszNewLines = NULL;
     char ***pppszArrayList = NULL;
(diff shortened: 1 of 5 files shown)


solv/tdnfrepo.c
```diff
@@ -311,7 +315,7 @@ readRpmsFromDir(
         if (isDir) {
             dwError = readRpmsFromDir(pRepo, pszPath);
             BAIL_ON_TDNF_ERROR(dwError);
-        } else if (strcmp(&pEnt->d_name[strlen(pEnt->d_name)-4], ".rpm") == 0) {
+        } else if (strcmp(&pszName[strlen(pszName)-4], ".rpm") == 0) {
```
sshedi: Please use a variable to store retval of strlen and validate it, this may result in negative array index and use strncmp.

client/plugins.c
```diff
@@ -263,33 +264,37 @@ _TDNFLoadPluginConfigs(
...
-                      pEnt->d_name,
+                      pszName,
                       NULL);
         BAIL_ON_TDNF_ERROR(dwError);
 
         dwError = _TDNFLoadPluginConfig(pszPluginConfig, &pPlugin);
         BAIL_ON_TDNF_ERROR(dwError);
 
-        dwError = TDNFAllocateStringN(pEnt->d_name, nLen - nExtLen, &pPlugin->pszName);
+        dwError = TDNFAllocateStringN(pszName, nLen - nExtLen, &pPlugin->pszName);
```
sshedi, line 297: Found by Claude:
```
pPlugin leaked when TDNFAllocateStringN fails


dwError = _TDNFLoadPluginConfig(pszPluginConfig, &pPlugin);  // allocates pPlugin
BAIL_ON_TDNF_ERROR(dwError);

dwError = TDNFAllocateStringN(pszName, nLen - nExtLen, &pPlugin->pszName);
BAIL_ON_TDNF_ERROR(dwError);  // → error: TDNFFreePlugins(pPlugins); goto cleanup
// pPlugin is NOT yet in pPlugins at this point → leaked
The error: path at line 329 calls TDNFFreePlugins(pPlugins), but pPlugin has not been linked into pPlugins yet (that happens at lines 303–311). Fix: call _TDNFFreePlugin(pPlugin) before the bail, or set pPlugin = NULL after linking it.
```

client/config.c
```diff
@@ -791,9 +789,13 @@ TDNFReadConfFilesFromDir(
...
+    if (ppDirEntries)
     {
-        closedir(pDir);
+        for (nDirIndex = 0; nDirIndex < nDirEntries; nDirIndex++)
+        {
+            free(ppDirEntries[nDirIndex]);
+        }
+        free(ppDirEntries);
     }
     TDNF_SAFE_FREE_MEMORY(pppszArrayList);
```
sshedi, line 811:
```
Inner string arrays in pppszArrayList leaked on error


cleanup:
    ...
    TDNF_SAFE_FREE_MEMORY(pppszArrayList);  // frees the outer pointer array only
If BAIL_ON_TDNF_ERROR fires inside the second for loop (lines 752–760) after some pppszArrayList[i] slots have been populated by TDNFReadFileToStringArray, those inner char** arrays are not freed — only the outer pointer array is. The success path correctly frees them (line 786). The error path needs an explicit loop, similar to what the success path does at lines 780–787.
```

sshedi (commented):
Please add a test for this

oliverkurth:
Thank you, that looks like some major issues. I trusted Claude too much, and pushed before doing a more thorough review.

oliverkurth:
> Thank you, that looks like some major issues. I trusted Claude too much, and pushed before doing a more thorough review.

Hm, looks like pre-existing issues. But need to be fixed in any case.

sshedi:
Changes LGTM, please add a tests for the same.

sshedi (approved):

PR #593, opened 2026-08-16
fix: allow disabling snapshot at runtime via --setopt=<repo>.snapshot=

sshedi:
Passing an empty value to --setopt previously stored "" in pszSnapshotUrl. The old NULL-only guard let the empty string reach TDNFUriIsRemote(), which rejects it with EINVAL (error 1622). Switch the guard to IsNullOrEmptyString() so that an empty value is treated the same as no snapshot being configured, skipping snapshot filtering entirely.

Also add test_disable_snapshot_via_setopt to test_snapshots.py to verify that previously excluded packages become visible when the snapshot is disabled at runtime.

### DIFF
--- a/client/repo.c
+++ b/client/repo.c
@@ -693,7 +693,7 @@ TDNFGetRepoMD(
                   &pRepoMD);
     BAIL_ON_TDNF_ERROR(dwError);
 
-    if (pRepoData->pszSnapshotUrl) {
+    if (!IsNullOrEmptyString(pRepoData->pszSnapshotUrl)) {
         time_t now = time(NULL);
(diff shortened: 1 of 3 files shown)


client/repo.c
```diff
@@ -37,8 +37,9 @@ TDNFApplySnapshot(
 
     for(i = 0; ppszSnapshotPackages[i]; i++) {
         pszPkg = ppszSnapshotPackages[i];
-        if (pszPkg[0] == '#')
```
oliverkurth, line 40: This looks unrelated - what is the use case for this?

oliverkurth (commented):
Just one question, the rest looks good to me.

PR #594, opened 2026-08-16
fix: apply global setopts to all repos including repofrompath

sshedi:
Global --setopt=<key>=<value> (no repo prefix) was silently ignored for repo-level keys like skip_md_filelists because cn_setopts was only passed to TDNFConfigFromCnfTree (global conf), which does not handle repo-level settings.

Apply cn_setopts to all repos via TDNFRepoConfigFromCnfTree before the per-repo setopt loop so that global setopts reach every repo, including those created by --repofrompath and --repofromdir. Per-repo setopts (with a repoid prefix) still take precedence as they are applied after.

Add tests for:
- Global and per-repo setopt overriding the .repo file value for skip_md_filelists and skip_md_other (test_skip_md.py)
- Global setopt (skip_md_filelists, skip_md_other) and per-repo setopt applied to --repofrompath repos (test_repofrompath.py)

### DIFF
--- a/client/repolist.c
+++ b/client/repolist.c
@@ -303,10 +303,39 @@ TDNFLoadRepoData(
         }
     }
 
-    /* look for setopt settings */
+    /* Apply per-repo and glob-pattern setopts.
+     * Pass 1: glob patterns first. Pass 2: exact repo-id matches, which
+     * always take precedence over globs regardless of argument order.
+     * For array-valued keys (baseurl, gpgkey), the exact match resets any
+     * values set by globs so it replaces rather than appends. */
     if (pTdnf->pArgs->cn_repoopts != NULL) {
         for (pRepo = pReposAll; pRepo; pRepo = pRepo->pNext) {
-            if ((cn_repo = find_child(pTdnf->pArgs->cn_repoopts, pRepo->pszId)) != NULL) {
+            if (strcmp(pRepo->pszId, CMDLINE_REPO_NAME) == 0)
+                continue;
+            for (cn_repo = pTdnf->pArgs->cn_repoopts->first_child;
+                 cn_repo; cn_repo = cn_repo->next) {
+                if (!TDNFIsGlob(cn_repo->name))
+                    continue;
+                if (fnmatch(cn_repo->name, pRepo->pszId, 0) == 0) {
+                    dwError = TDNFRepoConfigFromCnfTree(pTdnf, pRepo, cn_repo);
+                    BAIL_ON_TDNF_ERROR(dwError);
+                }
+            }
+        }
+        for (pRepo = pReposAll; pRepo; pRepo = pRepo->pNext) {
+            for (cn_repo = pTdnf->pArgs->cn_repoopts->first_child;
+                 cn_repo; cn_repo = cn_repo->next) {
+                struct cnfnode *cn_key;
+                if (TDNFIsGlob(cn_repo->name))
+                    continue;
+                if (strcmp(cn_repo->name, pRepo->pszId) != 0)
+                    continue;
+                for (cn_key = cn_repo->first_child; cn_key; cn_key = cn_key->next) {
+                    if (strcmp(cn_key->name, TDNF_REPO_KEY_BASEURL) == 0)
+                        TDNF_SAFE_FREE_STRINGARRAY(pRepo->ppszBaseUrls);
+                    else if (strcmp(cn_key->name, TDNF_REPO_KEY_GPGKEY) == 0)
+                        TDNF_SAFE_FREE_STRINGARRAY(pRepo->ppszUrlGPGKeys);
+                }
                 dwError = TDNFRepoConfigFromCnfTree(pTdnf, pRepo, cn_repo);
                 BAIL_ON_TDNF_ERROR(dwError);
             }
--- a/tools/cli/lib/parseargs.c
+++ b/tools/cli/lib/parseargs.c
@@ -440,7 +440,8 @@ ParseOption(
 
                 *pseq_dot = 0;
 
-                if (!TDNFStrIsValidRepoName(pszCopyArgs)) {
+                if (!TDNFIsGlob(pszCopyArgs) &&
+                    !TDNFStrIsValidRepoName(pszCopyArgs)) {
                     dwError = ERROR_TDNF_INVALID_REPO_NAME;
                     BAIL_ON_CLI_ERROR(dwError);
                 }
--- a/client/prototypes.h
+++ b/client/prototypes.h
@@ -903,11 +903,6 @@ TDNFGetFileSize(
     int *pnSize
     );
 
-int
-TDNFIsGlob(
-    const char* pszString
-    );
-
 uint32_t
 TDNFUtilsMakeDir(
     const char* pszPath
--- a/client/utils.c
+++ b/client/utils.c
@@ -123,22 +123,6 @@ TDNFGetCurlError(
     return dwCurlError;
 }
 
-int
-TDNFIsGlob(
-    const char *pszString
-    )
-{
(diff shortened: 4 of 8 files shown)


oliverkurth:
I don't like this - it would apply global setopts (from the command line) to the repos, but not the global options from `tdnf.conf`. Both should be handled the same. Additionally, I do not think it makes sense to do that for all options. Some options are exclusively global, and some are repo specific.

What you actually want is apply a setting to all repos - and a better syntax for that would be to use globs, like `--setopt *.skip_md_filelists=1`. That way you could even select all `photon` repos, like `--setopt photon*.skip_md_filelists=1`.

sshedi:
> I don't like this - it would apply global setopts (from the command line) to the repos, but not the global options from `tdnf.conf`. Both should be handled the same. Additionally, I do not think it makes sense to do that for all options. Some options are exclusively global, and some are repo specific.
> 
> What you actually want is apply a setting to all repos - and a better syntax for that would be to use globs, like `--setopt *.skip_md_filelists=1`. That way you could even select all `photon` repos, like `--setopt photon*.skip_md_filelists=1`.

Fair point on the design. One concern with moving to glob-only: it would break existing usage of `--setopt=key=value` for settings that are currently handled globally. Could we keep the existing no-prefix form working as-is, and add glob support on top? That way there's no regression and users get the more expressive syntax you're describing.

oliverkurth:
> Fair point on the design. One concern with moving to glob-only: it would break existing usage of `--setopt=key=value` for settings that are currently handled globally. Could we keep the existing no-prefix form working as-is, and add glob support on top? That way there's no regression and users get the more expressive syntax you're describing.

I don't understand - how would this cause a regression? You just need to change the code that parses `foo.key=value` to also accept `*.key=value` so it applies to all repos, or `photon*.key=value` so it applies to all repos starting with "photon". How will this affect usage of `--setopt=key=value` ?

sshedi:
> > Fair point on the design. One concern with moving to glob-only: it would break existing usage of `--setopt=key=value` for settings that are currently handled globally. Could we keep the existing no-prefix form working as-is, and add glob support on top? That way there's no regression and users get the more expressive syntax you're describing.
> 
> I don't understand - how would this cause a regression? You just need to change the code that parses `foo.key=value` to also accept `*.key=value` so it applies to all repos, or `photon*.key=value` so it applies to all repos starting with "photon". How will this affect usage of `--setopt=key=value` ?

You're right, sorry for the confusion. I've implemented the glob support as you suggested - *.key=value applies to all repos, photon*.key=value to matching ones. The existing --setopt=key=value form is unchanged.

client/repolist.c
```diff
@@ -303,6 +303,22 @@ TDNFLoadRepoData(
...
+                if (strcmp(pRepo->pszId, CMDLINE_REPO_NAME) == 0)
+                    continue;
+                if (fnmatch(cn_glob->name, pRepo->pszId, 0) == 0) {
+                    dwError = TDNFRepoConfigFromCnfTree(pTdnf, pRepo, cn_glob);
+                    BAIL_ON_TDNF_ERROR(dwError);
+                }
+            }
+        }
+    }
+
```
oliverkurth: Instead of adding the block, you can modify the block below (`if (pTdnf->pArgs->cn_repoopts != NULL) { ...`)  to also check for globs (you won't be able to use `find_child()` - either implent a direct search, or create a `find_child_byglob()`).

And that means also you do not need to add the `cn_globopts` field.

tools/cli/lib/parseargs.c
```diff
@@ -440,19 +441,31 @@ ParseOption(
...
+                    cn_repo = find_child(pCmdArgs->cn_globopts, pszCopyArgs);
+                    if (!cn_repo) {
+                        cn_repo = create_cnfnode(pszCopyArgs);
+                        append_node(pCmdArgs->cn_globopts, cn_repo);
+                    }
+                    cn = create_cnfnode(pseq_dot+1);
+                    cnfnode_setval(cn, psep_eq + 1);
+                    append_node(cn_repo, cn);
+                } else {
+                    if (!TDNFStrIsValidRepoName(pszCopyArgs)) {
```
oliverkurth: Just as in my previous comment, this block can be merged into the other one below.

However, I think we shuld still check with `TDNFStrIsValidRepoName()` if `pszCopyArgs` is not a glob, to catch mistyped repo names.

oliverkurth (changes requested):
Thank you for the changes, this looks like what  suggested in the result. But I think the code can be simplified, see comments.

client/repolist.c
```diff
@@ -303,12 +303,21 @@ TDNFLoadRepoData(
         }
     }
 
-    /* look for setopt settings */
+    /* Apply per-repo and glob-pattern setopts.
+     * Glob patterns (e.g. *.key=val, photon*.key=val) are stored in cn_repoopts
+     * alongside exact repo IDs and matched via fnmatch. Globs are not applied
+     * to the @cmdline repo whose gpgcheck state is managed separately. */
```
oliverkurth: "Globs are not applied to the @cmdline repo whose gpgcheck state is managed separately."

What is that supposed to mean? That comment makes no sense.

oliverkurth (commented):
Changes look good to me, but that comment is puzzling.

client/repolist.c
oliverkurth: I let Claude run a review, and this what it came up with - I think points 1 and 2 are acceptable limitations. Just don't use multiple `setopt` args that may overlap. I do not think that would be a common use case.

I think point 3 would have higher priority, but I can live with that too.

## Review of #594

Thanks for adding glob support to `--setopt` repo matching! Found a couple of correctness issues around how overlapping matches combine, plus a minor cleanup suggestion.

### 1. No precedence between exact and glob `--setopt` matches (correctness, security-relevant)

`client/repolist.c` (around line 317, in `TDNFLoadRepoData`)

Before this PR, at most one `cn_repoopts` node could ever match a given repo. Now the loop applies `TDNFRepoConfigFromCnfTree` once per **every** matching node — exact and glob alike — with no specificity/precedence rule between them.

For scalar options (e.g. `gpgcheck`), whichever node happens to be processed last (i.e. command-line argument order) wins, regardless of specificity:

```sh
tdnf --setopt=updates.gpgcheck=1 --setopt=up*.gpgcheck=0 install pkg
```

This silently disables `gpgcheck` for `updates` even though the exact, more specific setting said otherwise — simply reordering the two flags flips the effective (security-relevant) setting.

### 2. Overlapping setopts silently duplicate array-valued options

`client/repolist.c` (around line 317)

For array-valued options (`baseurl`, `gpgkey`), values are appended via `TDNFAddStringArray` for every matching node instead of the more specific match winning:

```sh
tdnf --setopt=updates.baseurl=A --setopt=up*.baseurl=B ...
```

...results in the `updates` repo ending up with **two** baseurls (`A` and `B`) instead of the single value presumably intended. This combination (exact + overlapping glob) is untested by the new pytest cases in `test_repofrompath.py`/`test_skip_md.py`, which only exercise glob-only or exact-only scenarios in isolation.

### 3. Minor: duplicated glob-detection logic

`client/repolist.c` (around line 314)

```c
strpbrk(cn_repo->name, "*?[")
```

This reimplements what the existing `TDNFIsGlob()` helper already does — and `TDNFIsGlob()` is already called elsewhere in this same file. Worth reusing it here to avoid the two checks drifting apart if the glob definition ever changes (e.g. adding brace-expansion support).

---

Everything else looks solid: the `@cmdline` glob-exclusion guard, `find_child` dedup of identical glob patterns, `fnmatch.h` inclusion/malformed-pattern handling, and the safety of the nested-loop refactor replacing the old single-match `if` all check out.

oliverkurth:
Thank you for making these changes.

PR #596, opened 2026-08-21
fix: call transaction_order() to get deterministic install ordering

sshedi:
solver_create_transaction() returns pTrans->steps in an order influenced by the libsolv job-queue, which mirrors CLI argument order. rpmtsOrder() then picks whichever Requires(pre) provider appears closest to the consumer in the RPM transaction as the ordering anchor, making the install sequence non-deterministic across different CLI argument orderings.

Calling transaction_order(pTrans, 0) immediately after solver_create_transaction() sorts pTrans->steps into proper dependency-graph order before it is used to populate the RPM transaction, ensuring a consistent provider is selected as the anchor regardless of CLI order.

Regression: installing 'filesystem linux-tools bash toybox' vs 'filesystem linux-tools toybox bash' could result in either bash or toybox being used to run linux's %pre scriptlet. Toybox sh exits 1 for an if-branch that is never taken (non-POSIX), causing the install to fail.

Add a regression test using four minimal packages that reproduce the non-deterministic ordering with two competing Requires(pre) providers.

### DIFF
--- a/client/goal.c
+++ b/client/goal.c
@@ -391,6 +391,7 @@ TDNFSolv(
             BAIL_ON_TDNF_ERROR(dwError);
         }
 
+        transaction_order(pTrans, 0);
         if (pTdnf->pConf->ppszProtectedPkgs) {
             /* catch protected obsoleted packages, and double check for removals */
             dwError = TDNFSolvCheckProtectPkgsInTrans(pTdnf, pTrans, pTdnf->pSack->pPool);
--- a/pytests/repo/tdnf-test-prein-base.spec
+++ b/pytests/repo/tdnf-test-prein-base.spec
@@ -1,0 +1,21 @@
+Summary:    Test package - prein ordering base dependency
+Name:       tdnf-test-prein-base
+Version:    1.0
+Release:    1
+Vendor:     VMware, Inc.
+Distribution:   Photon
+License:    VMware
+Url:        http://www.vmware.com
+Group:      Applications/tdnftest
+
+%description
+Base package required by tdnf-test-prein-dep-a to give it more graph
+weight than tdnf-test-prein-dep-b for transaction ordering tests.
+
+%prep
+%build
+%install
+%files
+%changelog
+*   Fri Aug 21 2026 tdnf team 1.0-1
+-   Initial package
--- a/pytests/repo/tdnf-test-prein-consumer.spec
+++ b/pytests/repo/tdnf-test-prein-consumer.spec
@@ -1,0 +1,27 @@
+Summary:    Test package - prein ordering regression, consumer
+Name:       tdnf-test-prein-consumer
+Version:    1.0
+Release:    1
+Vendor:     VMware, Inc.
+Distribution:   Photon
+License:    VMware
+Url:        http://www.vmware.com
+Group:      Applications/tdnftest
+Requires(pre): tdnf-test-prein-cap
+
+%description
+Consumer of tdnf-test-prein-cap via Requires(pre). The %pre scriptlet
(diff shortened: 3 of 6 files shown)


sshedi:
To test the existing issue, use the following commands in 5.0-91 (with Update968 snapshot)
```
tdnf --releasever 5.0 --installroot $PWD/rootdir install -y filesystem linux-tools toybox bash; rm -rf $PWD/rootdir

tdnf --releasever 5.0 --installroot $PWD/rootdir install -y filesystem linux-tools bash toybox; rm -rf $PWD/rootdir
```

oliverkurth:
Looks good to me. But I used Claude for a review again, it found a few nits with the tests:

## Review of #596

The core fix — adding `transaction_order(pTrans, 0);` right after `solver_create_transaction()` in `client/goal.c` — was traced end-to-end and is correct: it makes install ordering deterministic based on `pTrans->steps` regardless of CLI argument order. No other call site of `solver_create_transaction()` needs the same treatment, and the four new `.spec` test packages follow the existing convention.

Two issues in the new regression test, `pytests/tests/test_transaction_order.py`:

### 1. Missing pre-test cleanup (test-coverage)

The autouse fixture only removes the test packages in teardown, never before the test starts:

```python
@pytest.fixture(scope="function", autouse=True)
def cleanup_prein_packages(utils):
    yield
    utils.run(["tdnf", "remove", "-y"] + PREIN_PKGS)
```

If a prior run was interrupted (CI timeout, crash) and left `tdnf-test-prein-*` packages installed, this test silently degrades: tdnf finds nothing to install, both order lists come out empty/degenerate, and the assertion passes without ever exercising the ordering logic — turning the regression test into a silent no-op.

### 2. Unchecked intermediate removal (test-coverage)

```python
utils.run(["tdnf", "remove", "-y"] + PREIN_PKGS)
```

between the two install phases doesn't check `retval`. A failed/partial removal (transient lock, dependency issue) leaves some packages installed, causing the second install's result set to differ from the first for reasons unrelated to ordering — a flaky failure that misattributes an environment issue to the fix being broken.

### Suggested fix

```diff
 @pytest.fixture(scope="function", autouse=True)
 def cleanup_prein_packages(utils):
+    utils.run(["tdnf", "remove", "-y"] + PREIN_PKGS)
     yield
     utils.run(["tdnf", "remove", "-y"] + PREIN_PKGS)


 def _install_and_get_order(utils, pkgs):
     ret = utils.run(["tdnf", "install", "-y", "--nogpgcheck"] + pkgs)
     assert ret["retval"] == 0, "tdnf install failed (retval={}):\n{}".format(
         ret["retval"], "\n".join(ret["stdout"] + ret["stderr"])
     )
     return [line for line in ret["stdout"] if line.startswith("Installing/Updating:")]


+def _remove_prein_packages(utils):
+    ret = utils.run(["tdnf", "remove", "-y"] + PREIN_PKGS)
+    assert ret["retval"] == 0, "tdnf remove failed (retval={}):\n{}".format(
+        ret["retval"], "\n".join(ret["stdout"] + ret["stderr"])
+    )
+
+
 # Regression test for transaction_order() fix in client/goal.c:
 ...
 def test_prein_install_order_is_deterministic(utils):
     order_a_first = _install_and_get_order(
         utils, [PREIN_DEP_A, PREIN_DEP_B, PREIN_CONSUMER]
     )

-    utils.run(["tdnf", "remove", "-y"] + PREIN_PKGS)
+    _remove_prein_packages(utils)

     order_b_first = _install_and_get_order(
         utils, [PREIN_DEP_B, PREIN_DEP_A, PREIN_CONSUMER]
     )
```

oliverkurth:
Thank you!

PR #601, opened 2026-08-30
stable-3.3: fix: skip printing problems filtered by --skipconflicts/--skipobsoletes

sshedi:
In SolvReportProblems, pr_err was unconditionally called after the SkipBasedOnType check, so skipped problems were still printed even though the exit code was correctly 0. Use continue to skip the entire problem (including the print) when SkipBasedOnType returns true.

### DIFF
--- a/solv/tdnfpackage.c
+++ b/solv/tdnfpackage.c
@@ -1730,11 +1730,12 @@ SolvReportProblems(
             }
         }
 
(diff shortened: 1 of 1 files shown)


oliverkurth:
Same issue in `dev`, already fixed: https://github.com/vmware/tdnf/commit/2d213641523466299c0a3340fea30c14a5298b65

PR #604, opened 2026-09-19
feat: add --dry-run flag to preview transactions without executing

sshedi:
Resolves dependencies and prints what would be installed/upgraded/erased, then exits without downloading packages or touching the system. Unlike --testonly, no RPM transaction set is built and no packages are fetched.

Includes pytest coverage for: no install/erase on dry-run, plan output, completion message, nothing-to-do on already-installed, no RPM downloaded to cache, dep resolution shown, and failure on unresolvable package.


tools/cli/lib/parseargs.c
```diff
@@ -62,6 +62,7 @@ static struct option pstOptions[] =
     {"skipobsoletes", no_argument, 0, 0},                  //--skipobsoletes to skip obsolete problems
     {"skipsignature", no_argument, &_opt.nSkipSignature, 1}, //--skipsignature to skip verifying RPM signatures
     {"source",        no_argument, &_opt.nSource, 1},
+    {"dry-run",       no_argument, &_opt.nDryRun, 1},
```
oliverkurth: This list is ordered alphabetically.

Also, I would prefer "dryrun". Almost all options have no dashes (convention from `dnf`), and the exceptions probably should not have them either. If we keep that way, it's easier to remember if it's with or without a dash.

oliverkurth (changes requested):
Thank you, I like this feature, it may be very useful. Just a few nits.

oliverkurth:
One other note though - the same can be achieved by just using `--asumeno`:

```
$ sudo ./bin/tdnf --releasever=5.0 --assumeno install lsof

Installing:
lsof                                                              aarch64                                           4.96.4-2.ph5                                                      photon-updates                                    236.58k                                                   122.32k
libtirpc                                                          aarch64                                           1.3.3-4.ph5                                                       photon-updates                                    204.24k                                                    93.49k
...
photon-updates                                    274.29k                                                   129.36k
filesystem                                                        aarch64                                           1.1-11.ph5                                                        photon-updates                                      8.27k                                                    25.38k

Total installed size:  42.64M
Total download size:  10.46M
```

The only difference is the exit code and the error message.

